PR #21, opened 2025-07-24
Sync with meta-vulnscout

ValentinBoudevin:
Modify example and removed useless scripts.

The .vulnscout folder now follow the hierarchy used in meta-vulnscout (https://github.com/savoirfairelinux/meta-vulnscout) with .vulnscout/name/output structure.

Also, removed vulnscout.sh script. The project setting is now managed by the new yocto layer and it doesn't require it. Only, the docker-compose method should be kept.

The script vulnscout.sh was used to update vulnscout, but will be handled with the meta-vulnscout configuration.


moh60 (approved):
LGTM

PR #22, opened 2025-07-24
spdx: address comments from pull request

katekm:
- Use regex for CVE detection
- Fix misplaced 'from' and 'to' in VEX relationship
- Add more tests


moh60:
Covered in https://github.com/savoirfairelinux/vulnscout/pull/25

PR #20, opened 2025-07-24
add SELinux support

dupremathieu:
Modified the docker run and the docker-compose template to add the Z option to the volume mounts.

PR #32, opened 2025-08-01
add SELinux support

ValentinBoudevin:
Modified the docker run and the docker-compose template to add the Z option to the volume mounts.

Based on the opened : https://github.com/savoirfairelinux/vulnscout/pull/20

(now we should use forks and not branch so can't push on the top of old PR)


rehsfl (approved):
LGTM

PR #151, opened 2025-11-21
Update architecture schema

cunysusanne:
### Changes proposed in this pull request:

* Add new schema with logos and colors
* Add svg format of the schema

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

## Pull Request Checklist

Please review and check all that apply before submitting your PR:

- [ ] The code compiles and passes all tests
- [ ] All new and existing tests are passing
- [ ] Documentation has been updated (if applicable)
- [ ] Code follows project style guidelines
- [ ] No sensitive information is included
- [ ] Linked relevant issues (if any)
- [ ] Added necessary reviewers

PR #152, opened 2025-11-24
Edit architecture schema

aishkandoth:

* Changes after the review of https://github.com/savoirfairelinux/vulnscout/pull/151

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### Additional notes

This PR is created on top of the PR from @cunysusanne. Only major change from the original PR is the removal of `NVD DB` element after review.

## Pull Request Checklist

### DIFF
--- a/frontend/src/pages/Metrics.tsx
+++ b/frontend/src/pages/Metrics.tsx
@@ -266,7 +266,9 @@ const packageColumns = [
     ),
     cell: (info: any) => (
       <div className="flex items-center justify-center h-full text-center">
-        {info.getValue()}
+        <span className="truncate max-w-full block" title={info.getValue()}>
+          {info.getValue()}
+        </span>
       </div>
     ),
     enableSorting: false,


ValentinBoudevin (changes requested):
Duplicated with https://github.com/savoirfairelinux/vulnscout/pull/151

rehsfl (changes requested):
Please note the following:
* We still need to emphasize the role of NVD.DB / EPSS.DB within the system even though it was included as a data entry, the files play a persistence role.
* OSV-Scanner is not currently in use by VulnScout, it is present in the script but its eventual output is not used, it might also be removed from sources

aishkandoth:
> 
> *OSV-Scanner support for SPDX3 is not available yet but the PR is on-going. So, I would modify it to "Coming Soon" *Concerning the NVD.DB, maybe we should re add-it but it should point to an explicit "Patch-Finder" text

Are these still valid?

PR #153, opened 2025-11-26
Remove justification when status is not not_affected

rawaha-e:

* When an assessment is edited, we need to make sure the justification is removed when it's no longer not_affected / false_positive (which itself is a case of not_affected)

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Open VS web dashboard, select a vuln with not_affected status then change it to another status, observe that there is no longer a justification associated to it.

## Pull Request Checklist


mrodriguezg1991 (approved):

ValentinBoudevin:
Doesn't work as expected for me in the OpenVex File. 
Indeed the justification is removed in the UI but the OpenVex database still contains it which is incorrect. 

Should be removed from the database too

ValentinBoudevin (changes requested):
Change code to remove database Openvex entry

PR #162, opened 2025-12-10
VulnScout rename "output" folder to "state"

ValentinBoudevin:

* Rename "output" to "state" in .vulnscout sub-folders

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Run VulnScout example and see a "state" older is created instead of the old "output" one

### DIFF
--- a/vulnscout.sh
+++ b/vulnscout.sh
@@ -20,7 +20,7 @@ show_help() {
   echo "Mandatory argument:"
   echo "  --name <project_name> name of the sub folder entry in .vulnscout/"
   echo ""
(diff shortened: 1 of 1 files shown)


rehsfl:
I have 2 remarks:
1. The /state is only updated in some files, there are still occurences of usage of /output in many parts of the application (especially merger_ci) that might still need modification.
2. Is there any mechanism to ensure that users of legacy VS versions don't end up with broken resource paths (especially OpenVEX assessment files) after an update?

PR #187, opened 2026-02-16
Remove start-example.sh

rawaha-e:

* Remove start-example.sh

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Use vulnscout.sh as the new entry point for VulnScout. Observe example usage in documentation.

## Pull Request Checklist


ValentinBoudevin (changes requested):
LGTM for the `start-example.sh` removal. 

But, I think it doesn't make sense to preserve both `example`, and `example-spdx3`.

Remove `example`, move `example-spdx3` to `example`, and remove the input folder. 

Should look like this: 

```
vboudevin@vboudevin-pc:~/Documents/vulnscout/.vulnscout$ ls
cache  docker-dev-override.yml  example  templates
vboudevin@vboudevin-pc:~/Documents/vulnscout/.vulnscout$ tree example/
example/
├── core-image-minimal-qemux86-64.rootfs.json
└── core-image-minimal-qemux86-64.rootfs.spdx.json
```

No more pre-defined YAML file and redundant example. 
Also involve a README update

rawaha-e:
> I would argue against removing example entirely. It contains the only SPDX2 files in the repository, and it would be nice to use them in some tests (actually, I make use of them in #190). Maybe renaming it to `example-spdx2` is better?

The idea was to use vulnscout.sh directly, with proper documentation examples for both SPDX2 and SPDX3 in order to maintain a consistent entry point command.

We might need to add new verification into vulnscout.sh to properly plug each file extensions to the right docker binding.

Issue #191, opened 2026-02-18
vulnscout.sh does not raise a file not found with incorrect filepaths


ValentinBoudevin:
How to reproduce: 

```
./vulnscout.sh --name test_nowhere --spdx $(pwd)/path_to_nowhere/example.rootfs.spdx.tar.zst --cve-check $(pwd)/path_to_nowhere/examp
le.rootfs.json
```

As a result, the container is starting but the arguments doesn't lead to a valid file:

```
Docker Compose command found: docker compose
Creating: Directory '/home/vboudevin/Documents/vulnscout/.vulnscout/test_nowhere'
Vulnscout Succeed: Docker Compose file set at /home/vboudevin/Documents/vulnscout/.vulnscout/test_nowhere/docker-test_nowhere.yml
latest: Pulling from sflinux/vulnscout
Digest: sha256:bbd3deeac7a9bb06b2cbfbfdd97acdd1aa16221e4004e37e7860fa8dd2fee2f6
Status: Image is up to date for sflinux/vulnscout:latest
docker.io/sflinux/vulnscout:latest
[+] Running 2/2
 ✔ Network test_nowhere_default  Created                                                                                                                                      0.0s 
 ✔ Container vulnscout           Created                                                                                                                                      0.1s 
Attaching to vulnscout
vulnscout  | VulnScout v0.9.0-85-gb200f5a
vulnscout  | Step (0/8): Server started
vulnscout  | Step (1/8): Extracting tar files
vulnscout  | Step (2/8): Searching SPDX JSON files
vulnscout  | Step (2/8): Merging 0 SPDX files
```

Expected: 

```
File not found for --spdx: $(pwd)/path_to_nowhere/example.rootfs.spdx.tar.zst --cve-check
File not found for --cve-check: $(pwd)/path_to_nowhere/example.rootfs.json
```

PR #194, opened 2026-02-19
vulnscout.sh: Arguments validity check & script cleanup

ValentinBoudevin:

Resolve the github issue: https://github.com/savoirfairelinux/vulnscout/issues/191

Now, an incorrect file provided in `--cve-check` or `--spdx` will raise a file not found error.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

```
./vulnscout.sh --name test_nowhere --spdx $(pwd)/path_to_nowhere/example.rootfs.spdx.tar.zst --cve-check $(pwd)/path_to_nowhere/example.rootfs.json
```

It triggers a file not found correctly

### DIFF
--- a/vulnscout.sh
+++ b/vulnscout.sh
@@ -62,6 +62,46 @@ show_help() {
   echo "  --stop            Stop running VulnScout container"
 }
 
+require_value(){
+    local flag="$1"
+    local value="$2"
+
+    if [[ -z "$value" || "$value" =~ ^-- ]]; then
+        echo "Error: $flag requires a value"
+        exit 1
+    fi
+}
+
+require_directory() {
+    local flag="$1"
+    local path="$2"
+
+    if [[ -z "$path" || "$path" =~ ^-- ]]; then
+        echo "Error: $flag requires a value"
+        exit 1
+    fi
+
+    if [[ ! -d "$path" ]]; then
+        echo "Error: Directory not found for $flag: $path"
+        exit 1
+    fi
+}
+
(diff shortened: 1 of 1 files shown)


QuentinL-SFL (changes requested):
Does not work on my side, the script don't give me the error but does not launch.

QuentinL-SFL (dismissed):
LGTM

rawaha-e (changes requested):
We need to check existing file for `openvex`, `cdx` and `report-template` as well.A

Also wouldn't it be better in this case to have a function to handle the checking in a centralized manner?

florentsfl (changes requested):

vulnscout.sh
```diff
@@ -113,174 +153,126 @@ fi
...
     --name)
-      if [[ -n "$2" && ! "$2" =~ ^-- ]]; then
-        VULNSCOUT_ENTRY_NAME="$2"
-        shift 2
-      else
-        echo "Error: --name requires a value"
-        exit 1
-      fi
-      ;;
+      require_value "$1" "${2:-}"
```
florentsfl: same here you can remove the default value for $2, it will pass an empty string if $2 does not exist:

`require_value "$1" "$2"`

florentsfl: my bad

florentsfl (approved):

rawaha-e (approved):

Issue #196, opened 2026-02-20
Mismatched env var name for FastSPDX


SkytAsul:
I want to use FastSPDX in the "Merging SPDX files" step of the `scan.sh` script. This is because, when having a large amount of big SPDX 2.2 files, the default SPDX merging algorithm can take _days_ to finish (see https://github.com/spdx/tools-python/issues/818).

If you want to use FastSPDX for parsing SPDX 2.2 files, you have to enable the `IGNORE_PARSING_ERRORS` variable, which is a rather weird name: https://github.com/savoirfairelinux/vulnscout/blob/0c50220c4ec6cde82b2b2f979baf24494e3bd742/src/bin/spdx_merge.py#L28-L30

The variable is used further down, at https://github.com/savoirfairelinux/vulnscout/blob/0c50220c4ec6cde82b2b2f979baf24494e3bd742/src/bin/spdx_merge.py#L50-L55

Maybe using FastSPDX is necessary for being able to ignore parsing errors, but in this case it would be nice to have another environment variable (like `USE_FAST_SPDX`) so we can use FastSPDX without ignoring parsing errors.

PR #201, opened 2026-02-24
 feat: added USE_FAST_SPDX variable to control the SPDX parsing algorithm

SkytAsul:

* cleaned boolean environment variable snippets such as `os.getenv(..., 'false') == 'true'` all over the place for a more flexible method handling casing and 0/1
* updated tests for this change
* added USE_FAST_SPDX variable to control the SPDX parsing algorithm
* **BREAKING:** setting IGNORE_PARSING_ERRORS to true no longer automatically uses the FastSPDX algorithm.
* also added --use-fast-spdx flag to vulnscout.sh

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Start a vulnscout scan with a lot of SPDX files with and without --use-fast-spdx and see the speedup.

### Additional notes

IGNORE_PARSING_ERRORS was controlling 2 things: ignoring the errors AND using the FastSPDX algorithm. Now it only has the parsing error responsibility, so it's clearer.

### Related Issue

#196 

## Pull Request Checklist

  - Cannot run Docker tests since I'm using Podman. The rest is passing.

### DIFF
--- a/src/bin/merger_ci.py
+++ b/src/bin/merger_ci.py
@@ -212,7 +212,7 @@ def read_inputs(controllers):
                   + " CycloneDX files, this mean there was an issue when analysing them.")
 
     use_fastspdx = False
-    if get_bool_env('IGNORE_PARSING_ERRORS'):
+    if get_bool_env('USE_FAST_SPDX'):
         use_fastspdx = True
         verbose("spdx_merge: Using FastSPDX parser")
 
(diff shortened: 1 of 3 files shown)


SkytAsul:
From what I've seen so far, I think we don't need any of the fancy parsing features used in the "default" algorithm. Only the most basic informations in the SPDX files are needed, and I think all of those are parsed with FastSPDX.

What we would lack with FastSPDX are error messages if parsed files have problems inside them but outside of the data we actually use.

PR #207, opened 2026-02-27
Use YAML_REQUIRES_UPDATE to track yaml-related option changes that should rebuild the yaml file 

WilsonZiweiWang:

Add YAML_REQUIRES_UPDATE flag to track changes in YAML-related options and to determine whether to reuse the existing ymal file or not.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

### Additional notes

### Related Issue

## Pull Request Checklist


ValentinBoudevin:
> > Instead of adding counter for every argument we could reuse the existing structure with something like `if { [ "$#" -eq 2 ] && [ -n "$VULNSCOUT_ENTRY_NAME" ] && [ -f "$YAML_FILE" ]; } || { [ "$#" -eq 3 ] && [ -n "$VULNSCOUT_ENTRY_NAME" ] && [ "$VULNSCOUT_DEV_MODE" = "true" ] && [ -f "$YAML_FILE" ]; } `
> 
> So you only consider two cases: `--name .. --dev` and `--name ...`. Any additional flags make the command not work as intended (If I understand the original intention correctly), for example, `--detach`, which shouldn't trigger a compose file rebuild. The current approach is more robust and doesn't depend on the argument count but on the flags that are actually related to the YMAL file.

I agree, `--detach` should not involve a YAML reconstruction.
LGTM

rawaha-e:
> > Instead of adding counter for every argument we could reuse the existing structure with something like `if { [ "$#" -eq 2 ] && [ -n "$VULNSCOUT_ENTRY_NAME" ] && [ -f "$YAML_FILE" ]; } || { [ "$#" -eq 3 ] && [ -n "$VULNSCOUT_ENTRY_NAME" ] && [ "$VULNSCOUT_DEV_MODE" = "true" ] && [ -f "$YAML_FILE" ]; } `
> 
> So you only consider two cases: `--name .. --dev` and `--name ...`. Any additional flags make the command not work as intended (If I understand the original intention correctly), for example, `--detach`, which shouldn't trigger a compose file rebuild. The current approach is more robust and doesn't depend on the argument count but on the flags that are actually related to the YMAL file.

This could be mitigated by putting all use cases in the conditions, this will keep a maintainable code.

```
if { [ "$ORIGINAL_ARGC" -eq 2 ] && [ -n "$VULNSCOUT_ENTRY_NAME" ] && [ -f "$YAML_FILE" ]; } || \
   { [ "$ORIGINAL_ARGC" -eq 3 ] && [ -n "$VULNSCOUT_ENTRY_NAME" ] && [ "$VULNSCOUT_DEV_MODE" = "true" ] && [ -f "$YAML_FILE" ]; } || \
   { [ "$ORIGINAL_ARGC" -eq 3 ] && [ -n "$VULNSCOUT_ENTRY_NAME" ] && [ "$VULNSCOUT_DETACH_MODE" = "true" ] && [ -f "$YAML_FILE" ]; } || \
   { [ "$ORIGINAL_ARGC" -eq 4 ] && [ -n "$VULNSCOUT_ENTRY_NAME" ] && [ "$VULNSCOUT_DEV_MODE" = "true" ] && [ "$VULNSCOUT_DETACH_MODE" = "true" ] && [ -f "$YAML_FILE" ]; } ; then
```

PR #271, opened 2026-04-09
feat: scan management commands

youennlj-sfl:

* Properly type hint the models for Scan, Variant and Project
* Add `--list-scans` command with 
* Add `--list-projects` command logic (was present in entrypoint but not implemented)
  * The 2 commands above accept an optional `--json` flag to output in JSON so it can be used in scripts
* Add `--delete-scan` command to drop a scan

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

## Pull Request Checklist

### DIFF
--- a/src/entrypoint.sh
+++ b/src/entrypoint.sh
@@ -47,6 +47,7 @@ Scan & output commands:
   --export-custom-assessments  Export custom (review) assessments as tar.gz to /scan/outputs/
   --import-custom-assessments <path>  Import custom assessments from .json or .tar.gz
   --match-condition <expr>  Exit code 2 if condition met (e.g. "cvss >= 9.0")
+  --delete-scan <id>        Delete a past scan by its ID
 
 Data retrieval commands:
   --list-projects           List all projects and their variants
@@ -426,6 +427,11 @@ cmd_do_get_data() {
     flask --app src.bin.webapp "${DATA_REQUESTED#--}" "${data_args[@]}"
 }
 
+cmd_delete_scan() {
+    scan_id="$1"
+    flask --app src.bin.webapp delete-scan "$scan_id"
+}
+
 cmd_daemon() {
     setup_user
     echo "VulnScout ready. Use '/scan/src/entrypoint.sh --help' for available commands."
@@ -553,6 +559,8 @@ while [[ $# -gt 0 ]]; do
             GRYPE_SCAN_REQUESTED=true; SCAN_REQUIRED=true; shift ;;
         --clear-inputs)
             cmd_clear_inputs; shift ;;
+        --delete-scan)
+            cmd_delete_scan "$2"; shift 2 ;;
         --serve)
             if [[ -n "$MATCH_CONDITION" ]]; then
                 echo "Error: --serve and --match-condition are incompatible."; exit 1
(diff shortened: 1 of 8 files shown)


src/entrypoint.sh
```diff
@@ -508,6 +538,8 @@ while [[ $# -gt 0 ]]; do
             PROJECT_NAME="$2"; shift 2 ;;
         --variant)
             VARIANT_NAME="$2"; shift 2 ;;
+        --json)
```
rawaha-e, line 541: Make sure to verify a Data Retrieval arg is provided because --json only applies in such case.

src/entrypoint.sh
```diff
@@ -527,6 +559,8 @@ while [[ $# -gt 0 ]]; do
             GRYPE_SCAN_REQUESTED=true; SCAN_REQUIRED=true; shift ;;
         --clear-inputs)
             cmd_clear_inputs; shift ;;
+        --delete-scan)
+            cmd_delete_scan "$2"; shift 2 ;;
```
rawaha-e, line 563: User can provide `--delete-scan` with empty value, verify $2 before calling `cmd_delete_scan`.
We could either verify this in `entrypoint` or let `@click.argument` throw an error (which could crash the app)

rawaha-e (changes requested):

src/entrypoint.sh
youennlj-sfl: There is no argument validation whatsoever in the rest of the entrypoint. I though the policy was "handle validation in `vulnscout`, entrypoint is for internal use and users should know what they are doing". If I am mistaken, then we should add validation for all commands.

src/entrypoint.sh
youennlj-sfl: Same answer as below.

src/entrypoint.sh
rawaha-e: @youennlj-sfl
As far as I understand, the implementation by @ValentinBoudevin  requires the args to be handled in both `vulnscout` and `entrypoint.sh`, so each command must be present in both scripts.
So the next step would probably be to add it to `vulnscout.sh` which handles verification then passes it into `entrypoint.sh`.

youennlj-sfl: I added the new args to the `vulnscout` script. However, there is no way to do argument validation with the argument parsing method used (raw handling of shell arguments). The rest of the commands don't do it, or they only do it when they handle files (easy since they can test that the files exist).
Anyway, the Python program will output `ValueError: Scan not found.` if the argument is incorrect so there is no real issue.

florentsfl:
Maybe add a "why" this changes were proposed ? What is the use case of using the CLI to list or delete a scan ?

youennlj-sfl:
> Maybe add a "why" this changes were proposed ? What is the use case of using the CLI to list or delete a scan ?

In a CI. I want to cache the database between runs so VulnScout does not have to fetch again and again information from Internet databases, but I don't want the scan information to leak between runs.
Therefore, the CI will do a list command chained to a delete command in order to delete scans once they have been done and reports generated.
This was suggested by @ValentinBoudevin and it suits my needs

florentsfl:
LGTM

florentsfl (approved):

rawaha-e (approved):
LGTM

PR #277, opened 2026-04-10
Templates: add new fields after DB migration

rawaha-e:

* Add new models to report templates

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Create `.vulnscout/templates/test.adoc` with this content:

```
Projects: {{ projects | length }}
{% for p in projects %}
- {{ p.name }} ({{ p.id }})
{% endfor %}

Variants: {{ variants | length }}
{% for v in variants %}
- {{ v.name }} -> project {{ v.project_id }}
{% endfor %}

Scans: {{ scans | length }}
{% for s in scans | sort_by_scan_date %}
- {{ s.timestamp | print_iso8601 }}: {{ s.description }}
{% endfor %}

SBOM Documents: {{ sbom_documents | length }}
{% for d in sbom_documents %}
- {{ d.source_name }} ({{ d.format }}): {{ d.path }}
{% endfor %}

Variants for first project: {{ variants | filter_by_project(projects[0].id) | length }}
```

Then generate it:
```bash
./vulnscout --project test --variant x86 --report test.adoc
```

Check `.vulnscout/outputs/test.adoc`

## Pull Request Checklist

### DIFF
--- a/src/views/templates.py
+++ b/src/views/templates.py
@@ -64,21 +64,22 @@ class Templates:
         kwargs["assessments"] = {}
 
         if self.projectsCtrl is not None:
-            kwargs["projects"] = self.projectsCtrl.serialize_list(self.projectsCtrl.get_all())
+            kwargs["projects"] = {p["id"]: p for p in self.projectsCtrl.serialize_list(self.projectsCtrl.get_all())}
         else:
-            kwargs["projects"] = []
+            kwargs["projects"] = {}
         if self.variantsCtrl is not None:
-            kwargs["variants"] = self.variantsCtrl.serialize_list(self.variantsCtrl.get_all())
+            kwargs["variants"] = {v["id"]: v for v in self.variantsCtrl.serialize_list(self.variantsCtrl.get_all())}
         else:
-            kwargs["variants"] = []
+            kwargs["variants"] = {}
         if self.scansCtrl is not None:
-            kwargs["scans"] = self.scansCtrl.serialize_list(self.scansCtrl.get_all())
+            kwargs["scans"] = {s["id"]: s for s in self.scansCtrl.serialize_list(self.scansCtrl.get_all())}
         else:
-            kwargs["scans"] = []
+            kwargs["scans"] = {}
         if self.sbomDocumentsCtrl is not None:
-            kwargs["sbom_documents"] = self.sbomDocumentsCtrl.serialize_list(self.sbomDocumentsCtrl.get_all())
+            all_docs = self.sbomDocumentsCtrl.serialize_list(self.sbomDocumentsCtrl.get_all())
+            kwargs["sbom_documents"] = {d["id"]: d for d in all_docs}
         else:
-            kwargs["sbom_documents"] = []
+            kwargs["sbom_documents"] = {}
 
         filter_date = None
         if "ignore_before" in kwargs and kwargs["ignore_before"] != "1970-01-01T00:00":
@@ -126,15 +127,15 @@ class Templates:
         else:
             kwargs["assessments"] = kwargs["unfiltered_assessments"]
 
-        scan_by_id = {s["id"]: s for s in kwargs["scans"]}
-        doc_by_id = {d["id"]: d for d in kwargs["sbom_documents"]}
-        variant_by_id = {v["id"]: v for v in kwargs["variants"]}
+        scan_by_id = kwargs["scans"]
+        doc_by_id = kwargs["sbom_documents"]
+        variant_by_id = kwargs["variants"]
 
-        for doc in kwargs["sbom_documents"]:
+        for doc in kwargs["sbom_documents"].values():
             scan = scan_by_id.get(doc["scan_id"])
             doc["variant_id"] = scan["variant_id"] if scan else None
 
-        for doc in kwargs["sbom_documents"]:
+        for doc in kwargs["sbom_documents"].values():
             doc["packages"] = {}
             for sbom_pkg in SBOMPackage.get_by_document(doc["id"]):
                 pkg_id = sbom_pkg.package.string_id
@@ -143,14 +144,14 @@ class Templates:
 
         pkg_to_docs: dict = {}
         pkg_to_variants: dict = {}
-        for doc in kwargs["sbom_documents"]:
+        for doc in kwargs["sbom_documents"].values():
             for pkg_id in doc["packages"]:
                 pkg_to_docs.setdefault(pkg_id, []).append(doc["id"])
                 if doc["variant_id"]:
                     pkg_to_variants.setdefault(pkg_id, set()).add(doc["variant_id"])
 
         for pkg_id, pkg in kwargs["packages"].items():
-            pkg["sbom_documents"] = [doc_by_id[d] for d in pkg_to_docs.get(pkg_id, []) if d in doc_by_id]
+            pkg["sbom_documents"] = {d: doc_by_id[d] for d in pkg_to_docs.get(pkg_id, []) if d in doc_by_id}
             pkg["variants"] = list(pkg_to_variants.get(pkg_id, set()))
             pkg["vulnerabilities"] = {}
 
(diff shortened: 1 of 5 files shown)


vulnscout
```diff
@@ -220,9 +220,12 @@ EOF
         env_args+=(-e "${key}=${value}")
     done < "$VULNSCOUT_CONFIG_FILE"
     [[ "$LEGACY_SETUP_DETECTED" == true ]] && env_args+=(-e LEGACY_SETUP_DETECTED=true)
+    local VULNSCOUT_TEMPLATES_DIR="${VULNSCOUT_BUILD_DIR}/templates"
+    mkdir -p "$VULNSCOUT_TEMPLATES_DIR"
     local -a vol_args=(
         -v "$VULNSCOUT_CACHE_DIR":/cache/vulnscout:Z
         -v "$VULNSCOUT_OUTPUTS_DIR":/scan/outputs:Z
+        -v "$VULNSCOUT_TEMPLATES_DIR":/scan/.vulnscout/templates:Z
         -v "$VULNSCOUT_CONFIG_FILE":/etc/vulnscout/config.env:ro,Z
```
youennlj-sfl: Why are you doing those changes? If you want to use a template that is not built-in, you can use `--report <path to file>` and the template file will be copied inside the container.

src/views/templates.py
```diff
@@ -47,6 +51,24 @@ def render(self, template_name, **kwargs):
...
+            kwargs["variants"] = self.variantsCtrl.serialize_list(self.variantsCtrl.get_all())
+        else:
+            kwargs["variants"] = []
+        if self.scansCtrl is not None:
+            kwargs["scans"] = self.scansCtrl.serialize_list(self.scansCtrl.get_all())
+        else:
+            kwargs["scans"] = []
+        if self.sbomDocumentsCtrl is not None:
+            kwargs["sbom_documents"] = self.sbomDocumentsCtrl.serialize_list(self.sbomDocumentsCtrl.get_all())
+        else:
```
youennlj-sfl, line 81: Once the database gets pretty big, wouldn't this cause massive slowdown and huge memory usage?

vulnscout
rawaha-e: The `--report` is is indeed properly handled in non-interactive mode.
The issue is specifically with the web dashboard: when you go to the Export tab then check custom reports, they won't show up.
We have 2 possible ways of handling the web dashboard case (@ValentinBoudevin):
1. Mount the entire templates folder and show all available reports in the web dashboard (which I did here). The rationale behind this is that all operations can be done through the web dashboard: managing SBOM, variants, etc... so why not templates, we can imagine a feature when we can upload templates there as well later.
2. Force  the user to use `--report` to see their report in the web dashboard.

youennlj-sfl: Managing SBOM via the web dashboard asks users to upload the SBOM files to the web dashboard and then it creates the file in the container. It does not rely on mounting a directory in the container nor on using the --add-spdx/--add-blablabla commands. I think report management should be the same, otherwise it would get confusing.
I fear that if we re-start mounting directories here and there in the containers it will become the same mess as it was before the entrypoint refactor :sweat_smile:

src/views/templates.py
rawaha-e: As far as the attributes we are working on are concerned, no because they are mostly metadata contained within a fraction of the database.
The real issue is when users will extract packages / vulns with very narrowed filters, which is not the scope of this PR.

vulnscout
rawaha-e: In such case we propose to either disable the custom templates in the web dashboard that is currently broken or implement a proper solution to it in this PR. This is a decision to be taken.

youennlj-sfl: Yes maybe a tooltip on the "custom templates" tab that explains that is it currently broken and that gives a hint towards the `--report` command to add a template file.

youennlj-sfl:
For my CI on SEAPATH, I will need some metadata on the Vulnerability object that you did not add in this PR: the names of the variants that contains the vulnerabilities. Do you think it's possible to add it?

rawaha-e:
> For my CI on SEAPATH, I will need some metadata on the Vulnerability object that you did not add in this PR: the names of the variants that contains the vulnerabilities. Do you think it's possible to add it?

@youennlj-sfl 
I was working on it, I added an open data model graph so you could traverse many more data relationships, including ability to see variants of each vuln or package. For example:
```
{% set variant_map = {} %}
{% for v in variants %}{% set _ = variant_map.update({v.id: v.name}) %}{% endfor %}

{% for vuln in vulnerabilities.values() %}
  {{ vuln.id }}: {% for vid in vuln.variant_ids %}{{ variant_map[vid] }}{% endfor %}
{% endfor %}
```
Most of the times the ID is exposed then you can map any of its attribute like in the example.

youennlj-sfl:
@rawaha-e your solution does not work: `Warning: could not generate 'cve-check-fail-template.txt.j2': access to attribute 'update' of 'dict' object is unsafe.`
After searching the web and asking an AI, it seems like there is no easy way to generate a dictionary from within a Jinja template.
I would suggest making the `variants` global variable a dictionary that maps a variant ID to its details, just like how it is made for the previously existing variables (vulnerabilities, packages, assessments)

There was another bug with missing controllers from the merger_ci file, I fixed it myself. But TBH the current way to manage controllers (passing a raw dictionary in a weird dependency injection manner) is really bad, it should be refactored at some point.

rawaha-e:
@youennlj-sfl  It was working for me because I disabled jinja2 sandbox on my environment (which is not a good practice) so data could be mutated, I will add a global variant variable soon.

For the missing controllers, I didn't catch it as I was mostly testing through the web dashboard which was working, thank you for fixing it!
The older controllers management was relevant when working with in-memory data that we used before moving to DB, so we had to grab the whole controller in order to access its data. We can have a separate PR to call SQLAlchemy queries instead.

src/views/templates.py
```diff
@@ -47,6 +61,24 @@ def render(self, template_name, **kwargs):
...
+            kwargs["variants"] = {}
+        if self.scansCtrl is not None:
+            kwargs["scans"] = self.scansCtrl.serialize_list(self.scansCtrl.get_all())
+        else:
+            kwargs["scans"] = []
+        if self.sbomDocumentsCtrl is not None:
+            kwargs["sbom_documents"] = self.sbomDocumentsCtrl.serialize_list(self.sbomDocumentsCtrl.get_all())
+        else:
+            kwargs["sbom_documents"] = []
```
youennlj-sfl: The behavior is inconsistent here: some properties are list, some others are dictionaries. I think it would be better that all of them are dictionaries indexed by the UUIDs. This gives easy indexing and users that want a list can easily get one using `| to_list`.

youennlj-sfl (changes requested):

youennlj-sfl:
>  It was working for me because I disabled jinja2 sandbox on my environment (which is not a good practice) so data could be mutated, I will add a global variant variable soon.

I see! Thanks for your global variable, it works now. I just made a comment so its consistent with the rest of the variables.

> The older controllers management was relevant when working with in-memory data that we used before moving to DB, so we had to grab the whole controller in order to access its data. We can have a separate PR to call SQLAlchemy queries instead.

OK thanks for the background information. It would definitely be nice to get rid of those things in a future PR.

I'll approve once you make the change to the variables and maybe revert the directory mounting thing?

youennlj-sfl:
@rawaha-e you still have the change to `vulnscout` where you mount the templates directory in the container that I don't agree with (see the conversation above)

rawaha-e:
> @rawaha-e you still have the change to `vulnscout` where you mount the templates directory in the container that I don't agree with (see the conversation above)

@youennlj-sfl I dropped the corresponding commit.

ValentinBoudevin (approved):
LGTM

had to modify the .adoc due to an error `Warning: could not generate 'test.adoc': 'str' object has no attribute 'get'`

```
Projects: {{ projects | length }}
{% for p in projects | as_list %}
- {{ p.name }} ({{ p.id }})
{% endfor %}

Variants: {{ variants | length }}
{% for v in variants | as_list %}
- {{ v.name }} -> project {{ v.project_id }}
{% endfor %}

Scans: {{ scans | length }}
{% for s in scans | sort_by_scan_date %}
- {{ s.timestamp | print_iso8601 }}: {{ s.description }}
{% endfor %}

SBOM Documents: {{ sbom_documents | length }}
{% for d in sbom_documents | as_list %}
- {{ d.source_name }} ({{ d.format }}): {{ d.path }}
{% endfor %}

Variants for first project: {{ variants | filter_by_project((projects | as_list)[0].id) | length }}
```

Added a commit to fix it

youennlj-sfl (approved):

Issue #276, opened 2026-04-10
SPDX to OpenVEX generation collapses distinct BOM components into ambiguous product identities


AenganZ:
While building a BOM-VEX validator / validation workflow, I tested VulnScout with an SPDX BOM and the generated OpenVEX output. The documents are syntactically valid, but the validator found that multiple distinct BOM components are collapsed into weaker VEX product identities, which makes cross-document matching ambiguous.

I am attaching the exact files used:

[vulscout_openvex.json](https://github.com/user-attachments/files/26619712/vulscout_openvex.json)
[vulscout_spdx.json](https://github.com/user-attachments/files/26619711/vulscout_spdx.json)

From the attached files, the SPDX BOM contains multiple distinct packages that share the same `name + version`, but the generated OpenVEX output represents them with the same primary product identity such as `name@version`.

Examples:

- In the SPDX BOM, `perl` version `5.26.1` appears more than once as distinct packages.
- In the OpenVEX output, they are represented as:
  - `perl@5.26.1`

A stronger example is `util-linux`:

- In the SPDX BOM, `util-linux` version `2.31.1` appears multiple times as distinct packages.
- In the OpenVEX output, they are collapsed into:
  - `util-linux@2.31.1`

From a BOM-VEX validation perspective, this weakens target identity and can make the VEX product ambiguous with respect to the source BOM.

I also noticed that stronger identifiers such as PURLs exist in the OpenVEX output under `identifiers.purl`, but the main `products[].@id` value is still the weaker `name@version` string. This makes cross-document matching harder than necessary.

Expected behavior:
If the source SPDX BOM contains multiple distinct components with the same `name + version`, the generated OpenVEX output should preserve enough identifier strength to keep those components distinguishable. At minimum, the primary VEX product identity should not collapse multiple BOM components into the same `name@version` value when stronger identifiers already exist in the source BOM or in the generated VEX metadata.

Steps to reproduce:
1. Use an SPDX BOM containing multiple distinct packages that share the same `name + version`.
2. Generate OpenVEX output with VulnScout from that SPDX BOM.
3. Compare the package identities in `vulscout_spdx.json` with the product identities in `vulscout_openvex.json`.
4. Observe that distinct BOM components are collapsed into the same OpenVEX product identity such as `perl@5.26.1` or `util-linux@2.31.1`.

Here, the problem is not document parsing, but loss of target identity strength during SPDX -> OpenVEX generation.

AenganZ:
[...]

PR #283, opened 2026-04-15
fix: PURL as OpenVEX product @id, not name@version

ValentinBoudevin:

Reference: https://github.com/savoirfairelinux/vulnscout/issues/276.

When multiple distinct BOM components share the same name and version (e.g. perl@5.26.1, util-linux@2.31.1), the generated OpenVEX output collapsed them into the same product identity, making cross-document BOM-VEX matching ambiguous. Use the PURL (already present in identifiers) as the primary product @id to preserve target uniqueness.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

1- Checkout on this PR and build the corresponding docker image

```
docker build -t sfl:fix_openvex .
```

2- Export the `VULNSCOUT_IMAGE` variable: 

```
export VULNSCOUT_IMAGE="sfl:fix_openvex"
```

3- Export a project in `OpenVEX` format. Now the `@id` value correspond to the `purl` not `package@version`

### DIFF
--- a/src/routes/assessments.py
+++ b/src/routes/assessments.py
@@ -203,11 +203,12 @@ def init_app(app):
                             name, version = pkg_str.rsplit("@", 1)
                         else:
                             name, version = pkg_str, ""
+                        purl = f"pkg:generic/{name}@{version}"
                         products.append({
-                            "@id": pkg_str,
+                            "@id": purl,
                             "identifiers": {
                                 "cpe23": f"cpe:2.3:*:*:{name}:{version}:*:*:*:*:*:*:*",
-                                "purl": f"pkg:generic/{name}@{version}",
+                                "purl": purl,
                             }
                         })
                     stmt["products"] = products
--- a/src/views/fast_spdx.py
+++ b/src/views/fast_spdx.py
@@ -4,6 +4,9 @@
 # SPDX-License-Identifier: GPL-3.0-only
 
 from ..models.package import Package
+from ..controllers.packages import PackagesController
+from ..controllers.vulnerabilities import VulnerabilitiesController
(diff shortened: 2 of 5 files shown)


QuentinL-SFL (dismissed):
LGTM

youennlj-sfl:
This does not seem to fix user's problem.
If I put the file the user provided in https://github.com/savoirfairelinux/vulnscout/issues/276#issuecomment-4233577175 (spdx_purl.json), run a Grype scan then export an OpenVex file, I still see the generic PURLs.

It looks like the PURLs are not parsed and stored in the DB in the first place.

youennlj-sfl:
After research, it looks like the SPDX merger code does not use the data found in the `externalRefs` part of the SPDX2 files (both for the "fast" parser and for the normal one). I'll try to fix that.

youennlj-sfl:
With those commits we can reproduce the user's problem and actually see the resolution.

But I think this PR still does not fix the problem as a whole:
> Expected behavior:
If the source SPDX BOM contains multiple distinct components with the same name + version, **the generated OpenVEX output should preserve enough identifier strength to keep those components distinguishable.** At minimum, the primary VEX product identity should not collapse multiple BOM components into the same name@version value when stronger identifiers already exist in the source BOM or in the generated VEX metadata.

Right now, you only export the 1st PURL, so in the OpenVEX there is only one package per (name, version) tuple, even if in the original SBOM there was multiple packages for the same tuple.

QuentinL-SFL (approved):
LGTM

ValentinBoudevinSFL (approved):
To be merged

PR #287, opened 2026-04-16
Fix: db lock after few assessments due to NVD taking control of the DB

ValentinBoudevin:

*fix: serialize SQLite writes with priority lock
*fix: remove patch-finder route usage

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Wait for the NVD/EPSS loading to start and try to assess 10000 cves in one shot several times. Should not crash.

### DIFF
--- a/src/routes/__init__.py
+++ b/src/routes/__init__.py
@@ -6,7 +6,6 @@ from .packages import init_app as init_pkg_app
 from .vulnerabilities import init_app as init_vuln_app
 from .assessments import init_app as init_assess_app
 from .documents import init_app as init_doc_app
-# from .patch_finder import init_app as init_patch_finder_app  # disabled — feature on standby
 from .nvd_progress import init_app as init_nvd_progress_app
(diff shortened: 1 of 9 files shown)


src/routes/__init__.py
```diff
@@ -6,7 +6,7 @@
 from .vulnerabilities import init_app as init_vuln_app
 from .assessments import init_app as init_assess_app
 from .documents import init_app as init_doc_app
-from .patch_finder import init_app as init_patch_finder_app
+# from .patch_finder import init_app as init_patch_finder_app  # disabled — feature on standby
```
QuentinL-SFL: For me we should not leave code as comment even if it only temporary

ValentinBoudevinSFL: done

QuentinL-SFL (approved):
LGTM

youennlj-sfl (approved):

Issue #286, opened 2026-04-16
Unsaved modifications passed between CVEs


QuentinL-SFL:
When you modified a CVE but do not save the modifications and try to change CVE you have a pop saying your unsaved modification will be lost. 
But even if you click on "Yes, navigate" the unsaved modification are passed on the next CVE. 
It should not be the case.

<img width="1701" height="831" alt="Image" src="https://github.com/user-attachments/assets/3959d1b3-bce3-4d0a-96f1-d51bf5387e0d" />
<img width="1701" height="831" alt="Image" src="https://github.com/user-attachments/assets/6273859f-9222-4598-bb3d-f94f2b709b4d" />
<img width="1701" height="831" alt="Image" src="https://github.com/user-attachments/assets/00b38239-75af-43d8-aadd-a1d464c0d726" />

ValentinBoudevin:
Fix merged. Will be present in v0.12.1

PR #289, opened 2026-04-16
fix: frontend data clear when navigate

ValentinBoudevin:

Now, changing to another CVE correctly clear the data

Reference: https://github.com/savoirfairelinux/vulnscout/issues/286

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Start to assess a CVE, move to a next one, the data should be correctly cleared.

### DIFF
--- a/frontend/src/components/VulnModal.tsx
+++ b/frontend/src/components/VulnModal.tsx
@@ -668,6 +668,7 @@ type AssessmentGroup = {
 
     return (
         <div
+            key={vuln.id}
             data-testid="vuln-modal-backdrop"
             tabIndex={-1}
             onMouseDown={(event) => {


frontend/src/components/VulnModal.tsx
youennlj-sfl, line 1: Those changes misuse the Effects system of React. There is a documentation page explicitly describing how to reset fields when a property change: https://react.dev/learn/you-might-not-need-an-effect#resetting-all-state-when-a-prop-changes

Simply adding `key={vuln.id}` on the top-level div of this modal (line 670) will fix the issue without needing your other changes. https://github.com/savoirfairelinux/vulnscout/blob/44e372506765806dc1b67a3fbb69a2f701354282/frontend/src/components/VulnModal.tsx#L670-L675

youennlj-sfl (changes requested):

youennlj-sfl (approved):

rawaha-e (approved):
LGTM

PR #299, opened 2026-04-20
fix: Include UTC offset in serialized timestamps

ValentinBoudevin:

Timezone wasn't applied properly on imported documents.

SQLite does not preserve timezone info on datetime columns. When datetimes are read back from the database they become naive, so .isoformat() produces strings without "+00:00". JavaScript's new Date() then interprets them as local time instead of UTC, shifting displayed hours by the user's timezone offset.

Add a shared ensure_utc_iso() helper in src/helpers/datetime_utils.py that re-attaches the UTC offset to naive datetimes before serialization.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Create a new project, assessments should be with the correct hours/min based on the timezone (e.g. Visible on Last Updated column in Vulnerability tab) .

### DIFF
--- a/src/helpers/datetime_utils.py
+++ b/src/helpers/datetime_utils.py
@@ -1,0 +1,22 @@
+# Copyright (C) 2026 Savoir-faire Linux, Inc.
+# SPDX-License-Identifier: GPL-3.0-only
+
+from datetime import timezone
+
+
+def ensure_utc_iso(dt) -> str | None:
+    """Return an ISO 8601 string that always carries the UTC offset.
+
+    SQLite does not preserve timezone info, so datetimes read back from
+    the database are naive even though they were stored as UTC.  This
+    helper re-attaches ``+00:00`` when the offset is missing so that
+    consumers (e.g. JavaScript ``new Date()``) can interpret the value
+    correctly.
+    """
+    if dt is None:
+        return None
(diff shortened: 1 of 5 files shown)


src/helpers/datetime_utils.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2024 Savoir-faire Linux, Inc.
```
florentsfl: why 2024 ?

src/helpers/datetime_utils.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
```
florentsfl: the encoding is probably utf-8 by default ? is this required ?

src/helpers/datetime_utils.py
ValentinBoudevin: Wrong copy paste from the codebase. I lot of files have it. Should be removed

florentsfl (approved):

ValentinBoudevinSFL (commented):
Tbm

FadiSheh (approved):
lgtm

ValentinBoudevinSFL (approved):

PR #302, opened 2026-04-22
SBOM: add CPE/PURL columns

ValentinBoudevin:

* PURL are important for OSV Scanner and other tool. This information was already stored in the database but not used.
* CPE moved from a button to a dedicated column, which is more readable
* Add a new filter `Columns`

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Open web interface and see the new columns "CPE", and "PURL"


rawaha-e (changes requested):
`Popup.tsx` component (which rendered the CPE bubble) is no longer used. If we no longer need it, we can delete the file. This is different from the component used to render the vulnerability description bubble.

rawaha-e (dismissed):
LGTM

florentsfl:
is it relevant to show the columns by default ? it seems to give little information to the user since we already have the package name and the version

ValentinBoudevin:
Ok, Maybe PURL and CPE can be off be default as most of the Users won't pay attention to it

florentsfl (approved):
ok

ValentinBoudevinSFL (approved):

PR #303, opened 2026-04-22
fix: variant enrichment uses correct scan IDs

ValentinBoudevin:

The previous code computed "latest scan" per variant using only max(timestamp), which returned a single scan. When that scan was a tool scan (NVD/OSV), SBOM findings had no observations for it, so variants appeared empty. Now reuses the same multi-source scan IDs (current_scan_ids) already used to load vulnerabilities.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Vulnerabilities tab should display the `variant` column correctly. Without this fix, it was empty

### DIFF
--- a/src/routes/vulnerabilities.py
+++ b/src/routes/vulnerabilities.py
@@ -389,48 +389,41 @@ def init_app(app):
                 for v in vulns:
                     v["packages_current"] = list(v["packages"])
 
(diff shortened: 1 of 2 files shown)


rawaha-e (approved):
LGTM

florentsfl (approved):
works for me

PR #305, opened 2026-04-23
fix: Remove the unknown URL

ValentinBoudevin:

Links provided wrong "unokown" URL due to datasource is a metadata field (values like "unknown", "nvd:cpe", etc.) not a URL.

The code was incorrectly spreading it into the links list with [vuln.datasource, ...vuln.urls]. Since vuln.urls already contains all actual link URLs from the backend, the fix is to simply use vuln.urls directly.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Open any CVEs and you shouldn't see any `unknown` URL


rawaha-e (approved):
LGTM

florentsfl (approved):

PR #300, opened 2026-04-21
Scan: Rework of the feature to include NVD/OSV/GRYPE

ValentinBoudevin:
### Changes proposed in this pull request:

* Scan tab: Add new scan buttons for Grype/NVD/OSV scanner
* Scan tab: delete scans option
* CLI: add the commands `--perform-nvd-scan` and `--perform-osv-scan`
* Fix: SPDX3 persistence: Get CVSS/Description/Severity from the SPDX3 file if present

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Build a container with this changes:

```
docker build -t sfl:scan .
```

And export `VULNSCOUT_IMAGE` variable:

```
export VULNSCOUT_IMAGE=sfl:scan
```

Then laod a project and go to Scan tab and try to interact with it

PR #317, opened 2026-05-04
Use --project and --variant for import/export custom assessment

### LINKED ISSUE #315: import-custom-assessments command does not use project
The "import-custom-assessments" command does not use the `--project` option.
https://github.com/savoirfairelinux/vulnscout/blob/8beb6937d5deaaba25556bb46d44e5bb0b2f3b88/src/bin/merger_ci.py#L729-L734
It fetches all existing variants and creates a dictionary mapping variant names to variants. This is a wrong logic: if you have multiple projects with variants named the same (e.g. "default"), the dictionary will only contain one variant.

On the same note, I think that users should be able to specify a variant name directly using `--variant`. Right now, the variant is inferred from the file name which I think is way too restrictive. https://github.com/savoirfairelinux/vulnscout/blob/8beb6937d5deaaba25556bb46d44e5bb0b2f3b88/src/bin/merger_ci.py#L919-L928

I can do a PR to fix that

youennlj-sfl:

* refactored how the `--variant` flag is handled in the vulnscout and entrypoint scripts so the Python code can distinguish between `--variant default` and no `--variant` flag at all.
* added `--project` and `--variant` flags to `--import-custom-assessment`
  * `--project` is required
  * `--variant` is only allowed when importing a single JSON file and takes precedence over the file name to determine which variant to import assessments in
* added `--project` and `--variant` flags to `--export-custom-assessment`
  * `--project` is required
    * the command no longer exports the assessments for all created projects (and it is no longer possible to do so - it caused name clashes in the archive when multiple variants with the same name exist)
    * Potential regression: assessments that are not tied to any variant/project can no longer be exported. They were previously exported with the "unassigned" filename, but this is incompatible with the new filtering architecture
  * when using `--variant`, a single JSON file is exported with the variant name as filename
### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

## Pull Request Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,13 +5,16 @@ FROM node:20 AS buildfront
 
 RUN mkdir -p /frontend /src/static
 WORKDIR /frontend
-COPY frontend .
+
+# First install packages so the layer can be reused when code changes
+COPY frontend/package.json frontend/package-lock.json .
+RUN npm ci
 
 # Create build .env with API_URL set as blank. This way, fetch call are made to '/api/...' on same origin.
 RUN echo "VITE_API_URL=\"\"" > .env
 
-RUN npm ci && \
-    npm run build
+COPY frontend .
+RUN npm run build
 
 
 FROM alpine:3.20
--- a/doc/source/container-entrypoint.md
+++ b/doc/source/container-entrypoint.md
@@ -33,8 +33,8 @@ docker exec vulnscout /scan/src/entrypoint.sh --serve
 
 | Flag | Description |
 |------|-------------|
-| `--project <name>` | Project name for subsequent input commands (default: `default`) |
-| `--variant <name>` | Variant name for subsequent input commands (default: `default`) |
+| `--project <name>` | Project name for subsequent commands (default: `default`) |
+| `--variant <name>` | Variant name for subsequent commands (default: `default`) |
 
 ### Input Commands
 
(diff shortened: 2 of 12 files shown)


youennlj-sfl:
One big issue left: the whole import/export code is completely duplicated between [`src/bin/merger_ci.py` ](https://github.com/youennlj-sfl/vulnscout/blob/b7310e423991b6cfa931ed09d89a20afc3a09770/src/bin/merger_ci.py#L558-L1015) and [`src/routes/assessments.py`](https://github.com/youennlj-sfl/vulnscout/blob/b7310e423991b6cfa931ed09d89a20afc3a09770/src/routes/assessments.py#L119-L449). Their respective unit tests are mostly duplicates as well.
I have not updated the `src/routes/assessments.py` file yet because it annoyed me to have that much code duplicated. Couldn't we factor this somehow? @rawaha-e @ValentinBoudevin

youennlj-sfl:
The duplication mentioned above is a problem for the project as a whole. There is so much code duplicated for handling by the CLI/the web app.
For instance, I just noticed that the "delete scan" code is different between the CLI and the web app: the CLI only deletes the scan and does not take care of the orphaned findings, whereas the webapp does more job and removes the orphaned findings. https://github.com/savoirfairelinux/vulnscout/pull/300/ which added deletion for the web app did not update the CLI code...
To prevent those issues we could use the Controllers to factor this kind of code so the CLI/web app simply call the controllers methods and the core logic is handled there.

ValentinBoudevin (approved):
LGTM

rawaha-e (approved):
LGTM

PR #321, opened 2026-05-05
fix: scope Yocto Description to observation

### LINKED ISSUE #311: "Yocto Description" is global to all projects
As of now, the "yocto_description" field is stored in the Vulnerability model: https://github.com/savoirfairelinux/vulnscout/blob/8beb6937d5deaaba25556bb46d44e5bb0b2f3b88/src/models/vulnerability.py#L37
It is being added when scanning a Yocto cve-check file: https://github.com/savoirfairelinux/vulnscout/blob/8beb6937d5deaaba25556bb46d44e5bb0b2f3b88/src/views/yocto_vulns.py#L79-L80
The problem is that this description is being globally stored to the model and is thus shared with all projects and variants, even though the "yocto description" is local to a cve-check file, it is not intrinsically part of the vulnerability metadata.

In my very case, one of my Yocto projects reported CVE-2017-6264 with the Yocto description "vulnerable_code_not_in_execute_path: Android specific". I just created another project and added a different SBOM and did a Grype scan and this Yocto description still appears, whereas it is completely unrelated.

One suggestion might be that, instead of adding the Yocto description to the Vulnerability object, VulnScout could create an assessment with the Yocto description test inside? Since the assessment is scoped to a project/variant, it fixes the problem.

youennlj-sfl:
## Fixes #311

Based on #324

* Deleted the `texts` in-memory attribute of the Vulnerability model
  * the only text stored in the Vulnerability model is the vuln description, stored in `description`
  * (**breaking**) the `texts` attribute of the Vulnerability object in the templates no longer exists. Users should now use `vuln.description`. Docs and existing templates have been updated to reflect that change.
* Created an SBOMObservation model that can link arbitrary text to a Vulnerability, a SBOMDocument and optionally a Package. In practice, the following are being recorded as SBOMObservation:
  * Yocto Description from cve_check files (linked to the package)
    * pre-existing Yocto Description stored in the Vulnerability model are migrated as new SBOMObservations
  * CDX Details and CDX Recommendations (not linked to any package)
  * (**new**) Grype "fix" state (either "wont-fix" or "fixed" with version details)
* Refactored "texts" handling in vulnerabilities and assessments routes
  * 1 DB query for all description instead of 1 per vulnerability in assessments route
  * using a list instead of a dict since there can be multiple observations with the same key (e.g. multiple Yocto Description coming from multiple packages / multiple scanned documents)
* In the Vulnerability modal (frontend), the name of the package(s) that provided the Yocto Description now appear

New DB schema:
erd 2
Example of Grype fix state text showing in the frontend:

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

### Additional notes

## Pull Request Checklist


ValentinBoudevin:
I would split the PR in two with the code refactoring (which is super cool) in another PR to not have a super huge codebase to review

ValentinBoudevin:
Two notes: 

1) You can't modify Vulnerability description field

Vulnerability description should stay the "official" description of the vulnerability. For example, the NVD description provided here https://nvd.nist.gov/vuln/detail/CVE-2022-0116 with `Inappropriate implementation in Compositing in Google Chrome prior to 97.0.4692.71 allowed a remote attacker to spoof the contents of the Omnibox (URL bar) via a crafted HTML page.`

You can't concatenate anything at the end of this description, as we are creating a mechanism to refresh data for vulnerabilities. Example, refresh CVE-2022-0116 will fetch details of the CVE and update the description/links/cvss was changed compare to the sbom/last refresh.

2) Yocto Description should be linked to an assessment for a given variant (instead of being linked the the Vulnerability as it is right now)

For example,  CVE-2018-6559 in `example/spdx3/core-image-minimal-qemux86-64.rootfs.json` have the yocto description `"description": "Issue only affects Ubuntu"` should be the assessment status_note. 
Maybe one of your variants will be based on ubuntu distro and this CVE will be application and the yocto description will be wrong. 
This is why it can't be global to a CVE.

Same for CVE-2025-38294 with `only affects 6.14 onwards`, but what if a variant is running kernel 6.14? This is why this info should be assessment specific only.

Note: I don't know about CDX details or Grype Extra info if they should be stored on the assessment or global to the vulnerability. Then SBOMObservation may be needed. But for Yocto Description, I don't think so

youennlj-sfl:
> I would split the PR in two with the code refactoring (which is super cool) in another PR to not have a super huge codebase to review

Done, #324

> Vulnerability description should stay the "official" description of the vulnerability. [...]
> You can't concatenate anything at the end of this description, as we are creating a mechanism to refresh data for vulnerabilities.

OK, in this case I'll put everything that is not the official description in an SBOMObservation.

> Yocto Description should be linked to an assessment for a given variant (instead of being linked the the Vulnerability as it is right now)

Yes, that's exactly what I did in this PR with SBOMObservation. It's a new model that depends on a SBOMDocument -> Scan -> Variant. cf. attached ERD
vulnscout erd 2

I will refactor SBOMObservation to directly link to a Vulnerability instead of having to go through a Finding, eliminating the need for a package (e.g. for CDX details)

src/models/vulnerability.py
```diff
@@ -408,22 +404,11 @@ def merge(self, other: "Vulnerability") -> bool:
...
-            tmp_inc = 2
-            while new_key in self.texts:
-                new_key = title + "_" + str(tmp_inc)
-                tmp_inc += 1
-            self.add_text(other.texts[title], new_key)
+        if other.description:
+            if self.description is None:
+                self.description = other.description
+            elif other.description not in self.description:
+                self.description += "\n\n" + other.description
```
youennlj-sfl, line 414: We have to decide what do we want to do here. Replace the old description by the new one?

ValentinBoudevin:
Can you remove content linked with refactoring? As already present in your other PR

youennlj-sfl:
> Can you remove content linked with refactoring? As already present in your other PR

No, those changes are needed for this PR. If you want to review without them, you can exclude the commits related with the other PR in the commits picker at the top of the review page.

EDIT: since the other PR has been merged, I rebased so it should be good for you now

src/routes/assessments.py
```diff
@@ -109,28 +111,53 @@ def review_assessments():
...
+            })
+
+        # Enrich with observation statuses
+        sbom_observation_query = (
+            select(
+                SBOMObservation.vulnerability_id,
+                SBOMObservation.key,
+                SBOMObservation.description,
+            )
+            .where(SBOMObservation.vulnerability_id.in_(vuln_ids))
```
WilsonZiweiWang: Do we need `.where(SBOMDocument.scan_id.in_(active_scan_ids))` here like `vulnerabilities.py` line 525?

youennlj-sfl: Ah yeah you may be right

youennlj-sfl: Yes nice catch, I'm adding this and writing a test

src/migrations/versions/7cb28bf85fcc_add_sbom_observation.py
```diff
@@ -0,0 +1,78 @@
...
+def upgrade():
+    # ### commands auto generated by Alembic ###
+    op.create_table('sbom_observation',
+    sa.Column('id', sa.Uuid(), nullable=False),
+    sa.Column('key', sa.String(), nullable=False),
+    sa.Column('description', sa.String(), nullable=False),
+    sa.Column('vulnerability_id', sa.String(length=50), nullable=False),
+    sa.Column('package_id', sa.Uuid(), nullable=True),
+    sa.Column('sbom_document_id', sa.Uuid(), nullable=False),
+    sa.ForeignKeyConstraint(['package_id'], ['packages.id'], ),
```
ValentinBoudevin, line 28: You use both vulnerability_id and package_id. 
Wouldn't it be better to have a foreign key on finding directly?

youennlj-sfl: No, as you can see package_id is nullable. Some texts are scoped to an entire vulnerability and not to a specific package having the vulnerability (e.g. CDX details).
I initially made the error though and I changed it later to not use Finding and explicitely use Vuln and optional Package.

ValentinBoudevin (dismissed):
LGTM

reh-sfl:
Migration down re-creates the old yocto_description but it also needs to backfill it with the existing data so it doesn't get lost.

e.g:

```python
    UPDATE vulnerabilities
    SET yocto_description = (
        SELECT description FROM sbom_observation
        WHERE vulnerability_id = vulnerabilities.id
          AND key = 'Yocto Description'
        LIMIT 1
    )
```

reh-sfl:
I tried running this branch with Postgres and it seems the following is SQLite only, could be done using more generic syntax or helper?
```sql
func.aggregate_strings(Package.name, ";"),
```

youennlj-sfl:
@reh-sfl 
> Migration down re-creates the old yocto_description but it also needs to backfill it with the existing data so it doesn't get lost.

I added a downgrade code but did not test it. When are downgrade migrations used anyway?

> I tried running this branch with Postgres and it seems the following is SQLite only, could be done using more generic syntax or helper?

Fixed by refactoring the way it works.

youennlj-sfl:
Just tested both migration and from fresh and it works fine :+1:

ValentinBoudevin (approved):
LGTM

reh-sfl (approved):

PR #323, opened 2026-05-05
Parse supplier info in SPDX 2.3

WilsonZiweiWang:

This PR brings support for parsing the supplier info in SPDX 2.3 and addresses #276.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

You may use the following files to run VulnScout:

[test-supplier-vulns.grype.json](https://github.com/user-attachments/files/27446521/test-supplier-vulns.grype.json)
[test-supplier-distinction.spdx.json](https://github.com/user-attachments/files/27446527/test-supplier-distinction.spdx.json)

Move these files to `/example/spdx2/`, then run: 
```
 ./vulnscout --dev \
   --add-spdx example/spdx2/test-supplier-distinction.spdx.json \
   --add-grype example/spdx2/test-supplier-vulns.grype.json \
   --serve
```

You should see an extra toggle for  suppliers in the columns dropdown for the SBOM table:

image

image

You can see the supplier info in the Vulnerability modal:

image

image

The openvex file will also contain the supplier info in the PURL:

image

### Additional notes

### Related Issue

## Pull Request Checklist

### DIFF
--- a/src/models/package.py
+++ b/src/models/package.py
@@ -24,6 +24,7 @@ class Package(Base):
     cpe = db.Column(db.JSON, nullable=True)
     purl = db.Column(db.JSON, nullable=True)
     licences = db.Column(db.String, nullable=True)
+    supplier = db.Column(db.String, nullable=True, default="")
 
     __table_args__ = (
         db.Index('ix_packages_name_version', 'name', 'version'),
@@ -44,6 +45,7 @@ class Package(Base):
         cpe: Optional[list] = None,
         purl: Optional[list] = None,
         licences: str = "",
+        supplier: str = "",
         **kwargs,
     ):
         version = str(version).strip().split("+git")[0]
@@ -63,6 +65,7 @@ class Package(Base):
         self.cpe = []
         self.purl = []
         self.licences = licences or ""
+        self.supplier = supplier or ""
         for c in cpes:
             self.add_cpe(c)
         for p in purls:
@@ -74,7 +77,9 @@ class Package(Base):
 
     @property
     def string_id(self) -> str:
-        """Return the human-readable ``'name@version'`` identifier."""
+        """Return the human-readable identifier, including supplier when present."""
+        if self.supplier:
+            return f"{self.name}@{self.version}::{self.supplier}"
         return f"{self.name}@{self.version}"
 
     # TODO: Remove in-memory logic in parsers to use DB directly. The following are concerned
@@ -130,12 +135,16 @@ class Package(Base):
         if not isinstance(other, Package):
             return NotImplemented
         try:
-            return self.name == other.name and self._parse_version() == other._parse_version()
+            return (self.name == other.name
+                    and self._parse_version() == other._parse_version()
+                    and (self.supplier or "") == (other.supplier or ""))
         except Exception:
-            return self.name == other.name and self.version == other.version
+            return (self.name == other.name
+                    and self.version == other.version
+                    and (self.supplier or "") == (other.supplier or ""))
 
     def __hash__(self) -> int:
-        return hash((self.name, self.version))
+        return hash((self.name, self.version, self.supplier or ""))
 
     def __str__(self) -> str:
         return self.string_id
@@ -147,17 +156,23 @@ class Package(Base):
         if self.name != other.name:
             return self.name < other.name
         try:
-            return self._parse_version() < other._parse_version()
+            if self._parse_version() != other._parse_version():
+                return self._parse_version() < other._parse_version()
         except Exception:
-            return self.version < other.version
+            if self.version != other.version:
+                return self.version < other.version
+        return (self.supplier or "") < (other.supplier or "")
 
     def __gt__(self, other) -> bool:
         if self.name != other.name:
             return self.name > other.name
         try:
-            return self._parse_version() > other._parse_version()
+            if self._parse_version() != other._parse_version():
+                return self._parse_version() > other._parse_version()
         except Exception:
-            return self.version > other.version
+            if self.version != other.version:
+                return self.version > other.version
+        return (self.supplier or "") > (other.supplier or "")
 
     def __le__(self, other) -> bool:
         return self < other or self == other
@@ -191,6 +206,7 @@ class Package(Base):
             "cpe": list(self.cpe or []),
             "purl": list(self.purl or []),
             "licences": self.licences or "",
+            "supplier": self.supplier or "",
         }
 
     @staticmethod
@@ -201,6 +217,7 @@ class Package(Base):
             cpe=data.get("cpe", []),
             purl=data.get("purl", []),
             licences=data.get("licences", ""),
+            supplier=data.get("supplier", ""),
         )
 
     # ------------------------------------------------------------------
--- a/src/models/package.py
+++ b/src/models/package.py
@@ -2,6 +2,8 @@
 # SPDX-License-Identifier: GPL-3.0-only
 
 import uuid
+import re
+import hashlib
 import semver
 from typing import Optional
 from ..extensions import db, Base
@@ -110,7 +112,15 @@ class Package(Base):
 
     def generate_generic_purl(self) -> str:
         """Build a generic purl string for the package, add it to the purl list and return it."""
-        item = f"pkg:generic/{self.name}@{self.version}"
+        if self.supplier:
+            name_part = re.sub(r'^[^:]+:\s*', '', self.supplier)   # strip "Organization: "
+            name_part = re.sub(r'\s*\(.*\)$', '', name_part).strip()  # strip email
+            slug = re.sub(r'[^\w]+', '-', name_part).strip('-').lower()
+            if not slug:
+                slug = "supplier-" + hashlib.sha1(self.supplier.encode()).hexdigest()[:8]
+            item = f"pkg:generic/{slug}/{self.name}@{self.version}"
+        else:
+            item = f"pkg:generic/{self.name}@{self.version}"
         self.add_purl(item)
         return item
 
--- a/src/models/package.py
+++ b/src/models/package.py
@@ -24,7 +24,7 @@ class Package(Base):
     cpe = db.Column(db.JSON, nullable=True)
     purl = db.Column(db.JSON, nullable=True)
     licences = db.Column(db.String, nullable=True)
-    supplier = db.Column(db.String, nullable=True, default="")
+    supplier = db.Column(db.String, nullable=False, default="")
 
     __table_args__ = (
         db.Index('ix_packages_name_version_supplier', 'name', 'version', 'supplier'),
@@ -65,7 +65,7 @@ class Package(Base):
         self.cpe = []
         self.purl = []
         self.licences = licences or ""
-        self.supplier = supplier or ""
+        self.supplier = supplier
         for c in cpes:
             self.add_cpe(c)
         for p in purls:
@@ -137,14 +137,14 @@ class Package(Base):
         try:
             return (self.name == other.name
                     and self._parse_version() == other._parse_version()
-                    and (self.supplier or "") == (other.supplier or ""))
+                    and (self.supplier) == (other.supplier))
         except Exception:
             return (self.name == other.name
                     and self.version == other.version
-                    and (self.supplier or "") == (other.supplier or ""))
+                    and (self.supplier) == (other.supplier))
 
     def __hash__(self) -> int:
-        return hash((self.name, self.version, self.supplier or ""))
+        return hash((self.name, self.version, self.supplier))
 
     def __str__(self) -> str:
         return self.string_id
@@ -161,7 +161,7 @@ class Package(Base):
         except Exception:
             if self.version != other.version:
                 return self.version < other.version
-        return (self.supplier or "") < (other.supplier or "")
+        return self.supplier < other.supplier
 
     def __gt__(self, other) -> bool:
         if self.name != other.name:
@@ -172,7 +172,7 @@ class Package(Base):
         except Exception:
             if self.version != other.version:
                 return self.version > other.version
-        return (self.supplier or "") > (other.supplier or "")
+        return self.supplier > other.supplier
 
     def __le__(self, other) -> bool:
         return self < other or self == other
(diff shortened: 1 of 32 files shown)


rawaha-e (changes requested):
1. In `merger_ci.py`, you need to apply the `split("::", 1)` before rsplit("@", 1) like you did in `assessments.py`
2. I think it would be better to not have a separate supplier column in frontend but instead add it as part of the Name column when a supplier exists, e.g: `mozilla / firefox`. Make sure to keep sorting on `Name` and not the whole `Supplier / Name`

ValentinBoudevin:
I also belive having "Organization: " is too long. 

Can be truncate to `openssl@3.0.0::Acme_Corp`. (I also replaced spaces with _ to have a consistent key)

image

src/models/package.py
```diff
@@ -26,9 +26,10 @@ class Package(Base):
     cpe = db.Column(db.JSON, nullable=True)
     purl = db.Column(db.JSON, nullable=True)
     licences = db.Column(db.String, nullable=True)
+    supplier = db.Column(db.String, nullable=True, default="")
```
youennlj-sfl: Why not keep the default None? This is confusing.
If you want the default to be the empty string then please put `nullable=False` (although personally I would advocate for using None when no supplier is given)

youennlj-sfl: Also for newer additions to DB models, please correctly type hint the columns so type-checkers can catch bugs
```suggestion
    supplier: Mapped[str | None] = db.Column(db.String, nullable=True, default="")
```

src/models/package.py
```diff
@@ -65,6 +67,7 @@ def __init__(
         self.cpe = []
         self.purl = []
         self.licences = licences or ""
+        self.supplier = supplier or ""
```
youennlj-sfl: No need for `or ""` since supplier is typed as `str` and not `str | None`

src/models/package.py
```diff
@@ -46,6 +47,7 @@ def __init__(
         cpe: Optional[list] = None,
         purl: Optional[list] = None,
         licences: str = "",
+        supplier: str = "",
```
youennlj-sfl, line 50: If you want to keep the `supplier` column nullable then you should make this `str | None` and give it `None` as the default value.

src/models/package.py
```diff
@@ -132,12 +137,16 @@ def __eq__(self, other) -> bool:
...
+                    and (self.supplier or "") == (other.supplier or ""))
         except Exception:
-            return self.name == other.name and self.version == other.version
+            return (self.name == other.name
+                    and self.version == other.version
+                    and (self.supplier or "") == (other.supplier or ""))
 
     def __hash__(self) -> int:
-        return hash((self.name, self.version))
+        return hash((self.name, self.version, self.supplier or ""))
```
youennlj-sfl: I don't understand why you are making all of those `self.supplier or ""`
(comment valid for the rest of the file)

src/routes/assessments.py
```diff
@@ -199,10 +199,12 @@ def _get_vuln(vuln_id: str):
...
                     for pkg_str in assess.packages:
-                        if "@" in pkg_str:
-                            name, version = pkg_str.rsplit("@", 1)
+                        _parts = pkg_str.split("::", 1)
+                        _base = _parts[0]
+                        if "@" in _base:
+                            name, version = _base.rsplit("@", 1)
                         else:
-                            name, version = pkg_str, ""
+                            name, version = _base, ""
```
youennlj-sfl: Why don't you use the the supplier part here?

src/routes/assessments.py
```diff
@@ -604,10 +612,13 @@ def add_assessments_batch():
...
-                                             if "@" in pkg_string_id
-                                             else (pkg_string_id, ""))
-                            db_pkg = Package.find_or_create(name, version)
+                            _parts = pkg_string_id.split("::", 1)
+                            _base = _parts[0]
+                            _supplier = _parts[1] if len(_parts) > 1 else ""
+                            name, version = (_base.rsplit("@", 1)
+                                             if "@" in _base
+                                             else (_base, ""))
+                            db_pkg = Package.find_or_create(name, version, supplier=_supplier)
```
youennlj-sfl: This code is duplicated thrice. It might be interesting to make a function that can be reused instead.

src/models/package.py
youennlj-sfl: Shouldn't this be updated to contain the supplier if present?

src/views/openvex.py
youennlj-sfl, line 1: Those changes are (I believe) unnecessary if the `generate_generic_purl` method of `src/models/package.py` is updated to contain the supplier.

tests/unit_tests/test_openvex_output.py
```diff
@@ -0,0 +1,85 @@
...
+    assert product["@id"] == "pkg:generic/foo@1.0"
+
+
+def test_openvex_supplier_uses_qualified_id():
+    pkg = _make_pkg("foo", "1.0", supplier="Organization: Acme Corp (x@a.com)")
+    assess = _make_assessment("CVE-2024-1", pkg)
+    result = _run_to_dict([pkg], [assess])
+    product = result["statements"][0]["products"][0]
+    assert product["@id"] == "pkg:generic/acme-corp/foo@1.0"
+    # identifiers.purl must remain generic
```
youennlj-sfl: Why?

tests/unit_tests/test_package_db.py
```diff
@@ -0,0 +1,108 @@
...
+    try:
+        application = create_app()
+        application.config.update({"TESTING": True})
+        with application.app_context():
+            _db.create_all()
+            yield application
+            _db.drop_all()
+    finally:
+        os.environ.pop("FLASK_SQLALCHEMY_DATABASE_URI", None)
+        os.environ.pop("SCAN_FILE", None)
```
youennlj-sfl: Use the monkeypatch cap of pytest instead of manually changing os.environ

```suggestion
def app(monkeypatch):
    monkeypatch.setenv("FLASK_SQLALCHEMY_DATABASE_URI", "sqlite:///:memory:")
    monkeypatch.setenv("SCAN_FILE", "/dev/null")
    application = create_app()
    application.config.update({"TESTING": True})
    with application.app_context():
        _db.create_all()
        yield application
        _db.drop_all()
```

tests/unit_tests/test_package_db.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def test_find_or_create_different_supplier_creates_new_row(app):
+    with app.app_context():
```
youennlj-sfl: I think you can remove this context manager since you are already in the app_context thanks to the app fixture.
Same for all the other tests in this file.

youennlj-sfl (dismissed):

src/models/package.py
WilsonZiweiWang: Good point, should have made the nullable false or removed the default like the `licenses`. For the `Mapped` typing, I'd like to leave it for another discussion or another refactor PR that enforces something like this for all the columns.

WilsonZiweiWang: seems like making nullable false is a better choice overall. Addressed.

src/models/package.py
WilsonZiweiWang: Addressed

src/models/package.py
WilsonZiweiWang: Addressed.

src/models/package.py
WilsonZiweiWang: Addressed.

src/routes/assessments.py
WilsonZiweiWang: The expected package string is in this format: `{package.name}@{package.version}::{self.supplier}`. After the split using `::`, the `_base` only contains the part `{package.name}@{package.version}`. The rest is just to respect the original logic, where the version doesn't exist in the `pkg_str` (`@` is missing`).

src/routes/assessments.py
WilsonZiweiWang: Good point, addressed.

src/models/package.py
WilsonZiweiWang: Yes, it makes sense to have the supplier here. I moved the changes (the strip code is still needed in my opinion, as we don't want everything in the raw string) from openvex.py to here. Addressed.

src/views/openvex.py
WilsonZiweiWang: Addressed

WilsonZiweiWang:
Rebased `staging`.

tests/unit_tests/test_openvex_output.py
WilsonZiweiWang: This test has been updated due to the changes I made to `generate_generic_purl`.

tests/unit_tests/test_package_db.py
WilsonZiweiWang: Updated.

tests/unit_tests/test_package_db.py
WilsonZiweiWang: addressed and removed it for other tests as well.

WilsonZiweiWang:
> I also belive having "Organization: " is too long.
> 
> Can be truncate to `openssl@3.0.0::Acme_Corp`. (I also replaced spaces with _ to have a consistent key)
> 

@ValentinBoudevin  Do we still want to display the supplier type somewhere?

WilsonZiweiWang:
> 2. I think it would be better to not have a separate supplier column in frontend but instead add it as part of the Name column when a supplier exists, e.g: `mozilla / firefox`. Make sure to keep sorting on `Name` and not the whole `Supplier / Name`

@rawaha-e Addressed. The full supplier string is shown on hovering the package name.

image

WilsonZiweiWang:
As agreed @ValentinBoudevin @rawaha-e, the supplier column is put back and is shown by default only when at least one package has supplier info.
image

WilsonZiweiWang:
As discussed @ValentinBoudevin, the following changes are made:
1. Packages with no supplier info will have `unknown supplier` appended when viewing the vulnerability page and the vuln modal.
2. Only the supplier names are displayed; the prefix (e.g., `Organization: `) and the postfix (e.g., email) are truncated.
image
image
image
3. Scan results now show supplier info.
image
image

ValentinBoudevin:
Works perfectly for the SBOM, Vulnerabilities, and Scan tab. 

Few modifications are still needed: 

1- Review tab: Add a supplier column

image

2- SBOM tab: Would it be possible to add a new filter based on Supplier info?

3- This PR was made to fix the git issue raised concerning the supplier. 

I gave a try the git issue file and ran into an error 

```
  File "/usr/lib/python3.12/site-packages/click/core.py", line 1406, in main
Hint: set IGNORE_PARSING_ERRORS=true to ignore this error
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/click/core.py", line 1873, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/click/core.py", line 1269, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/click/core.py", line 824, in invoke
    return callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/click/decorators.py", line 34, in new_func
    return f(get_current_context(), *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/flask/cli.py", line 386, in decorator
    return ctx.invoke(f, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/click/core.py", line 824, in invoke
    return callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/scan/src/bin/cmd_process.py", line 237, in process_command
    _run_main()
  File "/scan/src/bin/cmd_process.py", line 328, in _run_main
    read_inputs(controllers, scan_id=scan_id)
  File "/scan/src/bin/cmd_process.py", line 160, in read_inputs
    raise e
  File "/scan/src/bin/cmd_process.py", line 136, in read_inputs
    spdx.load_from_file(doc.path)
  File "/scan/src/views/spdx.py", line 54, in load_from_file
    try_reading = parse_file(spdx_file)
                  ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/spdx_tools/spdx/parser/parse_anything.py", line 34, in parse_file
    return json_parser.parse_from_file(file_name, encoding)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/spdx_tools/spdx/parser/json/json_parser.py", line 16, in parse_from_file
    return JsonLikeDictParser().parse(input_doc_as_dict)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/spdx_tools/spdx/parser/jsonlikedict/json_like_dict_parser.py", line 88, in parse
    raise_parsing_error_if_logger_has_messages(self.logger)
  File "/usr/lib/python3.12/site-packages/spdx_tools/spdx/parser/parsing_functions.py", line 26, in raise_parsing_error_if_logger_has_messages
    raise SPDXParsingError(logger.get_messages())
spdx_tools.spdx.parser.error.SPDXParsingError: ["Error while parsing Document: ['Error while parsing version 3.28.0: 3.28.0 is not a valid version string']"]
```

Due to "licenseListVersion": "3.28.0" considered as invalid. Can we add the fix for this issue in the PR please?

WilsonZiweiWang:
> Works perfectly for the SBOM, Vulnerabilities, and Scan tab.
> 
> Few modifications are still needed:
> 
> 1- Review tab: Add a supplier column
> 
> 2- SBOM tab: Would it be possible to add a new filter based on Supplier info?
> 
> 3- This PR was made to fix the git issue raised concerning the supplier.
> 
> Due to "licenseListVersion": "3.28.0" considered as invalid. Can we add the fix for this issue in the PR please?

@ValentinBoudevin All addressed.

ValentinBoudevin (approved):
LGTM

rawaha-e (approved):
LGTM

PR #324, opened 2026-05-11
Backend typing and small refactoring

youennlj-sfl:

* _Lots_ of typing fixes
  * it allows to catch bugs introduced by refactorings much quicker and easier. New code should ALWAYS be properly typed (especially in DB models)
* Refactored some parts of the code (current scan management)
* Deletion of dead code
* Added doc for Alembic migration creation

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

## Pull Request Checklist

### DIFF
--- a/src/bin/cmd_export.py
+++ b/src/bin/cmd_export.py
@@ -44,7 +44,7 @@ def export_command(export_format: str, output_dir: str) -> None:
     try:
         if fmt == "spdx2":
             spdx = SPDX(ctrls)
-            content = spdx.output_as_json(author)
+            content = spdx.output_as_json(author=author)
             out_path = os.path.join(output_dir, "sbom_spdx_v2_3.spdx.json")
             with open(out_path, "w") as fh:
                 fh.write(content)
--- a/src/controllers/packages.py
+++ b/src/controllers/packages.py
@@ -1,11 +1,9 @@
 # Copyright (C) 2026 Savoir-faire Linux, Inc.
 # SPDX-License-Identifier: GPL-3.0-only
 
-from ..models.package import Package
+from ..models import Package, Finding, SBOMDocument, SBOMPackage
 from ..helpers.verbose import verbose
-from ..models.finding import Finding
 from ..extensions import db
-from ..models.sbom_package import SBOMPackage
 
 
 class PackagesController:
@@ -20,7 +18,7 @@ class PackagesController:
 
     def __init__(self):
         self._cache: dict[str, Package] = {}
-        self._current_sbom_document_id = None
+        self._current_sbom_document: SBOMDocument | None = None
         # Fast PK lookup: string_id → DB UUID.  Avoids SELECT in
         # get_by_string_id for packages we already persisted.
         self._db_id_cache: dict = {}
@@ -54,9 +52,15 @@ class PackagesController:
         except Exception as e:
             verbose(f"[PackagesController._preload_cache findings] {e}")
 
-    def set_sbom_document(self, doc_id) -> None:
+    @property
+    def current_sbom_document(self) -> SBOMDocument | None:
+        return self._current_sbom_document
+
+    @current_sbom_document.setter
+    def current_sbom_document(self, doc: SBOMDocument | None) -> None:
         """Set (or clear with ``None``) the SBOM document that subsequent :meth:`add` calls belong to."""
-        self._current_sbom_document_id = doc_id
+        self._current_sbom_document = doc
+        verbose(f"[PackagesController] Now handling SBOM document {repr(doc)}")
 
     # ------------------------------------------------------------------
     # Fast accessors for other controllers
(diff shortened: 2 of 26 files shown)


src/controllers/vulnerabilities.py
```diff
@@ -230,8 +230,7 @@ def add(self, vulnerability: Vulnerability) -> Optional[Vulnerability]:
         and have not gained new packages in this call — avoiding redundant
         get_by_id SELECTs and update_record commits on every re-encounter.
         """
-        if vulnerability is None:
-            return
+        assert vulnerability is not None
```
rawaha-e: I would avoid `assert` on input fields because some Python runtimes will not react properly to it (e.g throwing an `AssertionError`). Could use `if ... is None` is enough.

src/controllers/vulnerabilities.py
```diff
@@ -73,6 +74,7 @@ def _should_refetch(fetched_at: Optional[datetime.datetime], delay: Optional[dat
         return True
     if delay is _NEVER:
         return False
+    assert delay
```
WilsonZiweiWang: If the timedelta is zero, this will raise an assertionError. Is this intended?

src/models/__init__.py
```diff
@@ -14,3 +14,5 @@
 from .assessment import Assessment  # noqa: F401
 from .time_estimate import TimeEstimate  # noqa: F401
 from .metrics import Metrics  # noqa: F401
+from .cvss import CVSS
```
rawaha-e, line 17: Do we need # noqa: F401 for the new imports?

src/bin/cmd_export.py
```diff
@@ -44,7 +44,7 @@ def export_command(export_format: str, output_dir: str) -> None:
     try:
         if fmt == "spdx2":
             spdx = SPDX(ctrls)
-            content = spdx.output_as_json(author)
+            content = spdx.output_as_json(author=author)
```
WilsonZiweiWang, line 47: This introduces a different behavior given the signature of the function `output_as_json()`. Since we are passing `author` here, the `(author=author)` seems to be the correct approach. Does it mean the `content` was wrong all along? If that's the case, I think we need a test for this.

src/models/__init__.py
youennlj-sfl: No, see the change in tox.ini

src/controllers/vulnerabilities.py
youennlj-sfl: Well, that is exactly why we should assert not None: we _want_ to get AssertionError since vulnerability should _not_ be None.
If you want I can even remove this assertion since an error will be raised anyway further below (at `if vulnerability.id in self.vulnerabilities:`)

WilsonZiweiWang: Is it necessary to assert here? Because it might stop all the `add()` for the rest of the vulns when one raises the assertion error, and stop the entire backend.

src/models/__init__.py
```diff
@@ -14,3 +14,4 @@
 from .assessment import Assessment  # noqa: F401
 from .time_estimate import TimeEstimate  # noqa: F401
 from .metrics import Metrics  # noqa: F401
```
WilsonZiweiWang: Since you added a universal rule in `tox.ini`, you may remove the comments after each line.

WilsonZiweiWang (changes requested):

src/controllers/vulnerabilities.py
rawaha-e: assert delay is not None instead

src/models/__init__.py
rawaha-e: You can remove the rules from this file then.

rawaha-e (changes requested):

src/controllers/vulnerabilities.py
youennlj-sfl: Tests still pass and the method signature says that vulnerability is not None. Typechecking does not raise a warning that a potential none value is passed to this function. Thus I assume it's fine

src/controllers/packages.py
```diff
@@ -89,6 +89,7 @@ def add(self, package: Package):
         """Persist a Package to the DB and keep it in the session cache."""
         if package is None:
             return
+        assert package.name and package.version
```
WilsonZiweiWang: This is confusing. I noticed some parsers accept an empty version string. For example, in `src/views/spdx.py`, line 104:
```python
pkg = Package(package.name, package.version or "", [], [], "", supplier=supplier)
```
in `src/views/fast_spdx.py`, line 42:
```python
package = Package(name, version or "", [], [], licences or "")
```

Which rule should we enforce here?

src/bin/cmd_export.py
youennlj-sfl: Content was correct. Author was being passed to the "validate" parameter and was coerced to a boolean here: https://github.com/spdx/tools-python/blob/cef432adee09dbff2eedf7e33e4aa8be7c11ccfd/src/spdx_tools/spdx/writer/write_utils.py#L14

src/controllers/packages.py
youennlj-sfl: In this case I'll remove this change and make the name and version nullable in Package.find_or_create which is why I initially made this change.

src/bin/cmd_export.py
youennlj-sfl: I wrote a test

src/controllers/vulnerabilities.py
youennlj-sfl: I removed the assertion since an error will be raised anyway further down the method body

src/controllers/vulnerabilities.py
youennlj-sfl: No. I changed for `assert delay is not None`

rawaha-e (approved):
LGTM

ValentinBoudevin (approved):
Lgtm

WilsonZiweiWang (approved):

PR #327, opened 2026-05-12
Handle cpe23Type in BOM and normalize purl epoch qualifier

### LINKED ISSUE #276: SPDX to OpenVEX generation collapses distinct BOM components into ambiguous product identities

WilsonZiweiWang:

This PR addresses https://github.com/savoirfairelinux/vulnscout/issues/276#issuecomment-4427368398.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Download the files mentioned in the issue comment above and run Vulnscout with the following command after moving the relevant files to the root path of Vulnscout. Also, you need to rename `syft_spdx.json` to `syft.spdx.json` as the `entrypoint` doesn't accept `*_spdx.json` (This should be a careless mistake by the issue author):
```
./vulnscout --dev \
   --add-spdx syft.spdx.json \
   --add-grype syft_spdx.grype.json \
   --serve
```
Then, navigate to the export page and download the OpenVEX file. The * vendor should be gone. For example, the string mentioned in the issue comment should not be found in the new OpenVEX file:
```
cpe:2.3:a:*:libtasn1-6:4.13-2:*:*:*:*:*:*:*
cpe:2.3:a:*:util-linux:2.31.1-0.4ubuntu3.7:*:*:*:*:*:*:*
```

A string like the following in BOM:
```
pkg:deb/ubuntu/procps@2%3A3.3.12-3ubuntu1.2?arch=arm64&distro=ubuntu-18.04
```
It will like this in VEX:
```
pkg:deb/ubuntu/libprocps6@2:3.3.12-3ubuntu1.2?arch=arm64&distro=ubuntu-18.04&upstream=procps
```

### Additional notes

### Related Issue

## Pull Request Checklist

### DIFF
--- a/src/models/package.py
+++ b/src/models/package.py
@@ -22,15 +22,19 @@ from packageurl import PackageURL
 def _normalize_purl(purl: str) -> str:
     """Normalize PURL to a canonical form.
 
-    Handles two normalizations for ``deb`` type PURLs:
+    Handles two normalizations for ``deb`` and ``rpm`` type PURLs:
     - URL-decodes percent-encoded characters in the version (e.g. ``2%3A`` → ``2:``)
     - Moves ``epoch=N`` qualifier into the version field as ``N:version``
+
+    ``epoch`` is the only qualifier that requires this special treatment per the PURL spec.
+    All other standard normalizations (qualifier ordering, component encoding) are handled
+    transparently by the ``packageurl-python`` library during parse and re-serialisation.
 
     Falls back to the original string if parsing fails.
     """
     try:
         p = PackageURL.from_string(purl)
-        if p.type == "deb" and p.qualifiers and "epoch" in p.qualifiers and p.version is not None:
+        if p.type in ("deb", "rpm") and p.qualifiers and "epoch" in p.qualifiers and p.version is not None:
             epoch = p.qualifiers["epoch"]
             new_qualifiers = {k: v for k, v in p.qualifiers.items() if k != "epoch"}
             new_version = f"{epoch}:{p.version}"
--- a/src/models/package.py
+++ b/src/models/package.py
@@ -16,6 +16,31 @@ from ..extensions import db, Base
 
 if typing.TYPE_CHECKING:
     from ..models import SBOMPackage, Finding
+from packageurl import PackageURL
+
+
+def _normalize_purl(purl: str) -> str:
+    """Normalize PURL to a canonical form.
+
+    Handles two normalizations for ``deb`` type PURLs:
+    - URL-decodes percent-encoded characters in the version (e.g. ``2%3A`` → ``2:``)
+    - Moves ``epoch=N`` qualifier into the version field as ``N:version``
+
+    Falls back to the original string if parsing fails.
+    """
+    try:
+        p = PackageURL.from_string(purl)
+        if p.type == "deb" and p.qualifiers and "epoch" in p.qualifiers and p.version is not None:
+            epoch = p.qualifiers["epoch"]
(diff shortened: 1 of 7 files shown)


src/models/package.py
```diff
@@ -16,6 +16,31 @@
...
+
+    Handles two normalizations for ``deb`` type PURLs:
+    - URL-decodes percent-encoded characters in the version (e.g. ``2%3A`` → ``2:``)
+    - Moves ``epoch=N`` qualifier into the version field as ``N:version``
+
+    Falls back to the original string if parsing fails.
+    """
+    try:
+        p = PackageURL.from_string(purl)
+        if p.type == "deb" and p.qualifiers and "epoch" in p.qualifiers and p.version is not None:
```
ValentinBoudevin: Seems a bit too specific. 
Is it only true for debian (deb) packages ?
The epoch can't apply to any other system?

ValentinBoudevin (changes requested):

src/models/package.py
ValentinBoudevin: Another question. Does the purl specification contains other variables like epoch or is it the only one? 
We should cover all possibilities

WilsonZiweiWang: I added a commit to handle rpm as well. Currently, it seems that everything else other than epoch can be handled by the `packageurl-python` lib.

ValentinBoudevin (approved):
Lgtm

rawaha-e (approved):

PR #329, opened 2026-05-13
Refactor Flask command scripts

rawaha-e:

* Refactoring cmd files for Flask CLI commands: move all shared code to _common

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

## Pull Request Checklist


src/bin/_common.py
```diff
@@ -0,0 +1,117 @@
...
+        _var = DBVariant.get_by_name_and_project(variant_name, project_obj.id)
+        if not _var:
+            click.echo(f"Error: variant not found: {variant_name}")
+            raise SystemExit(1)
+        variant_obj = _var
+
+    return project_obj, variant_obj
+
+
+def build_controllers(*, preload_cache: bool = False, include_all: bool = False) -> dict:
```
youennlj-sfl: Wouldn't it be the time to refactor the controllers dict? Having an untyped dict is quite bad since typecheckers cannot do their jobs properly.
Maybe we should have a dedicated data structure like a dataclass? Or use a real dependency injection framework?

src/bin/_common.py
```diff
@@ -0,0 +1,117 @@
...
+    if include_all:
+        from ..controllers.projects import ProjectController
+        from ..controllers.variants import VariantController
+        from ..controllers.scans import ScanController
+        from ..controllers.sbom_documents import SBOMDocumentController
+        controllers.update({
+            "projects": ProjectController(),
+            "variants": VariantController(),
+            "scans": ScanController(),
+            "sbom_documents": SBOMDocumentController(),
```
youennlj-sfl: Those controllers don't hold any data, they only contain static methods. I think it's not necessary to create and pass instances. This would require changing code in the commands though.

src/bin/cmd_export.py
```diff
@@ -87,20 +78,11 @@ def report_command(template_name: str, output_dir: str, output_format: str | Non
     Also honours the GENERATE_DOCUMENTS env var (comma-separated list) when
     invoked; TEMPLATE_NAME is always generated regardless.
     """
-    pkgCtrl = PackagesController()
-    vulnCtrl = VulnerabilitiesController(pkgCtrl)
-    assessCtrl = AssessmentsController(pkgCtrl, vulnCtrl)
+    controllers = build_controllers(include_all=True)
+    vulnCtrl = controllers["vulnerabilities"]
+    pkgCtrl = controllers["packages"]
     vulnCtrl = VulnerabilitiesController.from_dict(pkgCtrl, vulnCtrl.to_dict())
```
youennlj-sfl, line 90: I don't understand why we are doing this?

src/bin/cmd_export.py
```diff
@@ -87,20 +78,11 @@ def report_command(template_name: str, output_dir: str, output_format: str | Non
     Also honours the GENERATE_DOCUMENTS env var (comma-separated list) when
     invoked; TEMPLATE_NAME is always generated regardless.
     """
-    pkgCtrl = PackagesController()
-    vulnCtrl = VulnerabilitiesController(pkgCtrl)
-    assessCtrl = AssessmentsController(pkgCtrl, vulnCtrl)
+    controllers = build_controllers(include_all=True)
+    vulnCtrl = controllers["vulnerabilities"]
+    pkgCtrl = controllers["packages"]
```
youennlj-sfl: Now it's untyped... Either do the dataclass refactoring I mentioned above or type hint all the variables.

src/bin/cmd_process.py
```diff
@@ -108,7 +102,7 @@ def read_inputs(controllers, scan_id=None):
     if use_fastspdx:
         verbose("merger_ci: Using FastSPDX parser")
 
-    pkgCtrl: PackagesController = controllers["packages"]
+    pkgCtrl = controllers["packages"]
```
youennlj-sfl: Don't remove typing

src/bin/cmd_process.py
```diff
@@ -303,18 +296,12 @@ def populate_observations(scan, vulnCtrl, log_prefix: str = "merger_ci") -> None
 
 def _run_main() -> dict:
     """Core processing logic (usable both from the CLI command and directly)."""
-    pkgCtrl = PackagesController()
-    # pkgCtrl._preload_cache()  # bulk-load pkg UUIDs + findings into cache; eliminates per-vuln SELECT queries
-    vulnCtrl = VulnerabilitiesController(pkgCtrl)
-    assessCtrl = AssessmentsController(pkgCtrl, vulnCtrl)
+    controllers = build_controllers()
+    vulnCtrl = controllers["vulnerabilities"]
+    assessCtrl = controllers["assessments"]
```
youennlj-sfl: Typing

src/bin/_common.py
```diff
@@ -0,0 +1,117 @@
...
+
+DEFAULT_VARIANT_NAME = "default"
+
+
+def get_default_author() -> str:
+    """Return the author name, reading AUTHOR_NAME at call time."""
+    return os.getenv("AUTHOR_NAME", "Savoir-faire Linux")
+
+
+def resolve_project(project: str):
```
youennlj-sfl: Type hint return value

src/bin/_common.py
```diff
@@ -0,0 +1,117 @@
...
+    from ..controllers.projects import ProjectController
+
+    project_obj = ProjectController.get_by_name(project)
+    if not project_obj:
+        click.echo(f"Error: project not found: {project}")
+        raise SystemExit(1)
+    return project_obj
+
+
+def resolve_project_variant(project: str, variant: str | None, *, create: bool = False):
```
youennlj-sfl: Type hint return values

youennlj-sfl (changes requested):

src/bin/_common.py
```diff
@@ -0,0 +1,124 @@
...
+        from ..controllers.projects import ProjectController
+        from ..controllers.variants import VariantController
+        from ..controllers.scans import ScanController
+        from ..controllers.sbom_documents import SBOMDocumentController
+        controllers.update({
+            "projects": ProjectController,
+            "variants": VariantController,
+            "scans": ScanController,
+            "sbom_documents": SBOMDocumentController,
+        })
```
youennlj-sfl: Simply passing the classes is practically useless. The commands that needs the controller classes should import them themselves. No need to put them in the dict.

src/bin/_common.py
youennlj-sfl: Yeah I really don't like the way it's handled. For example in cmd_export, you use `build_controllers(include_all=True)` but it's not immediately obvious why are you actually including all. You have to dive deep in the code to understand that it is because the controllers dict is passed to the `Templates` constructor, and insides it needs all the controllers.
This dict architecture makes it so easy to forget one controller and the bugs will be hard to catch. No static analysis such as type checking can catch them. We should find a better approach.

tests/end_to_end_tests/test_merger_ci.py
```diff
@@ -940,6 +940,95 @@ def test_import_custom_assessments_targz_not_openvex_inside(
...
+    with app.app_context():
+        runner = app.test_cli_runner()
+        result = runner.invoke(args=[
+            "import-custom-assessments",
+            "--project", _PROJECT_NAME,
+            str(d),
+        ])
+    assert result.exit_code == 0, result.output
+    assert "Imported" in result.output
+
```
youennlj-sfl: All of those tests are duplicates of existing ones, or at least have the same name.

youennlj-sfl (changes requested):

youennlj-sfl (dismissed):
LGTM. Controllers dict shall be refactored in a later PR.

src/bin/cmd_assessments.py
```diff
@@ -235,191 +121,32 @@ def import_custom_assessments_command(file_path: str, project: str, variant: str
...
-                except Exception as e:
-                    errors.append({
-                        "vuln_id": vuln_name,
-                        "package": pkg_string_id,
-                        "error": str(e),
-                    })
-        return created, errors, skipped
+        _, variant_obj = resolve_project_variant(project, variant, create=False)
 
     variant_by_name = build_variant_by_name_map()
```
WilsonZiweiWang: What if multiple projects have a variant with the same name? Will it cause issues?

rawaha-e: You are correct, I will fix this.

WilsonZiweiWang (approved):

ValentinBoudevin (approved):

PR #330, opened 2026-05-13
Refactor controllers

rawaha-e:

* Refactor controllers

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

## Pull Request Checklist

### DIFF
--- a/src/controllers/_base.py
+++ b/src/controllers/_base.py
@@ -1,6 +1,7 @@
 # Copyright (C) 2026 Savoir-faire Linux, Inc.
 # SPDX-License-Identifier: GPL-3.0-only
 
+import typing
 import uuid
 
 
@@ -11,11 +12,15 @@ def ensure_uuid(value: uuid.UUID | str) -> uuid.UUID:
     return value
 
 
-def resolve_entity(entity, getter, label: str = "Record"):
+def resolve_entity[T](
+    entity: T | uuid.UUID | str,
+    getter: typing.Callable[[uuid.UUID], T | None],
+    label: str = "Record",
+) -> T:
     """Resolve *entity* to a model instance.
 
     *entity* may already be a model instance, a UUID, or a UUID string.
-    *getter* is a callable that accepts a UUID-or-string and returns the
+    *getter* is a callable that accepts a UUID and returns the
     instance or ``None``.
 
     :raises ValueError: when the entity cannot be found.
@@ -23,6 +28,8 @@ def resolve_entity(entity, getter, label: str = "Record"):
     # If it's not a basic type (UUID / str), assume it's already a model instance.
     if not isinstance(entity, (uuid.UUID, str)):
         return entity
+    if isinstance(entity, str):
+        entity = uuid.UUID(entity)
     found = getter(entity)
     if found is None:
         raise ValueError(f"{label} not found.")
--- a/src/controllers/_base.py
+++ b/src/controllers/_base.py
@@ -4,6 +4,8 @@
 import typing
 import uuid
 
+T = typing.TypeVar("T")
+
 
 def ensure_uuid(value: uuid.UUID | str) -> uuid.UUID:
     """Normalise *value* to a :class:`uuid.UUID`."""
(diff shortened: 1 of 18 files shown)


src/controllers/progress_tracker.py
```diff
@@ -0,0 +1,92 @@
...
+    * ``_completed_message`` – message written by :meth:`complete`.
+
+    Each concrete subclass maintains its **own** singleton instance.
+    """
+
+    _default_phase: str = "enrichment"
+    _completed_message: str = "Enrichment completed successfully"
+
+    _instance = None
+    _lock = Lock()
```
WilsonZiweiWang: Now, EPSS tracker and NVD tracer do not declare their own lock, `self._lock` in other functions will resolve to the parent class's lock. This means EPSS and NVD can't run concurrently now. Is that intended?

rawaha-e: Addressed in 4a0a9080171f031167d4020d26a245b3b6b43e87

WilsonZiweiWang (dismissed):
LGTM

src/controllers/_base.py
```diff
@@ -0,0 +1,58 @@
...
+    if isinstance(value, str):
+        return uuid.UUID(value)
+    return value
+
+
+def resolve_entity(entity, getter, label: str = "Record"):
+    """Resolve *entity* to a model instance.
+
+    *entity* may already be a model instance, a UUID, or a UUID string.
+    *getter* is a callable that accepts a UUID-or-string and returns the
```
youennlj-sfl: Why accept a UUID string here? It's better to handle all strings at some point so the logic is not duplicated everywhere.

src/controllers/_base.py
```diff
@@ -0,0 +1,58 @@
...
+
+    :raises ValueError: when the entity cannot be found.
+    """
+    # If it's not a basic type (UUID / str), assume it's already a model instance.
+    if not isinstance(entity, (uuid.UUID, str)):
+        return entity
+    found = getter(entity)
+    if found is None:
+        raise ValueError(f"{label} not found.")
+    return found
```
youennlj-sfl: This whole method is completely untyped. Please type hint it so typecheckers can catch bugs. You can use generics for this:

```suggestion
def resolve_entity[T](
    entity: T | uuid.UUID | str,
    getter: typing.Callable[[uuid.UUID], T | None],
    label: str = "Record",
) -> T:
    """Resolve *entity* to a model instance.

    *entity* may already be a model instance, a UUID, or a UUID string.
    *getter* is a callable that accepts a UUID-or-string and returns the
    instance or ``None``.

    :raises ValueError: when the entity cannot be found.
    """
    # If it's not a basic type (UUID / str), assume it's already a model instance.
    if not isinstance(entity, (uuid.UUID, str)):
        return entity
    if isinstance(entity, str):
        entity = uuid.UUID(entity)
    found = getter(entity)
    if found is None:
        raise ValueError(f"{label} not found.")
    return found
```

src/controllers/progress_tracker.py
youennlj-sfl, line 1: This class is esentially meant to be used as a superclass with only two class attributes changed and no method implemented or overriden. It is quite overkill. You could instead remove the singleton pattern and create a proper constructor that accepts `default_phase` and `completed_message` as parameters. ProgressTracker implementing classes (EPSSProgressTracker and NVDProgressTracker) can be deleted and simply replaced by variables:
```py
epss_progress_tracker = ProgressTracker(default_phase="epss_enrichment", completed_message="EPSS enrichment completed successfully")
nvd_progress_tracker = ...
```
Those will be initialized only once (since Python modules are only initialized once per interpreter) sp there is no need for specific singleton code.

youennlj-sfl (changes requested):

WilsonZiweiWang (dismissed):

youennlj-sfl:
LGTM except for the cmd_scan change. I commented on #335

SkytAsul (approved):

youennlj-sfl (dismissed):
Sorry reviewed with wrong user

youennlj-sfl (approved):

WilsonZiweiWang (approved):

PR #340, opened 2026-05-25
Settings: Add field for NVD API key

rawaha-e:

* Settings: Add field for NVD API key
* Add ARIA / roles to settings page for accessibility

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Open the web dashboard, go to the settings page then add / edit / delete the NVD API key.

## Pull Request Checklist

### DIFF
--- a/frontend/src/pages/Explorer.tsx
+++ b/frontend/src/pages/Explorer.tsx
@@ -19,6 +19,16 @@ import Assessments, { removeDuplicateAssessments, STATUS_VEX_TO_GRAPH } from '..
 import Config from "../handlers/config";
 import type { AppConfig } from "../handlers/config";
 
+const tabLabels: Record<string, string> = {
+        metrics: 'Metrics',
+        packages: 'SBOM',
+        vulnerabilities: 'Vulnerabilities',
+        scans: 'Scans',
+        review: 'Review',
+        exports: 'Export',
+        settings: 'Settings',
+};
+
 type Props = {
   darkMode: boolean;
   setDarkMode: (mode: boolean) => void;
@@ -197,16 +207,6 @@ function Explorer({ darkMode, setDarkMode }: Readonly<Props>) {
         }
         setTab(newTab);
     }
-
-    const tabLabels: Record<string, string> = {
-        metrics: 'Metrics',
-        packages: 'SBOM',
-        vulnerabilities: 'Vulnerabilities',
-        scans: 'Scans',
-        review: 'Review',
-        exports: 'Export',
-        settings: 'Settings',
-    };
 
     return (
         <div className="w-screen h-screen bg-gray-200 dark:bg-neutral-800 dark:text-[#eee] flex flex-col overflow-hidden">
--- a/frontend/src/pages/Explorer.tsx
+++ b/frontend/src/pages/Explorer.tsx
@@ -198,19 +198,38 @@ function Explorer({ darkMode, setDarkMode }: Readonly<Props>) {
         setTab(newTab);
     }
 
+    const tabLabels: Record<string, string> = {
+        metrics: 'Metrics',
+        packages: 'SBOM',
+        vulnerabilities: 'Vulnerabilities',
+        scans: 'Scans',
+        review: 'Review',
+        exports: 'Export',
+        settings: 'Settings',
+    };
+
     return (
         <div className="w-screen h-screen bg-gray-200 dark:bg-neutral-800 dark:text-[#eee] flex flex-col overflow-hidden">
-            <NavigationBar
-                key={selectorKey}
-                tab={tab}
-                changeTab={handleTabChange}
-                darkMode={darkMode}
-                setDarkMode={setDarkMode}
-                defaultProject={defaultConfig.project}
-                defaultVariant={defaultConfig.variant}
-                onApply={handleApply}
-            />
+            <a
+                href="#main-content"
+                className="sr-only focus:not-sr-only focus:absolute focus:z-[100] focus:top-2 focus:left-2 focus:px-4 focus:py-2 focus:bg-cyan-800 focus:text-white focus:rounded focus:text-sm focus:font-semibold"
+            >
+                Skip to content
+            </a>
+            <header>
+                <NavigationBar
+                    key={selectorKey}
+                    tab={tab}
+                    changeTab={handleTabChange}
+                    darkMode={darkMode}
+                    setDarkMode={setDarkMode}
+                    defaultProject={defaultConfig.project}
+                    defaultVariant={defaultConfig.variant}
+                    onApply={handleApply}
+                />
+            </header>
 
+            <main id="main-content" aria-label={tabLabels[tab] ?? 'Content'} className="flex-1 flex flex-col overflow-hidden">
             <div className="px-8 pt-4">
                 <MessageBanner
                     type={bannerType}
@@ -221,9 +240,9 @@ function Explorer({ darkMode, setDarkMode }: Readonly<Props>) {
             </div>
 
             {isLoadingData && (
-                <div className="absolute inset-0 z-50 flex items-center justify-center bg-black/40">
+                <div className="absolute inset-0 z-50 flex items-center justify-center bg-black/40" role="status" aria-live="polite">
                     <div className="flex flex-col items-center gap-3 text-white">
-                        <div className="w-10 h-10 border-4 border-white border-t-transparent rounded-full animate-spin"></div>
+                        <div className="w-10 h-10 border-4 border-white border-t-transparent rounded-full animate-spin" aria-hidden="true"></div>
                         <span className="text-sm font-semibold">{loadingMessage}</span>
                     </div>
                 </div>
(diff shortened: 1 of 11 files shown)


ValentinBoudevinSFL (dismissed):
1) I entered a wrong key: 

Image

Shouldn't appear as a success. Please test the key to confirm it's validity

2) I'm excepting to see the key in config.env as explained with the NVD_API_KEY variable

3) NVD API calls doesn't seem to use the key (e.g refresh CVE)

WilsonZiweiWang (changes requested):
Please do not use "Address review comments" as the git commit message.

frontend/src/pages/Settings.tsx
```diff
@@ -283,6 +286,82 @@ function Settings({ onDataChanged, onLoadingMessage }: Readonly<Props>) {
...
+  const handleSaveNvdKey = async () => {
+    setNvdBusy(true);
+    setNvdMsg(null);
+    try {
+      const res = await fetch(import.meta.env.VITE_API_URL + "/api/config/nvd-api-key", {
+        method: "PUT",
+        mode: "cors",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify({ api_key: nvdKeyInput }),
+      });
```
WilsonZiweiWang: Move the request to the corresponding handler just like others.

frontend/src/pages/Settings.tsx
```diff
@@ -283,6 +286,82 @@ function Settings({ onDataChanged, onLoadingMessage }: Readonly<Props>) {
...
+      const data = await res.json();
+      if (!res.ok) {
+        setNvdMsg({ text: data.error || "Failed to save API key", type: "error" });
+      } else {
+        setNvdHasKey(data.has_key);
+        setNvdKeyInput("");
+        // Refresh masked key
+        const r2 = await fetch(import.meta.env.VITE_API_URL + "/api/config/nvd-api-key", { mode: "cors" });
```
WilsonZiweiWang: No need to fire another request just to get the key; you can return it in the previous set request.

frontend/src/pages/Settings.tsx
```diff
@@ -283,6 +286,82 @@ function Settings({ onDataChanged, onLoadingMessage }: Readonly<Props>) {
...
+    setConfirmDeleteNvdKey(false);
+    setNvdBusy(true);
+    setNvdMsg(null);
+    try {
+      const res = await fetch(import.meta.env.VITE_API_URL + "/api/config/nvd-api-key", {
+        method: "PUT",
+        mode: "cors",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify({ api_key: "" }),
+      });
```
WilsonZiweiWang: Same here, move the request function to its own handler file.

frontend/src/pages/Settings.tsx
```diff
@@ -283,6 +286,82 @@ function Settings({ onDataChanged, onLoadingMessage }: Readonly<Props>) {
...
+        setNvdKeyInput("");
+        // Refresh masked key
+        const r2 = await fetch(import.meta.env.VITE_API_URL + "/api/config/nvd-api-key", { mode: "cors" });
+        const d2 = await r2.json();
+        setNvdMaskedKey(d2.masked_key || "");
+        setNvdEditing(false);
+        setNvdMsg({ text: data.has_key ? "API key saved." : "API key removed.", type: "success" });
+      }
+    } catch {
+      setNvdMsg({ text: "Network error", type: "error" });
```
WilsonZiweiWang: All errors that end up here will show a "Network error", which can be misleading. The lazy approach is to log the error message itself using `catch (e)`.

src/routes/config.py
```diff
@@ -54,6 +85,32 @@ def set_nvd_api_key():
...
+                return jsonify({"error": "Invalid NVD API key: rejected by the NVD API."}), 400
+            if status_code not in {200}:
+                return jsonify(
+                    {"error": f"Unexpected NVD API response (HTTP {status_code}). "
+                              "The key could not be confirmed."}
+                ), 502
+
+        # Persist to config.env so the key survives container restarts.
+        _write_config_key('NVD_API_KEY', api_key if api_key else None)
+
```
WilsonZiweiWang, line 147: This write function is run and forgotten. The exception is not handled here, including the write failure.

src/routes/config.py
```diff
@@ -85,26 +87,50 @@ def set_nvd_api_key():
         if data is None or "api_key" not in data:
             return {"error": "Missing 'api_key' field"}, 400
         api_key = data["api_key"].strip()
+        validation_warning = None
 
         if api_key:
+            if not _NVD_API_KEY_PATTERN.fullmatch(api_key):
```
WilsonZiweiWang: I don't think this is actually useful. The API key format can change; the best way to verify its validity is to send a request with it.

WilsonZiweiWang (changes requested):

src/routes/config.py
WilsonZiweiWang: I got a new error where I couldn't save the key. Looks like the default config file is read-only as per the `vulnscout` file at line 222: `-v "$VULNSCOUT_CONFIG_FILE:/etc/vulnscout/config.env:ro,Z"`. This can be verified by inspecting the vulnscout container: 
```
            {
                "Type": "bind",
                "Source": "/mnt/myfiles/projects/vulnscout/.vulnscout/cache/config.env",
                "Destination": "/etc/vulnscout/config.env",
                "Mode": "ro,Z",
                "RW": false,
                "Propagation": "rprivate"
            }
```

src/controllers/vulnerabilities.py
```diff
@@ -544,6 +544,7 @@ def fetch_nvd_data(self):
         tracker.start("nvd_enrichment")
 
         # NVD lookups via API
+        self.nvd_api = NVD_DB(nvd_api_key=os.getenv("NVD_API_KEY"))
```
WilsonZiweiWang: I think we can just use the NVD_API_KEY env by default in the `__init__` of class NVD_DB. So we don't need to run `getenv()` every time.

WilsonZiweiWang (changes requested):

WilsonZiweiWang (commented):
31117ba62716c25df9a6fde1b3189a7da05a444b still has the unclear commit message "Address review comments"

src/routes/config.py
```diff
@@ -2,10 +2,49 @@
...
+        if os.path.exists(config_file):
+            with open(config_file, 'r') as fh:
+                existing_lines = [ln for ln in fh.readlines() if not ln.startswith(f'{key}=')]
+        with open(config_file, 'w') as fh:
+            fh.writelines(existing_lines)
+            if value:
+                fh.write(f'{key}={value}\n')
+        return True
+    except OSError:
+        return False
```
WilsonZiweiWang, line 56: It should use a universal exception here and log the error.

src/routes/config.py
WilsonZiweiWang: @rawaha-e This write permission issue persists.

src/routes/config.py
```diff
@@ -35,3 +76,75 @@ def get_config():
...
+        if not key:
+            return jsonify({"has_key": False, "masked_key": ""})
+        return jsonify({"has_key": True, "masked_key": _mask_nvd_api_key(key)})
+
+    @app.route('/api/config/nvd-api-key', methods=['PUT'])
+    def set_nvd_api_key():
+        data = request.get_json(silent=True)
+        if data is None or "api_key" not in data:
+            return {"error": "Missing 'api_key' field"}, 400
+        api_key = data["api_key"].strip()
```
WilsonZiweiWang, line 103: We need to add a type check here for `api_key`. Only a string is allowed; otherwise, the `strip()` will break.

WilsonZiweiWang (changes requested):

src/routes/config.py
```diff
@@ -2,10 +2,51 @@
...
+
+
+def _write_config_key(key: str, value: str | None) -> bool:
+    """Write or remove *key* in the persistent config.env file.
+
+    If *value* is ``None`` or an empty string the key is removed; otherwise
+    it is written as ``KEY=value``.
+    """
+    config_file = _config_file_path()
+    try:
```
WilsonZiweiWang, line 35: Add a file lock so there is no concurrent file writes.

frontend/src/pages/Explorer.tsx
```diff
@@ -198,19 +198,38 @@ function Explorer({ darkMode, setDarkMode }: Readonly<Props>) {
         setTab(newTab);
     }
 
+    const tabLabels: Record<string, string> = {
```
WilsonZiweiWang: This is trivial, but it is better to move a constant object like this outside the component so it doesn't get recreated during the component lifecycle.

src/controllers/vulnerabilities.py
```diff
@@ -544,6 +544,7 @@ def fetch_nvd_data(self):
         tracker.start("nvd_enrichment")
 
         # NVD lookups via API
+        self.nvd_api = NVD_DB()
```
WilsonZiweiWang: I think it is better to use a local variable here instead of mutating the class member every time the method is called.

WilsonZiweiWang (changes requested):

WilsonZiweiWang (approved):

ValentinBoudevin (approved):
LGTM

PR #353, opened 2026-06-02
summarize vulnerability status across current scope

rawaha-e:

* summarize vulnerability statuses across current scope

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

- Open vulnerability tab.
- Pick vuln with multiple assessments affecting uneven set of packages in current project / variant scope.
- Verify Status column shows summary like `Fixed (2), Pending Assessment (1)` instead of single top status.
- Click status helper icon in table header. Verify tooltip explains scoped aggregation.
- Open Status filter. Select `Pending Assessment` or `Exploitable`. Verify row remains visible if summary contains such status, even when dominant status differs.

## Pull Request Checklist

### DIFF
--- a/frontend/src/handlers/vulnerabilities.ts
+++ b/frontend/src/handlers/vulnerabilities.ts
@@ -99,20 +99,40 @@ const getStatusSummaryEntries = (counts: Record<string, number>): { status: stri
 }
 
 const buildStatusSummary = (assessments: Assessment[]): StatusSummary => {
-    const uniqueStatuses = assessments.reduce((acc, assessment) => {
-        const simplified = assessment.simplified_status || 'unknown';
-        acc.add(simplified);
-        return acc;
-    }, new Set<string>());
+    if (assessments.length === 0) {
+        return {
+            counts: { unknown: 1 },
+            ordered: [{ status: 'unknown', count: 1 }],
+            total_assessments: 0,
+            dominant_status: 'unknown',
+            has_active_status: false,
+        };
+    }
 
-    const counts = Array.from(uniqueStatuses).reduce((acc, status) => {
-        acc[status] = 1;
+    // Group assessments by variant_id. Assessments without a variant_id are
+    // grouped together under a single fallback key so they contribute one slot.
+    const byVariant = new Map<string, Assessment[]>();
+    assessments.forEach((assessment) => {
+        const key = assessment.variant_id ?? '__no_variant__';
+        if (!byVariant.has(key)) byVariant.set(key, []);
+        byVariant.get(key)!.push(assessment);
+    });
+
+    // For each variant keep only the most recent assessment.
+    const latestPerVariant: Assessment[] = [];
+    byVariant.forEach((variantAssessments) => {
+        const latest = variantAssessments.reduce((best, a) =>
+            new Date(a.timestamp).getTime() > new Date(best.timestamp).getTime() ? a : best
+        );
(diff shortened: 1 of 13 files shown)


ValentinBoudevinSFL (changes requested):
Doesn't work I changed a variant from Fixed to Exploitable but still see Fixed(2): 

Image

Also, please remove the quantity of a status (e.g Fixed is enough, no need of Fixed(2)). For the reason that the count is false. I indeed have two variants but I also have 3 packages concerning the vulnerabilities. So I have in reality 2*3 combinations.
You could display Fixed(6) but it wouldn't make any sense for 99% of the users.

ValentinBoudevinSFL (changes requested):
Still doesn't work: 

Image

But it should be both `Fixed`/`Exploitable` as set right here: 

Image

ValentinBoudevin (changes requested):
Image

Works but requires to press F5 to refresh VS once the vulnerability gets a new assessment. 
Should not be the case

ValentinBoudevin (dismissed):
LGTM

ValentinBoudevin (commented):
LGTM

frontend/src/handlers/vulnerabilities.ts
```diff
@@ -55,12 +55,133 @@ type Vulnerability = {
...
+        counts: Record<string, number>;
+        ordered: { status: string; count: number }[];
+        total_assessments: number;
+        dominant_status: string;
+        has_active_status: boolean;
+    };
     assessments: Assessment[];
 };
 
+type StatusSummary = NonNullable<Vulnerability['status_summary']>;
```
youennlj-sfl: Why not doing it the other way around? This is quite convoluted.
```suggestion
    status_summary?: StatusSummary;
    assessments: Assessment[];
};

type StatusSummary = {
    counts: Record<string, number>;
    ordered: { status: string; count: number }[];
    total_assessments: number;
    dominant_status: string;
    has_active_status: boolean;
};
```

frontend/src/handlers/vulnerabilities.ts
```diff
@@ -55,12 +55,133 @@ type Vulnerability = {
     };
     status: string;
```
youennlj-sfl: What is the meaning of this field now we have status_summary? Shouldn't it be removed?

youennlj-sfl (changes requested):
Sorry I'm not very familiar with React development so I find it very hard to review frontend changes. The code is quite hard to follow with few documentation on the methods responsibilities.

I commented on obvious flaws otherwise it looks OK I guess ?

frontend/src/handlers/vulnerabilities.ts
reh-sfl: The status variable was used across multiple sites so I kept it to avoid making the PR much bigger with some mapping to the new structure.
I addressed it in commit 4a5f59a4d2f8d606131fa90416d847589e562690

frontend/src/handlers/vulnerabilities.ts
reh-sfl: Addressed in 90e636601adce54af220929d7358ce8c8b3182b0

ValentinBoudevinSFL (approved):
lgtm

youennlj-sfl (approved):

fshehadeh-sfl (commented):
lgtm

PR #367, opened 2026-06-10
fix: scope affected-pkg list to the latest SBOM

ValentinBoudevin:

* Fix the vulnerability list which joined package name+version
without bounding them to the current scan, so packages from older or
unrelated SBOM documents leaked into a vulnerability's affected-package
list.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Load two projects, both having a package in common (e.g: linux-stm32mp), but with one of them containing a custom supplier. 

Now the suppliers/package list doesn't contaminate the second project anymore in SBOM/Vulnerabilities tab. 

image

image

(Without this fix the variant OpenEmbedded would still appear)

### DIFF
--- a/tests/end_to_end_tests/test_observation_scoping.py
+++ b/tests/end_to_end_tests/test_observation_scoping.py
@@ -230,6 +230,92 @@ def test_observations_not_duplicated_on_second_run(app):
         )
 
 
+def test_affected_pkg_list_excludes_stale_sbom_supplier_variant(app):
+    """GET /api/vulnerabilities?variant_id must not surface supplier-variant
+    packages that only exist in a *stale* (non-current) SBOM document.
+
+    Regression test for the affected-package scoping fix: the supplier
+    expansion joins packages by name+version, so a same-name+version package
+    referenced only by an older SBOM document would otherwise leak into a
+    vulnerability's affected-package list.
+    """
+    import json
+    from datetime import datetime, timedelta, timezone
+
+    from src.bin.merger_ci import _run_main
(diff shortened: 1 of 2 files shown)


reh-sfl (dismissed):

tests/end_to_end_tests/test_observation_scoping.py
```diff
@@ -230,6 +230,89 @@ def test_observations_not_duplicated_on_second_run(app):
...
+        db.session.commit()
+        assert stale_cairo.id != cairo.id
+        SBOMPackage.create(stale_doc.id, stale_cairo.id)
+
+        # The injected stale scan must not become the active SBOM scan.
+        assert Scan.get_latest().id == latest_scan.id
+
+        # --- hit the endpoint ---
+        client = app.test_client()
+        app._INT_SCAN_FINISHED = True  # bypass the "scan not finished" 503 guard
```
WilsonZiweiWang, line 297: Unset this variable when this test is done so it doesn't leak into other tests. I think you can use try-catch-finally.

WilsonZiweiWang (changes requested):

tests/end_to_end_tests/test_observation_scoping.py
ValentinBoudevin: Fixed

reh-sfl (approved):

WilsonZiweiWang (approved):

PR #368, opened 2026-06-10
fix: CVSS scope (scanner or custom)

ValentinBoudevin:

*Standard scanner scores (NVD/Grype/Yocto) are identical for every variant, so storing them per variant duplicated each score N times and leaked other variants' scores into a project view. 
*Custom user scores, in contrast, must stay isolated per variant.

* Metrics.from_cvss now forces variant_id=NULL for any non-custom origin and keeps custom scores variant-scoped. 
* The list/detail endpoints return the union of global scanner rows plus the current variant/project's custom rows (deduped). 
* A migration re-collapses previously per-variant scanner rows back to a single global row. Tests updated to the new contract.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Load multiples projects with multiples variants and the CVSS overlap won't appear anymore

With this fix: 

image

Without this fix: 

image

### DIFF
--- a/src/routes/vulnerabilities.py
+++ b/src/routes/vulnerabilities.py
@@ -152,8 +152,13 @@ def _variant_scoped_metrics_and_effort_overrides(
     overrides: dict[str, _ScopedOverrides] = {}
     for r in records:
         metric_bucket = metrics_by_vuln.get(r.id, {"scoped": [], "fallback": []})
-        selected_metrics = metric_bucket["scoped"] or metric_bucket["fallback"]
+        # CVSS is a union: global (scanner, variant_id IS NULL) scores are
+        # standard data shown for every variant, plus this variant's own
+        # custom scores. Custom scores add to — they don't hide — scanner data.
+        selected_metrics = metric_bucket["fallback"] + metric_bucket["scoped"]
 
+        # Effort is a single value per vuln: a variant-scoped estimate overrides
+        # the global one, falling back to global when none is set.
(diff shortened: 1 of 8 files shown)


src/routes/vulnerabilities.py
```diff
@@ -478,21 +519,38 @@ def _vuln_ids_for_scans(scan_ids):
...
                     )
                     .join(Finding, TimeEstimate.finding_id == Finding.id)
-                    .where(Finding.vulnerability_id.in_(vuln_ids_subq))
+                    .where(Finding.vulnerability_id.in_(vuln_ids_subq), _te_var_filter)
                 ).all()
                 effort_by_vuln: dict[str, tuple] = {}
-                for vid, opti, like, pess in te_rows:
+                _fallback_effort_by_vuln: dict[str, tuple] = {}
+                for vid, te_variant_id, opti, like, pess in te_rows:
+                    if te_variant_id is not None:
```
WilsonZiweiWang, line 546: This is saying any non-null variant. Do we intend to show any variant here without specifying a specific variant ID?

ValentinBoudevin: The query already constrains rows to the project's own variant IDs plus NULL, so "any non-null" means "this project's variant"

reh-sfl (approved):

WilsonZiweiWang (approved):

PR #374, opened 2026-06-11
fix: refresh CVSS gauges after saving custom score

ValentinBoudevin:

When a custom CVSS score is saved from the vulnerability modal while viewing all variants, the PATCH response is variant-scoped and cannot populate the union CVSS gauge view. 

As a result the gauges did not reflect the newly added score until the modal was reopened or the page reloaded.

Re-fetch the vulnerability in the current (project) scope after a successful save so the gauges update immediately, mirroring a reload. The variant-scoped path is unchanged since its PATCH response already carries the correct data.

Add a VulnModal test covering the all-variants re-fetch path, asserting the project-scoped request is issued and patchVuln receives the union CVSS list rather than the single variant-scoped score.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Enter a custom CVSS score (e.g. `CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H`), and now the UI doesn't require to press F5 to appear properly

With this fix: 

image

### DIFF
--- a/frontend/src/components/VulnModal.tsx
+++ b/frontend/src/components/VulnModal.tsx
@@ -1,6 +1,6 @@
 import type { Vulnerability } from "../handlers/vulnerabilities";
 import type { CVSS } from "../handlers/vulnerabilities";
-import { asCVSS, buildStatusSummary } from "../handlers/vulnerabilities";
+import Vulnerabilities, { asCVSS, buildStatusSummary } from "../handlers/vulnerabilities";
 import type { Assessment } from "../handlers/assessments";
 import { asAssessment } from "../handlers/assessments";
 import { escape } from "lodash-es";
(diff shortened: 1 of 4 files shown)


frontend/src/components/VulnModal.tsx
```diff
@@ -812,10 +812,32 @@ type VariantScopedSnapshot = {
...
+                // variant: the PATCH response is already variant-scoped.
                 vuln.severity.cvss = updatedSeverity;
                 patchVuln(vuln.id, vuln);
+            } else if (!variantId) {
+                // All-variants mode: the PATCH response is variant-scoped, so it
+                // can't populate the union gauge view. Re-fetch the vulnerability
+                // in the current (project) scope so the CVSS gauges reflect the new
+                // custom score immediately, mirroring a page reload.
+                try {
+                    const refreshUrl = new URL(
```
WilsonZiweiWang: Move the fetch-related code to the corresponding handler as we always do.

youennlj-sfl (dismissed):

frontend/src/components/VulnModal.tsx
```diff
@@ -812,10 +812,32 @@ type VariantScopedSnapshot = {
...
+                        import.meta.env.VITE_API_URL + `/api/vulnerabilities/${encodeURIComponent(vuln.id)}`,
+                        window.location.href,
+                    );
+                    if (projectId) refreshUrl.searchParams.set('project_id', projectId);
+                    const refreshed = await fetch(refreshUrl.toString(), { mode: 'cors' });
+                    if (refreshed.ok) {
+                        const refreshedVuln = asVulnerability(await refreshed.json());
+                        if (!Array.isArray(refreshedVuln) && Array.isArray(refreshedVuln.severity?.cvss)) {
+                            vuln.severity.cvss = refreshedVuln.severity.cvss;
```
WilsonZiweiWang: `vuln` is the state passed from the parent; we should not mutate it like this. We should either use the setter (`setVuln()`) that comes with it or use a local variable here and pass the local variable to `patchVuln`.

WilsonZiweiWang (changes requested):

frontend/src/components/VulnModal.tsx
```diff
@@ -812,10 +812,32 @@ type VariantScopedSnapshot = {
                 : data?.severity?.cvss;
 
             if (Array.isArray(updatedSeverity) && variantId) {
-                // Only update the local vuln object when viewing a specific variant.
-                // In all-variants mode the snapshot refresh below handles the display.
+                // Only use the response severity directly when viewing a specific
+                // variant: the PATCH response is already variant-scoped.
                 vuln.severity.cvss = updatedSeverity;
```
WilsonZiweiWang: This has the same mutation issue.

WilsonZiweiWang (changes requested):

reh-sfl:
0. using project with All variants
1. open cve, add cvss, it should show up for both variants
2.refresh page, only 1 cvss is displayed

frontend/src/components/VulnModal.tsx
ValentinBoudevin: fixed

frontend/src/components/VulnModal.tsx
ValentinBoudevin: fixed

frontend/src/components/VulnModal.tsx
ValentinBoudevin: fixed

ValentinBoudevin:
> 0. using project with All variants
> 
>     1. open cve, add cvss, it should show up for both variants
>        2.refresh page, only 1 cvss is displayed

fixed

reh-sfl (approved):

WilsonZiweiWang (approved):

PR #377, opened 2026-06-11
fix: replace asserts in code to gracefully exit

ValentinBoudevin:

* Replace the `assert text.packages` , and skip
already-present packages to prevent duplicates. Adds review-endpoint
tests.

* Swap data-dependent `assert`s in the SBOM/vuln parsers and vulnerability
route for graceful handling (skip/fallback), so unexpected real input no
longer raises AssertionError.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### DIFF
--- a/src/routes/_scan_queries.py
+++ b/src/routes/_scan_queries.py
@@ -542,8 +542,10 @@ def fetch_vulnerabilities_texts(
             found = False
             for text in texts:
                 if text.title == obs_key and text.content == obs_desc:
-                    assert text.packages
-                    text.packages.append(pkg)
+                    if text.packages is None:
+                        text.packages = []
+                    if pkg not in text.packages:
+                        text.packages.append(pkg)
                     found = True
                     break
             if not found:
--- a/src/routes/vulnerabilities.py
+++ b/src/routes/vulnerabilities.py
@@ -271,11 +271,11 @@ def _populate_found_by(
             non_dedicated = doc_formats - _DEDICATED_SCANNER_FORMATS
             chosen = non_dedicated if non_dedicated else doc_formats
             for fmt in chosen:
-                assert isinstance(fmt, str)
+                if not isinstance(fmt, str):
+                    continue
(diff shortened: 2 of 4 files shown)


src/views/grype_vulns.py
```diff
@@ -211,7 +211,8 @@ def load_from_dict(self, data: dict):
             if fix_state:
                 for package_id in packages:
                     package = self.packagesCtrl.get(package_id)
-                    assert package
+                    if package is None:
+                        continue
```
youennlj-sfl: Keep assert here. The packages list is populated at lines 169 and 177 which adds packages to the packages controller.This assures that the package exists.
Removing the assert and silently discarding this case will make it harder to debug later if an issue arises.

src/views/openvex.py
```diff
@@ -138,17 +138,17 @@ def to_dict(self, strict_export=False, author=None) -> dict:
...
-                if pkg is not None:
-                    if not pkg.cpe:
-                        pkg.generate_generic_cpe()
-                    if not pkg.purl:
-                        pkg.generate_generic_purl()
-                    assert pkg.cpe and pkg.purl
+                cpe = pkg.cpe[0] if (pkg is not None and pkg.cpe) else (
+                    pkg.generate_generic_cpe() if pkg is not None else "")
+                purl = pkg.purl[0] if (pkg is not None and pkg.purl) else (
+                    pkg.generate_generic_purl() if pkg is not None else "")
```
youennlj-sfl: This is hard to read

youennlj-sfl (changes requested):

src/views/openvex.py
ValentinBoudevin: Done, reverted

src/views/grype_vulns.py
ValentinBoudevin: Done, reverted

youennlj-sfl (approved):

reh-sfl (approved):

PR #379, opened 2026-06-11
fix: Custom Time Estimates appear correctly in Review

ValentinBoudevin:

* fix: list one review time-estimate row per variant

The review time-estimates endpoint keyed its result map by vuln_id alone,
so when several variants had their own estimate for the same vulnerability
they overwrote each other and only one row was returned.

* model fix : allow multiple time estimates per finding

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Create a project with two variants, or more, add a time estimate per variant. 

The Review tab now display all time estimate properly: 

image

### DIFF
--- a/src/helpers/vuln_helpers.py
+++ b/src/helpers/vuln_helpers.py
@@ -75,7 +75,10 @@ def _apply_effort(record, variant_id, opt, lik, pes, log_prefix: str = ""):
             if variant_id is not None:
                 existing = TimeEstimate.get_by_finding_and_variant(finding.id, variant_id)
             else:
-                existing = finding.time_estimate
+                existing = next(
+                    (te for te in finding.time_estimates if te.variant_id is None),
+                    None,
+                )
             if existing is not None:
                 existing.update(optimistic=opt, likely=lik, pessimistic=pes)
             else:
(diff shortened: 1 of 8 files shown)


src/helpers/vuln_helpers.py
```diff
@@ -75,7 +75,10 @@ def _apply_effort(record, variant_id, opt, lik, pes, log_prefix: str = ""):
             if variant_id is not None:
                 existing = TimeEstimate.get_by_finding_and_variant(finding.id, variant_id)
             else:
-                existing = finding.time_estimate
+                existing = next(
+                    (te for te in finding.time_estimates if te.variant_id is None),
+                    None,
+                )
```
youennlj-sfl, line 81: This is a nice example of a bug that would have been caught by the type checker :)

ValentinBoudevin: Can't wait to have it :)

youennlj-sfl (dismissed):

reh-sfl (approved):

WilsonZiweiWang (approved):

PR #389, opened 2026-06-16
feat: export scope SBOM/VEX to a project or variant

ValentinBoudevin:

Exports previously loaded all packages, vulnerabilities and assessments
from every project/variant, so `flask export` and the Review tab download
emitted data unrelated to the current selection. Scope the export pipeline
end-to-end instead.

- Add ExportScope (package-id + variant-id sets) and compute_export_scope()
  in src/helpers/export_scope.py
- Thread an optional scope through ControllersCache into the packages,
  vulnerabilities and assessments controllers, guarding every DB-fallback
  path so a scoped controller never leaks global data. All four views
  (SPDX, SPDX3, CycloneDX, OpenVex) inherit the scope automatically;
  OpenVex now reads assessments via assessmentsCtrl.get_all().
- flask export gains --project/--variant options (missing project/variant
  warns and falls back to a wider scope); entrypoint.sh passes the
  container's PROJECT_NAME/VARIANT_NAME.
- /api/documents/<name> reads variant_id/project_id query params and builds
  a scoped controller for SBOM/VEX exports; reports stay global.
- Frontend: Explorer passes the current variant/project to Exports, which
  forwards them to FileTag so download URLs carry variant_id/project_id.
- Update the OpenVex unit test for the get_all() delegation.

Change-Id: I4f7762296cad601d8d26e62c179a712b7f5f03d4

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Make some assessments in differents projects. Exporting the data in Review tab, won't include all the projects anymore.

The command `./vulnscout --project scarthgap  --export-spdx` export a project correctly

### DIFF
--- a/src/controllers/vulnerabilities.py
+++ b/src/controllers/vulnerabilities.py
@@ -103,9 +103,10 @@ class VulnerabilitiesController:
     safe_url_regex = r"[^a-zA-Z0-9_\-\.]"
     """Regex to remove unsafe characters from URLs."""
 
-    def __init__(self, pkgCtrl: PackagesController):
+    def __init__(self, pkgCtrl: PackagesController, scope=None):
         """Take an instance of PackagesController to resolve package dependencies as parameter."""
         self.packagesCtrl: PackagesController = pkgCtrl
+        self._scope = scope
         self.vulnerabilities: dict[str, Vulnerability] = {}
         """A dictionary of vulnerabilities, indexed by their id."""
         self.alias_registered: dict[str, str] = {}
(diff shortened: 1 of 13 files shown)


src/controllers/vulnerabilities.py
```diff
@@ -140,10 +141,24 @@ def _preload_cache(self) -> None:
...
         try:
             for rec in Vulnerability.get_all():
-                # Populate transient package list from eager-loaded findings
-                for f in (rec.findings or []):
-                    if f.package:
+                # Populate transient package list from eager-loaded findings.
+                # When an export scope is active, only consider findings whose
+                # package is in scope — and skip the vulnerability entirely if
+                # none of its findings touch an in-scope package.
+                allowed = self._scope.package_ids if self._scope is not None else None
```
reh-sfl: Why this is reloaded on each loop? Can't see 'rec' used there.

reh-sfl (dismissed):

src/controllers/vulnerabilities.py
ValentinBoudevin: good catch

WilsonZiweiWang (approved):

reh-sfl (approved):

PR #394, opened 2026-06-19
fix: key package enrichment by UUID and scope variants

ValentinBoudevin:

Package enrichment was keyed by "name@version", which conflated distinct package rows (e.g. different supplier or near-identical versions) and leaked unrelated variants into the SBOM tab.

- Expose id/package_id from Package.to_dict and select Package.id in the enrichment query.
- Key the variants/sources/sbom_documents lookup by package UUID instead of name@version.
- Keep package variants strictly package-scoped on the frontend instead of merging vulnerability-scoped variants.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Context: st-image-core-openstlinux-rt-stm32mp2 is linked with 6.6.116.rt66-stm32mp-rt-r3 and core-image-minimal-stm32mp25-eval, st-image-optenstlinux-stm32mp25-eval with 6.6.116-stm32mp-r3

Before this change list of packages is not link with correct variants:

image

After this change:

image

### DIFF
--- a/src/models/package.py
+++ b/src/models/package.py
@@ -259,6 +259,8 @@ class Package(Base):
 
     def to_dict(self) -> dict:
         return {
+            "id": self.string_id,
+            "package_id": str(self.id) if self.id is not None else None,
             "name": self.name,
             "version": self.version,
             "cpe": list(self.cpe or []),
--- a/frontend/src/handlers/packages.ts
+++ b/frontend/src/handlers/packages.ts
@@ -125,7 +125,10 @@ class Packages {
                 vulnerabilities: counts,
                 maxSeverity: severities,
                 source: [...new Set([...pkg.source, ...vulnerabilities.map((vuln) => vuln.found_by).flat()])],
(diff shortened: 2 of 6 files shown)


src/models/package.py
```diff
@@ -259,6 +259,8 @@ def __contains__(self, item: "Package | str") -> bool:
 
     def to_dict(self) -> dict:
         return {
+            "id": self.string_id,
+            "package_id": str(self.id),
```
reh-sfl: in case the id is not set yet, use
"package_id": str(self.id) if self.id is not None else None

reh-sfl (changes requested):

src/models/package.py
ValentinBoudevinSFL: Done

WilsonZiweiWang (approved):

reh-sfl (approved):

PR #395, opened 2026-06-19
fix: report variants and packages in assessment success message

ValentinBoudevin:

When adding an assessment from VulnModal, the success toast only counted variants and showed a generic message that ignored packages. Track the variant IDs and packages actually touched by the created assessments and build the message from those counts, with correct singular/plural wording.

- "Successfully added assessment to N variants and M packages."
- Falls back to variant-only, package-only, or a plain count message when one dimension is empty.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Before this change, the message displayed is wrong:

image

image

After this change: 

image

### DIFF
--- a/frontend/src/components/VulnModal.tsx
+++ b/frontend/src/components/VulnModal.tsx
@@ -761,6 +761,8 @@ type VariantScopedSnapshot = {
 
         let successCount = 0;
         let lastCasted: Assessment | null = null;
+        const touchedVariantIds = new Set<string>();
+        const touchedPackages = new Set<string>();
 
         setSubmittingMessage('Adding assessment...');
         try {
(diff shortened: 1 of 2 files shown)


WilsonZiweiWang (dismissed):

reh-sfl (approved):

reh-sfl (changes requested):
"Added to N package across M variants" make more sense to me.

ValentinBoudevinSFL:
> "Added to N package across M variants" make more sense to me.

Looks the same to me

reh-sfl:
@ValentinBoudevin 

"Added assessment to N variants and M packages" reads as if variants and packages are two independent targets.

"Added to N packages across M variants" reads more like the assessment was added to packages, with the variant count providing context about where those packages are distributed.

Not a strong objection either way, but I found the latter easier to parse.

ValentinBoudevinSFL:
> @ValentinBoudevin
> 
> "Added assessment to N variants and M packages" reads as if variants and packages are two independent targets.
> 
> "Added to N packages across M variants" reads more like the assessment was added to packages, with the variant count providing context about where those packages are distributed.
> 
> Not a strong objection either way, but I found the latter easier to parse.

Done

reh-sfl (approved):

WilsonZiweiWang (approved):

PR #396, opened 2026-06-19
fix(metrics): count each vulnerability once per status

ValentinBoudevin:

The vulnerability status chart summed per-variant status counts, so a CVE present in multiple variants was counted multiple times and a CVE with mixed statuses inflated several buckets at once.

Count each vulnerability at most once per status bucket (Not affected, Fixed, Pending Assessment, Exploitable) based on whether it has any variant in that status, instead of adding the raw per-variant counts.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Before it accumulates for every variants: 

image

After this change:

image

### DIFF
--- a/frontend/src/pages/Metrics.tsx
+++ b/frontend/src/pages/Metrics.tsx
@@ -354,10 +354,13 @@ const packageColumns = [
                         label: '# of Vulnerabilities',
                         data: vulnerabilities.reduce((acc, vuln) => {
                           const summary = getVulnerabilityStatusSummary(vuln);
-                          acc[0] += summary.counts['Not affected'] ?? 0;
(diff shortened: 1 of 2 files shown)


WilsonZiweiWang:
So how can we tell which variant the pie chart is showing now?

ValentinBoudevin:
> So how can we tell which variant the pie chart is showing now?

 It does not show variants, just the SUM of all status of the current vulnerabilities.

E.g a click on pending assessment should redirect to the same amount of Pending assessment vulns in the review tab

reh-sfl (approved):

WilsonZiweiWang (approved):

PR #397, opened 2026-06-19
fix: Attribute source from exact per-document provenance

ValentinBoudevin:

The found_by source was inferred heuristically from the earliest scan that observed a vulnerability, preferring SBOM formats over dedicated scanners. This wrongly attributed origins when several documents/scanners contributed in the same merge.

- Record a provenance marker (SBOMObservation) per (vulnerability, current SBOM document) at ingestion, deduped within a run.
- Populate found_by from these markers via SBOMDocument.format, falling back to in-scope observing documents for legacy rows without markers.
- Drop the scanner-vs-BOM preference heuristic; map yocto_cve_check explicitly in _FORMAT_TO_FOUND_BY.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Without this PR: 

image

With this PR:

image

### DIFF
--- a/src/routes/vulnerabilities.py
+++ b/src/routes/vulnerabilities.py
@@ -23,6 +23,7 @@ from ..models import (
     Metrics,
     SBOMDocument,
     SBOMPackage,
+    SBOMObservation,
     Iso8601Duration
 )
 from ..helpers.datetime_utils import ensure_utc_iso
@@ -65,11 +66,10 @@ def _sbom_pkg_filter(pkg_ids):
 
 
 # Formats that are exclusively vulnerability scanners (never pure package BOMs)
-_DEDICATED_SCANNER_FORMATS = frozenset({"grype", "yocto_cve_check"})
-
 # Mapping from SBOMDocument.format to the found_by string exposed by the API
 _FORMAT_TO_FOUND_BY: dict[str, str] = {
     "grype": "grype",
+    "yocto_cve_check": "yocto_cve_check",
     "spdx": "spdx3",
     "cdx": "cyclonedx",
     "openvex": "openvex",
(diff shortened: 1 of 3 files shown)


src/controllers/vulnerabilities.py
```diff
@@ -236,6 +238,24 @@ def add(self, vulnerability: Vulnerability) -> Vulnerability:
...
+            marker_key = (vuln_id, current_doc.id)
+            if marker_key in self._source_markers_seen:
+                return
+            self._source_markers_seen.add(marker_key)
+            SBOMObservation.create(
+                vulnerability_id=vuln_id,
+                sbom_document_id=current_doc.id,
+                package_id=None,
+                key="Vulnerability source",
+                description="Detected in this input document",
```
WilsonZiweiWang: The text in the observation created by `_record_source_marker` now shows in the tooltip of the vulns:
image

Is this intended?

ValentinBoudevin: You are right, this part was wrong and removed. Nice catch

WilsonZiweiWang (approved):

WilsonZiweiWang:
I tested it with the following:

[test-scan2.grype.json](https://github.com/user-attachments/files/29262848/test-scan2.grype.json)
[test-scan2.spdx.json](https://github.com/user-attachments/files/29262850/test-scan2.spdx.json)

```
 ./vulnscout --project verify --variant test \
   --add-spdx  example/spdx2/test-scan2.spdx.json \
   --add-grype example/spdx2/test-scan2.grype.json \
   --serve
```

2 CVEs from the Grype file are attributed to spdx3. The misattribution seems to persist.

reh-sfl (approved):

PR #399, opened 2026-06-19
fix: drop OpenEmbedded supplier from packages

ValentinBoudevin:

* Remove supplier OpenEmbedded for Yocto SPDX2 packages. All SBOMS, by default contains the supplier which makes the interface unreadable. 
Dropping the supplier for this specific case does involve any miss information on the user side, as it is the "default" provider

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Tested with the command:

```
./vulnscout --serve --add-spdx example/spdx2/example.rootfs.spdx.tar.zst --add-cve-check example/spdx2/example.rootfs.json
```

Without this change all packages are duplicated on yocto scarhgap:

image

With this change:

image

### DIFF
--- a/src/migrations/versions/q3f4a5b6c7d8_clean_openembedded_suppliers.py
+++ b/src/migrations/versions/q3f4a5b6c7d8_clean_openembedded_suppliers.py
@@ -1,0 +1,154 @@
+"""clean OpenEmbedded/OpenEmnedded supplier values in packages
+
+Revision ID: q3f4a5b6c7d8
+Revises: p2e3f4a5b6c7
+Create Date: 2026-06-18 00:00:00.000000
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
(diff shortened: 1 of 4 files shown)


WilsonZiweiWang:
Now that the packages with the default supplier and the ones with no supplier have the same look in the table column `supplier`. Should we do something about it to distinguish them?

ValentinBoudevinSFL:
> Now that the packages with the default supplier and the ones with no supplier have the same look in the table column `supplier`. Should we do something about it to distinguish them?

We want to remove the distinction, it is volunteer to mix them. The OpenEmbedded supplier from SPDX2 duplicated everything in cve-check which didn't contain it. OpenEmbedded is added as the only hardcoded exception and should not appear anywhere

WilsonZiweiWang (approved):

reh-sfl (approved):

Issue #408, opened 2026-06-26
Import Custom Assessments reports warning


kili236:
Hello everyone,

I tried to do an custom assessment import, but ended up with the following warning:
`Warning: {'vuln_id': 'CVE-2026-5450', 'package': 'generic/glibc@2.39', 'error': 'Multiple rows were found when one or none was required'}`
The assessment from the specific CVE is then not taken into account.

Small note, I got the assessment file by calling `--export-custom-assessments`.

Thank you

ValentinBoudevin:
Addressed in the same PR #410 

Thank you for your return

kili236:
I tested it with the lastest version. It works great.

Thanks

Issue #407, opened 2026-06-26
Import Custom Assessments not working using docker exec


kili236:
Hello everyone,
I am working on a yocto layer using the meta-vulnscout layer. I noticed the following issue with `--import-custom-assessments` when using directly the command `docker exec vulnscout entrypoint.sh`.

#### Description
The filename must start by `vulnscout_stage_`, so a rename occurs and the file is added to the argument list. If no rename is performed, the file name is never added to the arguments list.

#### Commit introducing the bug
I guess this was introduced in [this commit.](https://github.com/savoirfairelinux/vulnscout/commit/983881dfa4a754a55a9f13e72b259bfe90bdf7fd#diff-b085bf173708890f767512dee6bcf4737c31e3f038639176fe6679abee5b5593)

#### Workaround
Rename the file `vulnscout_stage_<filename>`

#### Potential fix (not tested)
- line 473, update the file variable instead: `import_args+=("$dest_file")` -> `file=$dest_file`
- always add the file argument: add `import_args+=("$file")` after line 433. 

Thank you again for the work and the tool

ValentinBoudevin:
I will look at it today. Also, I just bump yocto layer from VulnScout v0.13 to v0.17. 

This will add a lot of improvements every VS feature

ValentinBoudevin:
To give you an update, we will perform a v0.17.1 release on Moday to fix both your issues.

Then, I also added in the backlog to do a pass on meta-vulnscout to add the missing commands as --import-custom-assessments

ValentinBoudevinSFL:
Both issues should be solved with 0.17.1. Please take a look at #410 if you need any example with the import/export command.

kili236:
somehow, setting `VULNSCOUT_IMAGE_VERSION = "v0.17.1" did take the v0.17 instead of v0.17.1, so I had to use `latest`. But it is fixed thanks!

PR #410, opened 2026-06-26
Fix the import-custom-assessments feature

ValentinBoudevin:

Based on the issue: https://github.com/savoirfairelinux/vulnscout/issues/408
Based on the issue: https://github.com/savoirfairelinux/vulnscout/issues/407

* Fix the `import-custom-assessments` CLI only accepted OpenVEX documents for
.json input, rejecting the web "export custom data" format ({version, assessments, cvss, time_estimates}) with "not a valid OpenVEX document".

* Restrict the dedup query to origin == "custom" so imports only collapse
against existing custom assessments, allowing deleted custom
assessments to be restored.

* The DELETE /api/assessments/<id> route remove the row from the DB

* scalar_one_or_none() raised "Multiple rows were found" when several
assessments shared the same finding/variant/status, dropping the item
with a warning. The dedup only needs existence, so .scalars().first() is
correct.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

1) Generate two assessments on the same vulnerability/package/version with the same status and export it in Review tab
2) Delete the custom assessments
3) In another terminal, export/import the custom assessments via `docker exec` command

```
docker exec vulnscout /scan/src/entrypoint.sh --project wrynose --export-custom-assessments
```
Output:

```
INFO  [alembic.runtime.migration] Context impl SQLiteImpl.
INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
Custom assessments exported: /scan/outputs/core-image-minimal-stm32mp25-eval.json
```

```
docker exec vulnscout /scan/src/entrypoint.sh --project wrynose --variant core-image-minimal-stm32mp25-eval --import-custom-assessments /scan/outputs/core-image-minimal-stm32mp25-eval.json
 ```

Should be successfully re-imported:

```
INFO  [alembic.runtime.migration] Context impl SQLiteImpl.
INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
Imported 0 assessments (3 skipped as duplicates)
```

### DIFF
--- a/src/helpers/assessment_io.py
+++ b/src/helpers/assessment_io.py
@@ -296,6 +296,7 @@ def import_statements(
                         DBAssessment.finding_id == finding.id,
                         DBAssessment.variant_id == variant_id,
                         DBAssessment.status == status,
+                        DBAssessment.origin == "custom",
                     )
                 ).scalars().first()
                 if existing is not None:
--- a/src/helpers/assessment_io.py
+++ b/src/helpers/assessment_io.py
@@ -747,6 +747,7 @@ def import_custom_data(
                             DBAssessment.finding_id == finding.id,
                             DBAssessment.variant_id == target_variant_id,
                             DBAssessment.status == status,
+                            DBAssessment.origin == "custom",
                         )
                     ).scalar_one_or_none()
                     if existing is not None:
--- a/src/helpers/assessment_io.py
+++ b/src/helpers/assessment_io.py
@@ -297,7 +297,7 @@ def import_statements(
                         DBAssessment.variant_id == variant_id,
                         DBAssessment.status == status,
                     )
-                ).scalar_one_or_none()
+                ).scalars().first()
                 if existing is not None:
                     skipped += 1
                     continue
@@ -749,7 +749,7 @@ def import_custom_data(
                             DBAssessment.status == status,
                             DBAssessment.origin == "custom",
                         )
-                    ).scalar_one_or_none()
+                    ).scalars().first()
                     if existing is not None:
                         result["assessments_skipped"] += 1
                         continue
--- a/src/bin/cmd_assessments.py
+++ b/src/bin/cmd_assessments.py
@@ -18,6 +18,7 @@ from ..helpers.assessment_io import (
     import_statements,
     import_archive_bytes,
     import_directory,
+    import_custom_data,
 )
 from ..models.assessment import Assessment as DBAssessment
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Review by reh-sfl — DISMISSED** [review]

**Review by WilsonZiweiWang — COMMENTED** [review]
## Code Review Finding

### 🟡 Missing `origin == "custom"` filter in `import_statements` dedup

**File:** `src/helpers/assessment_io.py` ~line 294–300

The PR correctly adds `DBAssessment.origin == "custom"` to the dedup query inside `import_custom_data` (around line 747), but the **identical dedup query in `import_statements` is missing this filter**.

`import_statements` is the code path invoked when re-importing an **OpenVEX `.json`** file — which is exactly what `--export-custom-assessments` produces. Without this guard, the restore cycle via OpenVEX will still incorrectly skip a deleted custom assessment whenever a scanner assessment shares the same finding/variant/status combination.

**Current code (unchanged by this PR):**
```python
existing = db.session.execute(
    db.select(DBAssessment).where(
        DBAssessment.finding_id == finding.id,
        DBAssessment.variant_id == variant_id,
        DBAssessment.status == status,
        # ← missing: DBAssessment.origin == "custom"
    )
).scalars().first()   # .scalars().first() fix is applied, but origin filter is not
```

**Suggested fix:** Add `DBAssessment.origin == "custom"` to the `where()` clause, mirroring the fix already applied in `import_custom_data`:
```python
existing = db.session.execute(
    db.select(DBAssessment).where(
        DBAssessment.finding_id == finding.id,
        DBAssessment.variant_id == variant_id,
        DBAssessment.status == status,
        DBAssessment.origin == "custom",  # ← add this
    )
).scalars().first()
```

A matching test case (similar to `test_import_custom_data_not_skipped_when_only_scanner_assessment_exists`) should be added for the `import_statements` code path as well.

src/helpers/assessment_io.py
```diff
@@ -297,7 +297,7 @@ def import_statements(
                         DBAssessment.variant_id == variant_id,
                         DBAssessment.status == status,
```
WilsonZiweiWang, line 298: Should you add ` DBAssessment.origin == "custom"`  just like what you did below in `import_custom_data()`?

ValentinBoudevin: Done

WilsonZiweiWang (approved):

reh-sfl (approved):

PR #412, opened 2026-06-26
ENISA EUVD support

reh-sfl:

The following is synced through the ENISA dump API endpoints:
* Full CVE - EUVD mapping: each vuln will have its EUVD attached
* KEV (Known Exploitable Vulnerability): marks specific vuns as exploiated in the wild 

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Web dashboard: go to Vulnerabilities tab, select vulns then refresh with EUVD.

## Pull Request Checklist

### DIFF
--- a/frontend/src/pages/TableVulnerabilities.tsx
+++ b/frontend/src/pages/TableVulnerabilities.tsx
@@ -21,6 +21,8 @@ import NVDProgressHandler from "../handlers/nvd_progress";
 import EPSSProgressHandler from "../handlers/epss_progress";
 import GHSAProgressHandler from "../handlers/ghsa_progress";
 import type { GHSAProgress } from "../handlers/ghsa_progress";
+import EUVDProgressHandler from "../handlers/euvd_progress";
+import type { EUVDProgress } from "../handlers/euvd_progress";
 
 type SourceBanner = { message: string; type: 'error' | 'success' } | null;
 
@@ -365,10 +367,12 @@ function TableVulnerabilities ({ vulnerabilities, filterLabel, filterValue, appe
     const [nvdProgress, setNvdProgress] = useState<NVDProgress | null>(null);
     const [epssProgress, setEpssProgress] = useState<EPSSProgress | null>(null);
     const [ghsaProgress, setGhsaProgress] = useState<GHSAProgress | null>(null);
+    const [euvdProgress, setEuvdProgress] = useState<EUVDProgress | null>(null);
     const [selectedRows, setSelectedRows] = useState<RowSelectionState>({});
     const [nvdBanner, setNvdBanner] = useState<SourceBanner>(null);
     const [epssBanner, setEpssBanner] = useState<SourceBanner>(null);
     const [ghsaBanner, setGhsaBanner] = useState<SourceBanner>(null);
+    const [euvdBanner, setEuvdBanner] = useState<SourceBanner>(null);
     const [generalBanner, setGeneralBanner] = useState<SourceBanner>(null);
     const [searchFilteredData, setSearchFilteredData] = useState<Vulnerability[]>([]);
     const [visibleColumns, setVisibleColumns] = useState<string[]>([
@@ -404,6 +408,9 @@ function TableVulnerabilities ({ vulnerabilities, filterLabel, filterValue, appe
     const prevGhsaInProgress = useRef<boolean | null>(null);
     const prevGhsaPhase = useRef<string | null>(null);
     const prevGhsaStartedAt = useRef<string | null>(null);
+    const prevEuvdInProgress = useRef<boolean | null>(null);
+    const prevEuvdPhase = useRef<string | null>(null);
+    const prevEuvdStartedAt = useRef<string | null>(null);
     const hasFetchedProgressOnce = useRef(false);
 
     const keyboardShortcuts = [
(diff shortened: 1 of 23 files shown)


src/migrations/versions/s5b6c7d8e9f0_add_euvd_enrichment.py
```diff
@@ -0,0 +1,36 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'r4a5b6c7d8e9'
+down_revision = 'q3f4a5b6c7d8'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    with op.batch_alter_table('vulnerabilities', schema=None) as batch_op:
```
ValentinBoudevin, line 27: Can´t we create a new table? I think Vulnerability table is big enough already

ValentinBoudevin (changes requested):
Previous comment is still valid. A PR/commit for cleanup vulnerability tab is required before merging this PR.

Also, --perform-euvd-scan doesn't exist in this code anymore. Can you update the PR description?

Concerning the feature itself:

-EUVD files should not be stored at the root of the cache
-Doesn't work on 20000 cves directly due to the current limitations in bulk_refresh.py

```
_EPSS_BATCH_SIZE = 100
_NVD_COMMIT_EVERY = 50
# HIGH: cap prevents unbounded background threads (no API key = 6 s/CVE × N seconds of work)
_MAX_CVE_IDS = 1000
# HIGH: only accept well-formed CVE identifiers to avoid wasting rate-limit quota
_CVE_RE = re.compile(r'^CVE-\d{4}-\d{4,}$')
_MAX_GHSA_IDS = 500
_GHSA_RE = re.compile(r'^GHSA-[A-Z0-9]{4}-[A-Z0-9]{4}-[A-Z0-9]{4}$')
_GHSA_COMMIT_EVERY = 20
_GHSA_SLEEP_INTERVAL = 1.0
```

_MAX_CVE_IDS doesn't apply to you.


But I don't see this part "KEV (Known Exploitable Vulnerability): marks specific vuns as exploiated in the wild" with a new column

src/models/vulnerability.py
```diff
@@ -617,16 +754,36 @@ def update_record(
             self.versions_data = versions_data
         if patch_url is not None:
             self.patch_url = patch_url
-        if nvd_last_modified is not None:
```
ValentinBoudevinSFL, line 620: Why is it removed?

src/models/vulnerability.py
```diff
@@ -605,10 +746,6 @@ def update_record(
             self.attack_vector = attack_vector
         if epss_score is not None:
             self.epss_score = Decimal(epss_score)
-        if epss_fetched_at is not None:
```
ValentinBoudevinSFL, line 608: Why is it removed?

src/models/vulnerability.py
```diff
@@ -55,17 +57,15 @@ class Vulnerability(Base):
     publish_date: Mapped[date | None]
     attack_vector: Mapped[str | None] = mapped_column(Text)
     epss_score: Mapped[Decimal | None]
-    epss_fetched_at: Mapped[datetime | None]
-    epss_data_updated_at: Mapped[datetime | None]
     links: Mapped[list | None] = mapped_column(JSON)
     weaknesses: Mapped[list | None] = mapped_column(JSON)
     versions_data: Mapped[dict | None] = mapped_column(JSON)
     patch_url: Mapped[list | None] = mapped_column(JSON)
-    nvd_last_modified: Mapped[str | None] = mapped_column(Text)
```
ValentinBoudevinSFL, line 64: Why is it removed?

src/migrations/versions/s5b6c7d8e9f0_add_euvd_enrichment.py
```diff
@@ -0,0 +1,36 @@
...
+        batch_op.add_column(sa.Column('euvd_id', sa.String(length=50), nullable=True))
+        batch_op.add_column(sa.Column('euvd_known_exploited', sa.Boolean(), nullable=True))
```
ValentinBoudevinSFL, line 29: I don't think we should do migration in this order. 
Why do we create new fields in Vulnerability to directly move them in vuln_refresh?

Better create vuln_refresh first, and then add new euvd fields in it

ValentinBoudevinSFL (changes requested):

src/migrations/versions/s5b6c7d8e9f0_add_euvd_enrichment.py
reh-sfl: This is part of the first commit where we implement EUVD, it's separate from the transition to a new table.
They nevertheless form a logical sequence to avoid breaking any existing setup.

src/models/vulnerability.py
reh-sfl: Because now it's handled by vln_refresh model.

src/models/vulnerability.py
reh-sfl: Because now it's handled by vln_refresh model.

src/models/vulnerability.py
ValentinBoudevinSFL: Then why vln_refresh doesn´t handle it too?

```
     # ENISA EUVD (European Vulnerability Database) enrichment
    euvd_id: Mapped[str | None] = mapped_column(String(50))
    euvd_known_exploited: Mapped[bool | None]
    euvd_kev_sources: Mapped[list | None] = mapped_column(JSON)
    euvd_date_added: Mapped[str | None] = mapped_column(String(32))
```

ValentinBoudevinSFL (changes requested):
LGTM for the fetch phase.

But, I don´t see the KEV column as described in the PR description.

And as I said euvd_cve_mapping.csv and euvd_kev.json should not be stored at the root of the cache

src/models/vulnerability.py
ValentinBoudevinSFL: Then why vlns_refresh doesn´t handle EUVD too?

```
       if euvd_id is not None:
            self.euvd_id = euvd_id
            self.add_alias(euvd_id)  # keep transient aliases in sync
        if euvd_known_exploited is not None:
            self.euvd_known_exploited = euvd_known_exploited
        if euvd_kev_sources is not None:
            self.euvd_kev_sources = euvd_kev_sources
        if euvd_date_added is not None:
            self.euvd_date_added = euvd_date_added
        # Refresh fields — delegated to the ``vuln_refresh`` sibling row.
```

reh-sfl:
Note that KEV column is present but hidden by default.

ValentinBoudevinSFL:
> Note that KEV column is present but hidden by default.

image

Am I missing something?

src/models/vulnerability.py
reh-sfl: This was fixed by changing the sequence of migrations: move previous fields to a vuln_refresh table then adding the EUVD in the next commit.

src/models/vulnerability.py
reh-sfl: Same as above

src/migrations/versions/s5b6c7d8e9f0_add_euvd_enrichment.py
reh-sfl: Same as above

reh-sfl:
@ValentinBoudevin Updated and verified the commit
1. Migration changes
2. Column is showing all the times to avoid confusion
3. EUVD cache moved to its own euvd folder

ValentinBoudevin (dismissed):
LGTM I just want to enable EU KEV by default but can be re discussed in another PR.

I want to do a second PR to discuss about encouraging users to sync with EUVD

frontend/src/pages/TableVulnerabilities.tsx
WilsonZiweiWang, line 1: EUVD refresh timestamps are not reflected in "Last Fetched"/"Last Updated" columns.

WilsonZiweiWang (approved):

ValentinBoudevinSFL (approved):

PR #413, opened 2026-06-29
fix: Remove Deprecated code to force YML update

ValentinBoudevin:

Docker YAML system doesn't exist since few months, and it seems safe to remove this condition right now.
I got it triggered for no reason. Can be safely removed

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

This PR just delete a deprecated temporary code


reh-sfl:
I think we still need to have this legacy verification because we don't know when users can be expected to upgrade their VS version, especially with enterprise environment where new version rollout are subject to qualification (e.g migration scripts in Firefox ESR).

Otherwise we should also remove the migration script as well as it has no direct use anymore given the context (and many more locations that directly relates to this feature).

ValentinBoudevin:
> I think we still need to have this legacy verification because we don't know when users can be expected to upgrade their VS version, especially with enterprise environment where new version rollout are subject to qualification (e.g migration scripts in Firefox ESR).
> 
> Otherwise we should also remove the migration script as well as it has no direct use anymore given the context (and many more locations that directly relates to this feature).

Removed the migration.sh script

WilsonZiweiWang (changes requested):
1. The vulnscout script still has `_detect_legacy()`, it should be removed as well.
2. Remove `notifications.py`, unregister the route in `src/routes/__init__.py`, and remove the frontend polling in `App.tsx`.

ValentinBoudevin:
> 1. The vulnscout script still has `_detect_legacy()`, it should be removed as well.
> 
>     2. Remove `notifications.py`, unregister the route in `src/routes/__init__.py`, and remove the frontend polling in `App.tsx`.

Nice catch, done

WilsonZiweiWang (approved):

reh-sfl (approved):

PR #418, opened 2026-07-01
fix(docs): change update and version command syntax

Allegra42:

Removes the leading dashes from `--update` to match the implementation. Also changes it for `--version` which accepts both, `--version` and `version`, for consistency.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Test the commands using the CLI interface.

## Pull Request Checklist

- [X] Linked relevant issues (if any)

### DIFF
--- a/vulnscout
+++ b/vulnscout
@@ -46,8 +46,8 @@ Container lifecycle:
   start                           Start VulnScout container (auto-started by other commands)
   stop                            Stop and remove VulnScout container
   restart                         Restart VulnScout container
-  --update                        Pull latest image and restart if a new version is available
-  --version                       Show current version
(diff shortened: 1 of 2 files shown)


vulnscout
kili236, line 400: To be checked with the devs, but another solution is to add `--update` here.

kili236, line 395: should this be removed then?

ValentinBoudevinSFL: Yes, I prefer to add the missing -- commands for consistency

ValentinBoudevin:
Fixed in v0.18 with the following PR: https://github.com/savoirfairelinux/vulnscout/pull/437

PR #420, opened 2026-07-01
fix: published-date filter when dates exist

ValentinBoudevin:

The published-date filter was disabled whenever NVD had not completed a sync, even though published dates can come from other sources (e.g. an sbom-cve-check scan)..

- TableVulnerabilities: compute hasAnyPublishedDate from the loaded vulnerabilities and pass it to PublishedDateFilter.
- PublishedDateFilter: disable only while NVD is actively syncing, or when no published-date data exists at all. Update the tooltip to distinguish "NVD sync in progress" from "No published dates available".

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Load Published date via any other way (e.g: sbom-cve-check), then the Published Date button is available as a valid filter

### DIFF
--- a/frontend/src/pages/TableVulnerabilities.tsx
+++ b/frontend/src/pages/TableVulnerabilities.tsx
@@ -158,17 +158,21 @@ type PublishedDateFilterProps = {
     setDateFrom: (value: string) => void;
     setDateTo: (value: string) => void;
     nvdProgress: NVDProgress | null;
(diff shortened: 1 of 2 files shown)


WilsonZiweiWang (approved):

QuentinL-SFL (approved):
LGTM

PR #421, opened 2026-07-01
Fix: Reduce post-assessment loading time from twenty sec to less than a second

ValentinBoudevin:

- Batch multi-variant adds into one request. Post all variants in a single /api/assessments/batch call to the backend
- Remove legacy OpenVEX file-writing artifact. Drop the auto-written /scan/outputs/openvex.json that predates the
database implementation.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Setup:

```
docker build -t sfl:assess
export VULNSCOUT_IMAGE="sfl:assess"
./vulnscout stop
./vulnscout --serve --dev
```

Create/Edit an assessment. The loading time goes from 20 seconds (without this fix) to less than a second (with this change)

### DIFF
--- a/frontend/src/components/VulnModal.tsx
+++ b/frontend/src/components/VulnModal.tsx
@@ -766,50 +766,54 @@ type VariantScopedSnapshot = {
 
         setSubmittingMessage('Adding assessment...');
         try {
-        for (const vid of variantIds) {
-            const body = vid ? { ...baseContent, variant_id: vid, timestamp: sharedTimestamp } : { ...baseContent, timestamp: sharedTimestamp };
-            const response = await fetch(import.meta.env.VITE_API_URL + `/api/vulnerabilities/${encodeURIComponent(vuln.id)}/assessments`, {
-                method: 'POST',
-                mode: 'cors',
-                headers: { 'Content-Type': 'application/json' },
-                body: JSON.stringify(body)
-            });
-            const data = await response.json();
-            if (data?.status === 'success') {
-                // Backend returns an array (one record per package); support legacy single too
-                const rawList: unknown[] = Array.isArray(data?.assessments)
-                    ? data.assessments
-                    : (data?.assessment ? [data.assessment] : []);
-                for (const raw of rawList) {
-                    const casted = asAssessment(raw);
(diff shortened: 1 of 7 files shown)


ososeitua (dismissed):
Works well for me!

WilsonZiweiWang:
Can you clarify the steps to reproduce the setup in which creating/editing the assessment takes 20 seconds? My current setup uses a small SBOM. The assessment editing happens immediately.

ValentinBoudevinSFL:
> Can you clarify the steps to reproduce the setup in which creating/editing the assessment takes 20 seconds? My current setup uses a small SBOM. The assessment editing happens immediately.

Uses the default project provided on Redmine. On Scarthgap is takes a while to compute. 
In every case the save_openvex is super slow and deprecated

frontend/src/components/VulnModal.tsx
```diff
@@ -766,50 +766,55 @@ type VariantScopedSnapshot = {
...
-                            }, 5500);
-                        }
-
-                        appendAssessment(casted);
-                        vuln.assessments.push(casted);
-                        // Keep allVulnAssessments in sync so variant tags appear immediately
-                        setAllVulnAssessments(prev => [...prev, casted]);
-                        vuln.simplified_status = casted.simplified_status;
+        // Post every variant in a single batch request so the backend regenerates the
+        // OpenVEX document once for the whole action instead of once per variant.
```
reh-sfl: I am confused, do we still generate openvex? Because it was removed in backend code.

reh-sfl (changes requested):

frontend/src/components/VulnModal.tsx
ValentinBoudevinSFL: No, we don´t it is deprecated since DB

reh-sfl: In such case you need to modify that comment.

ValentinBoudevinSFL: Done

reh-sfl (approved):

ValentinBoudevinSFL (approved):

PR #422, opened 2026-07-01
Fix: Reduce time to load Scan tab

ValentinBoudevin:

-Speed up scan-tab list serialization. The scan history list recomputed global results per scan, issuing a JOIN
query per contributing-scan set plus separate observation scans for findings and vulns.

-Cache scan-tab list and finish prefetch refactor. The scan history list is expensive to serialize (per-scan global-result
diffs). Its content only changes when scans are imported, deleted or
renamed, so recomputing it on every request is wasteful.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Go on the Scan tab, should be faster to load the first time. Then, move to another tab and come back to Scan tab. 
Should be very fast thanks to the cache

### DIFF
--- a/src/routes/_scan_diff.py
+++ b/src/routes/_scan_diff.py
@@ -261,6 +261,7 @@ def _global_result_id_sets(
     tool_scans: Dict[str, Scan],
     *,
     filter_tool_by_sbom_pkgs: bool = False,
+    _cache: Optional[dict] = None,
 ) -> Tuple[set[uuid.UUID], set[str], set[uuid.UUID]]:
     """Return ``(finding_ids, vuln_ids, package_ids)`` for a global result.
 
@@ -287,6 +288,13 @@ def _global_result_id_sets(
         s.id for s in tool_scans.values()
     ]
     contributing_ids = list(dict.fromkeys(contributing_ids))
+
+    cache_key = None
+    if _cache is not None:
+        cache_key = (frozenset(contributing_ids), filter_tool_by_sbom_pkgs)
+        cached = _cache.get(cache_key)
+        if cached is not None:
+            return cached
 
     sbom_pkg_ids = _packages_by_scan_ids([sbom_scan.id]).get(
         sbom_scan.id, set()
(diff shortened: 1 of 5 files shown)


src/routes/_scan_diff.py
```diff
@@ -616,14 +715,61 @@ def _global_result_full(
...
+    )
+
+
+def serialize_list_with_diff_cached(scope_key: str, scans: list[Scan]) -> List[dict]:
+    """Return the serialised scan-history list for *scans*, cached per *scope_key*.
+
+    The heavy diff computation runs once and its result is reused for later
+    requests until the underlying set of scans changes (import / delete /
+    rename), matching an application reload.
+    """
```
WilsonZiweiWang, line 737: When an assessment is updated/deleted (`src/routes/assessments.py`, `update_assessment()`, `delete_assessment()`), the scan is not touched. The fingerprint stays the same. But the cache should be invalidated in these cases, no?

ValentinBoudevinSFL: You are right, if the user delete a SBOM assessment, then it invalidates the cache/. 
But custom assessment don´t affect it. 

I will rework this PR

src/routes/_scan_diff.py
```diff
@@ -616,14 +715,77 @@ def _global_result_full(
...
 
+# In-memory cache of the (expensive) serialised scan-history list, keyed by
+# request scope ("all", "project:<id>", "variant:<id>").  The cache is keyed on
+# a fingerprint of the scan rows (id / description / timestamp), so it is
+# recomputed automatically when scans are imported, deleted or renamed.  The
+# fingerprint does NOT cover the per-scan assessment counts shown in the view:
+# those include automated (non-custom) assessments, so editing or deleting such
+# an assessment must invalidate the cache explicitly (see
+# ``invalidate_scan_list_cache``).  Custom (manually-created) assessments are
+# excluded from the view and therefore never require invalidation.
```
QuentinL-SFL: LGTM but we should not left these kind of comment in the code. 
It need to be shrink

QuentinL-SFL (changes requested):

src/routes/_scan_diff.py
ValentinBoudevin: Done

src/routes/_scan_diff.py
ValentinBoudevin: Done in the thrid commit

QuentinL-SFL (approved):

WilsonZiweiWang (approved):

PR #424, opened 2026-07-01
Feature: Add variant filtering

ValentinBoudevin:

* Add new filters on SBOM, Vulnerability, and Review tab. Uncheck a variant would remove all elements which are only affected by variant inactives 

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Play with the variant filter in SBOM, Vulnerability, and Review tab
image

### DIFF
--- a/frontend/src/pages/TablePackages.tsx
+++ b/frontend/src/pages/TablePackages.tsx
@@ -1,6 +1,6 @@
 import type { Package, VulnCounts, Severities } from "../handlers/packages";
 import { createColumnHelper, Row } from '@tanstack/react-table'
-import { useMemo, useState, useRef, useEffect } from "react";
+import { useMemo, useState, useRef, useEffect, useCallback } from "react";
 import SeverityTag from "../components/SeverityTag";
 import TableGeneric from "../components/TableGeneric";
 import debounce from 'lodash-es/debounce';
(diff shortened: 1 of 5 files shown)


frontend/src/pages/TablePackages.tsx
```diff
@@ -52,6 +52,7 @@ function TablePackages({ packages, onShowVulns }: Readonly<Props>) {
     const [selectedSources, setSelectedSources] = useState<string[]>([]);
     const [selectedSbomDocs, setSelectedSbomDocs] = useState<string[]>([]);
     const [selectedSuppliers, setSelectedSuppliers] = useState<string[]>([]);
+    const [selectedVariants, setSelectedVariants] = useState<string[]>([]);
```
WilsonZiweiWang: I noticed a first-render flash on the package page and the vuln page. It is likely because `selectedVariants` starts with an empty array and the rows are hidden at the first render. Then `selectedVariants` is populated within the `useEffect`, and the rows are shown again, causing a visual flash.

QuentinL-SFL (dismissed):
LGTM

frontend/src/pages/TablePackages.tsx
ValentinBoudevin: Done

QuentinL-SFL (dismissed):

ososeitua:
It might be worth explaining the difference between this variant filter and the variant selector somewhere. But otherwise it looks good to me!

frontend/src/pages/TablePackages.tsx
WilsonZiweiWang: The same fix should be applied to the Review page as well.

ValentinBoudevin: Done

WilsonZiweiWang (approved):

ValentinBoudevinSFL (approved):

PR #429, opened 2026-07-03
feat(search): add only:<term> operator to filter by every affected package

ValentinBoudevin:

Introduce an "only:<regex>" search operator that keeps a row only when every value returned by the new forAllValues accessor matches the given case-insensitive regex. Rows with no affected values are excluded, and an invalid regex falls back to a literal match.

Wire the operator into the Vulnerabilities tab (SBOM-affected packages), the SBOM/Packages tab (package name) and the Review tab (affected packages), and document it in each search-syntax help panel.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

image

image

### DIFF
--- a/frontend/src/components/TableGeneric.tsx
+++ b/frontend/src/components/TableGeneric.tsx
@@ -22,6 +22,13 @@ type Props<DataType> = {
     hasPagination?: boolean;
     onFilteredDataChange?: (filteredData: DataType[]) => void;
     onFocusedRowChange?: (rowIndex: number | null) => void;
+    /**
+     * Values checked by the `only:<term>` search operator. When provided, a row
+     * is kept only if EVERY string returned here satisfies the (optionally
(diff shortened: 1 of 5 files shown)


ososeitua:
- I think the description could be more clear to explain that "only rows where _all_ entries contain 'native' will be shown"
- "regex" was a bit confusing in the description because I thought it supports regex-like searches with "only", but that doesn't seem to be the case

ValentinBoudevin:
> * I think the description could be more clear to explain that "only rows where _all_ entries contain 'native' will be shown"
> 
>     * "regex" was a bit confusing in the description because I thought it supports regex-like searches with "only", but that doesn't seem to be the case

Done

QuentinL-SFL (approved):
LGTM

ososeitua (approved):
lgtm

PR #432, opened 2026-07-06
fix(table): stop filtered rows blinking

ValentinBoudevin:

The filteredData memo listed forAllValues in its dependency array, but every table passes it as an inline arrow, so its identity changed on each render. Any unrelated re-render (e.g. showing a hover tooltip over an ID cell) re-ran fuse.search and returned a new array, making the virtualized table constantly re-render/re-measure.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Type a word in the search bar and overlay a vulnerability in "ID" column. 

With this fix, it doesn´t blink anymore and the vulnerability can be correctly open with a click on it

image

### DIFF
--- a/frontend/src/components/TableGeneric.tsx
+++ b/frontend/src/components/TableGeneric.tsx
@@ -68,6 +68,9 @@ function TableGeneric<DataType> ({
         });
     }, [fuseKeys, data]);
 
+    const forAllValuesRef = useRef(forAllValues);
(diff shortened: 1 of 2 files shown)


ososeitua (approved):
works well for me!

ValentinBoudevinSFL (approved):

PR #437, opened 2026-07-07
fix(cli): accept both bare and -- prefix for all commands

reh-sfl:

In the vulnscout host wrapper, lifecycle and configuration commands (start, stop, restart, update, list-projects, list-scans, config, config-list, config-clear, delete-scan) previously only accepted one form or the other, inconsistently with --serve, --add-spdx and other flags that already used --.

- Accept both 'cmd' and '--cmd' forms for: start, stop, restart, update, list-projects, list-scans, config, config-list, config-clear, delete-scan
- Update help text and example to use the -- prefix for list-* and config commands, and show 'cmd|--cmd' for lifecycle commands
- Update the 'stop' note to mention both forms
- Fix help example: 'vulnscout config' -> 'vulnscout --config'

In the container entrypoint, add the missing --clear-inputs entry to the help text (the flag was already parsed but undocumented).

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

## Pull Request Checklist

### DIFF
--- a/src/entrypoint.sh
+++ b/src/entrypoint.sh
@@ -615,11 +615,6 @@ cmd_daemon() {
     tail -f /dev/null
 }
 
-cmd_clear_inputs() {
-    rm -f "$INPUTS_DIR"/*/*
-    echo "Cleared all inputs"
-}
-
 #######################################
 # Print status update to file + console
 # Globals:
@@ -692,8 +687,6 @@ while [[ $# -gt 0 ]]; do
             OSV_SCAN_REQUESTED=true; SCAN_REQUIRED=true; shift ;;
         --perform-sbom-cve-check-scan)
             SBOM_CVE_CHECK_SCAN_REQUESTED=true; SCAN_REQUIRED=true; shift ;;
-        --clear-inputs)
-            cmd_clear_inputs; shift ;;
         --delete-scan)
             cmd_delete_scan "$2"; shift 2 ;;
         --serve)
--- a/vulnscout
+++ b/vulnscout
@@ -43,15 +43,15 @@ VulnScout - Vulnerability Scanner
 Usage: vulnscout [--project <name>] [--variant <name>] <command> [options]
 
 Container lifecycle:
-  start                           Start VulnScout container (auto-started by other commands)
-  stop                            Stop and remove VulnScout container
-  restart                         Restart VulnScout container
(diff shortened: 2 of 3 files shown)


vulnscout
WilsonZiweiWang, line 1: Do we need the `clear-inputs` in `parse_and_run()` ?

vulnscout
```diff
@@ -83,14 +83,14 @@ Scan & output commands:
   --delete-scan <id>                Delete a past scan by its ID
 
 Data retrieval commands:
-  list-projects                   List all projects and their variants
-  list-scans                      List all past scans
+  --list-projects                 List all projects and their variants
```
ValentinBoudevin, line 86: Agree with Ziwei feedback. In the end do we need  --list-projects and  --list-scans anymore?

For me it is the kind of stuff you want to get with an API call

vulnscout
reh-sfl: I added clear-inputs option in the host wrapper

src/entrypoint.sh
```diff
@@ -65,6 +65,7 @@ Scan & output commands:
   --import-custom-assessments <path>  Import custom assessments from .json, .tar.gz, or directory
   --match-condition <expr>  Exit code 2 if condition met (e.g. "cvss >= 9.0")
   --delete-scan <id>        Delete a past scan by its ID
+  --clear-inputs            Clear all staged input files
```
ValentinBoudevinSFL: Should be removed

vulnscout
```diff
@@ -81,16 +81,17 @@ Scan & output commands:
   --compress                      Compress export output into a .tar.gz archive
   --import-custom-assessments <path>  Import custom assessments from .json, .tar.gz, or directory
   --delete-scan <id>                Delete a past scan by its ID
+  --clear-inputs                  Clear all staged input files
```
ValentinBoudevinSFL: Should be removed

vulnscout
ValentinBoudevinSFL: No need anymore

ValentinBoudevinSFL (changes requested):

sfl-zwang (approved):

vulnscout
reh-sfl: Addressed

ValentinBoudevinSFL (approved):

PR #455, opened 2026-07-15
Templates: sanitize HTML in CVE descriptions

reh-sfl:

* Templates: sanitize HTML in CVE descriptions

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Find a CVE within your project scope that have HTML in it and that outputs malformed AsciiDoc report. Run an export. You should see all these formatting issues do not exist anymore.

## Pull Request Checklist

### DIFF
--- a/src/views/templates.py
+++ b/src/views/templates.py
@@ -1,6 +1,8 @@
 # Copyright (C) 2026 Savoir-faire Linux, Inc.
 # SPDX-License-Identifier: GPL-3.0-only
 
+from html.parser import HTMLParser
+from html import unescape as html_unescape
 from jinja2 import sandbox, FileSystemLoader, ChoiceLoader
 import subprocess
 import os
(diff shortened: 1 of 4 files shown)


src/views/templates.py
```diff
@@ -347,9 +381,18 @@ def escape_adoc(value: Optional[str]) -> str:
...
+        HTML tags and entities are stripped first using Python's stdlib HTML
+        tokenizer so that Asciidoctor never sees raw HTML syntax.
         """
         if not value:
             return ""
+        # Strip any embedded HTML tags and decode entities (e.g. &lt; -> <,
+        # &amp; -> &) so Asciidoctor receives plain text instead of raw HTML
+        # that would malform the rendered output.
+        stripper = _HtmlStripper()
+        stripper.feed(value)
```
ValentinBoudevinSFL: Decoded entities can reintroduce raw HTML after the single parser pass. For example, `&lt;b&gt;bold&lt;/b&gt;` is emitted as `<b>bold</b>` because `HTMLParser` decodes charrefs inside `handle_data` but does not tokenize the decoded text as tags. That leaves Asciidoctor seeing raw HTML syntax again. Decode entities before stripping tags, or run the stripped/decoded output through the tag stripper again until no encoded markup can become tags.

ValentinBoudevinSFL (changes requested):
The change is not ready because the sanitizer still emits raw HTML when tags are entity-encoded, which is a common form for CVE descriptions and directly undermines the intended fix.

src/views/templates.py
reh-sfl: Addressed.

src/views/templates.py
```diff
@@ -327,6 +361,33 @@ def get_env_var(key: str, default: str = "") -> str:
...
+        the literal string ``<b>bold</b>`` and survive as raw HTML. Re-run
+        the stripper until it no longer changes the text (or the pass limit
+        is reached) so that no depth of encoded markup can slip through.
+        """
+        for _ in range(TemplatesExtensions._MAX_HTML_STRIP_PASSES):
+            stripper = _HtmlStripper()
+            stripper.feed(value)
+            stripper.close()
+            stripped = stripper.get_text()
+            if stripped == value:
```
ValentinBoudevinSFL: The bounded loop can still return partially decoded markup while the text is changing. For example, six layers of `&amp;...lt;b&amp;...gt;bold...` leave an encoded `<b>` after the fifth pass, so the sanitizer has not reached the fixed point it relies on. Decode entities before tokenizing, or continue until stable while rejecting/truncating inputs that exceed the maximum instead of returning the unsafe intermediate value.

ValentinBoudevinSFL (changes requested):
The previous issue is only partially addressed: normal and double-encoded cases are covered, but deeply encoded markup can still escape the sanitizer due to the hard pass limit.

src/views/templates.py
ValentinBoudevinSFL: Thanks for addressing the single-pass issue. The new implementation handles common cases, but the current pass cap can still return before reaching a fixed point for deeper entity nesting, so I left a follow-up comment on that edge case.

ValentinBoudevinSFL (dismissed):
The revised decode-before-strip ordering addresses the previously reported sanitization issues, and the added tests cover encoded markup and decode-budget exhaustion. Ready to merge.

CHANGELOG.md
sfl-zwang, line 1: I think it is too early to update the CHANGELOG.

reh-sfl: Addressed

src/views/templates.py
```diff
@@ -327,6 +379,49 @@ def get_env_var(key: str, default: str = "") -> str:
...
+            if next_decoded == decoded:
+                break
+            decoded = next_decoded
+
+        stripper = _HtmlStripper()
+        stripper.feed(decoded)
+        stripper.close()
+        return stripper.get_text()
+
     @staticmethod
```
ValentinBoudevinSFL, line 425: `HTMLParser.feed()` can raise `AssertionError` for malformed marked sections such as `<![foo]>`, allowing an untrusted CVE description to abort report rendering. Handle malformed declarations safely or use a tolerant sanitizer, and add a regression test for this input.

ValentinBoudevinSFL (changes requested):
The decoding order fixes the previously reported entity issues, but malformed HTML can still crash template rendering.

sfl-zwang (approved):

ValentinBoudevin (approved):

src/views/templates.py
```diff
@@ -327,6 +379,49 @@ def get_env_var(key: str, default: str = "") -> str:
...
     def escape_adoc(value: Optional[str]) -> str:
         """Neutralise AsciiDoc structural markup in arbitrary free-form text.
```
ValentinBoudevinSFL, line 427: Removing tags can concatenate surrounding text into a new raw tag. For example, `&lt;<span>b</span>&gt;bold&lt;/<span>b</span>&gt;` becomes `<b>bold</b>` after decoding and this single stripping pass. Re-sanitize until stable with safe handling when a depth limit is exceeded, or use a sanitizer that guarantees its output cannot form markup; add a regression test for split tag names.

ValentinBoudevinSFL (dismissed):
The decode-before-strip ordering fixes the previously identified entity issue, but a single stripping pass remains bypassable because removed elements can join text fragments into fresh HTML markup.

PR #460, opened 2026-07-17
Transfer: multiple source-target modes

reh-sfl:

1. Adding a reflexive option to copy within the same variant, we will have:
- Copy to a different variant
- Copy within the same variant

2. Fix the bug where it showed the source variant in the target list

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Open the web dashboard, go to the Transfer tab then use the new "I want to copy…" selector

## Pull Request Checklist

### DIFF
--- a/src/routes/settings.py
+++ b/src/routes/settings.py
@@ -491,9 +491,6 @@ def init_app(app: Flask) -> None:
             return None, err
         target_uuid = cast(uuid.UUID, target_uuid)
 
-        if source_uuid == target_uuid:
-            return None, (jsonify({"error": "Source and target variants must be different."}), 400)
-
         source = VariantController.get(source_id)
         target = VariantController.get(target_id)
         if source is None or target is None:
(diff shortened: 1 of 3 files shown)


ValentinBoudevin (dismissed):
1) "Copy assessments to other matching findings in one variant.¨ seems a bit difficult to understand as we don´t use findings anywhere else. 

Should be something closer to "Copy assessments from previous packages versions in a variant". 

2) 

Image

Detected 0 assessments but this is wrong. 

Example linux-yocto@6.6.129 (the current version) doesn´t have any custom assessments yet: 

Image

src/routes/settings.py
```diff
@@ -893,6 +894,7 @@ def copy_variant_assessments() -> ResponseReturnValue:
                     src_finding = assessment.finding
```
ValentinBoudevinSFL, line 894: Within one variant, a selected target can also be the source assessment of another selection. In overwrite mode, processing A→B before B→A mutates B before its values are copied, making results order-dependent and potentially collapsing both assessments to A's values. Snapshot all source assessment fields before applying any writes, or reject cyclic/conflicting selections; add a reciprocal-copy test.

ValentinBoudevinSFL (changes requested):
The same-variant flow is generally well integrated, but overlapping source/target selections can corrupt copy results depending on operation order and must be addressed before merge.

ValentinBoudevinSFL (approved):

sfl-zwang (approved):

PR #461, opened 2026-07-17
SBOM tab: add match condition filtering

reh-sfl:

SBOM tab: add match condition filtering

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

1. Open the web dashboard, go to SBOM tab.
2. Enter `cvss >= 9` in Match condition then click Apply.
6. Click Show Vulnerabilities for one package.
7. Confirm only vulnerabilities for that package satisfying `cvss >= 9` are displayed.

The PR supports the following conditions, the same as those supported in the reports and CI mode:

```text
cvss_min >= 7
epss >= 50%
effort < 3600
effort_min <= 1800
effort_max >= 7200
fixed
ignored
affected
pending
new
not fixed
cvss >= 7 and pending
cvss >= 9 or epss >= 50%
id == CVE-2014-6271
```

A zero-result condition is valid. For example:

```text
cvss > 11
```

This also supports nested Condition

```text
((cvss >= 9.0 or (cvss >= 7.0 and epss >= 30%)) and (pending or affected))
```

When you input invalid Conditions you will get an error:

```text
unknown == 2
true == true == true
true == true and
cvss > CVE-TEST-0001
```

Finally, click Reset Filters and confirm the condition and filtered results are cleared.

## Pull Request Checklist


src/routes/vulnerabilities.py
```diff
@@ -296,6 +299,40 @@ def init_app(app: Flask) -> None:
...
+            return {"error": f"items must contain at most {MATCH_CONDITION_MAX_ITEMS} entries"}, 400
+
+        parser = ConditionParser()
+        matching_ids: list[str] = []
+        try:
+            for item in items:
+                if (
+                    not isinstance(item, dict)
+                    or not isinstance(item.get("id"), str)
+                    or not isinstance(item.get("data"), dict)
```
ValentinBoudevinSFL, line 326: The condition is parsed again for every item because `parser.evaluate(condition, ...)` receives the source string inside the loop. At the permitted 50,000 items and 1,000-character conditions, one request can trigger tens of millions of parsing operations and become a CPU DoS vector. Parse and validate the expression once, then evaluate the parsed representation for each item's data.

src/controllers/conditions_parser.py
```diff
@@ -97,19 +111,17 @@ def _eval_internal(self, condition: list) -> bool | int | float | str:
...
                 # TypeError at runtime, which marks the condition as invalid.
-                return self._eval_internal(condition[0]) < self._eval_internal(condition[2])  # type: ignore[operator]
-            if condition[1] == ">":
-                return self._eval_internal(condition[0]) > self._eval_internal(condition[2])  # type: ignore[operator]
-            if condition[1] == "<=":
-                return self._eval_internal(condition[0]) <= self._eval_internal(condition[2])  # type: ignore[operator]
-            if condition[1] == ">=":
-                return self._eval_internal(condition[0]) >= self._eval_internal(condition[2])  # type: ignore[operator]
+                return self._compare_ordered(
+                    condition[1],
```
ValentinBoudevinSFL, line 121: Ordering comparisons pass `unknown_as_literal=True`, so an unknown RHS is accepted whenever the left value is also orderable as a string (for example, `id > TYPO` performs a lexical comparison). This contradicts the documented rule that literals are only allowed with `==`/`!=` and can hide typos. Evaluate ordered RHS expressions without `unknown_as_literal`.

frontend/src/pages/TablePackages.tsx
```diff
@@ -191,12 +199,41 @@ function TablePackages({ packages, onShowVulns }: Readonly<Props>) {
...
+        if (!condition) {
+            setMatchingVulnerabilityIds(null);
+            setMatchConditionError('');
+            return;
+        }
+        setIsMatchingCondition(true);
+        setMatchConditionError('');
+        try {
+            setMatchingVulnerabilityIds(await Vulnerabilities.matchCondition(condition, vulnerabilities));
+        } catch (error) {
```
ValentinBoudevinSFL, line 244: An in-flight evaluation can restore stale results after `vulnerabilities` changes: the effect clears the IDs, but the old request later resolves and calls `setMatchingVulnerabilityIds`. Multiple requests can also complete out of order because Enter remains active while the Apply button is disabled. Track a request generation or use an AbortController, and ignore results when the vulnerability snapshot or latest request has changed.

ValentinBoudevinSFL (changes requested):
The feature is well covered, but it currently introduces a high-cost server evaluation path, accepts undocumented ordering literals, and can display stale asynchronous results. These issues should be addressed before merge.

ValentinBoudevinSFL (changes requested):
Image

Great design, but cvss condition doesn't seem to work.

frontend/src/pages/Explorer.tsx
```diff
@@ -228,7 +229,8 @@ function Explorer({ darkMode, setDarkMode }: Readonly<Props>) {
         setTab('vulnerabilities');
     }
```
ValentinBoudevinSFL, line 232: The matching-ID snapshot remains stored after drilling down, while `TablePackages` is unmounted. A subsequent vulnerability refresh therefore leaves `TableVulnerabilities` filtering fresh data with stale IDs. Clear `filterVulnerabilityIds` when the vulnerability snapshot or scope changes.

ValentinBoudevinSFL (changes requested):
The filtering flow is mostly coherent, but drill-down filters can become stale after vulnerability data refreshes.

ValentinBoudevinSFL (changes requested):
Image

The feature looks good to me (cvss works now). 

But the frontend need to receive some clarity:

1- Search Syntax is only applicable to the search bar, should stay next to it
2- Add another information mark next to match condition to provide some explanations
3- We need to find a way to make obvious the fact the Apply button is only for Match Condition and not the search bar

ValentinBoudevinSFL (changes requested):
Image

Frontend looks good
Match condition field looks good

But, it doesn't respect dod with "Dod: We have an efficiant way to get all remaining assessments to do based on a generic condition"
After this match condition entered I should get a new column, or something else, to get the quantity of Vulnerabilities per package which match this condition. 
Right now, it only get rid of the packages which doesn't have a vulnerability which respect it

reh-sfl:
@ValentinBoudevinSFL I added an small design change to the existing Vulnerabilities to indicate a match condition count, details shown when you hover.

image

ValentinBoudevin (changes requested):
LGTM can you just remove the column "remaining pending vulnerabilities"

ValentinBoudevin (approved):

PR #467, opened 2026-07-20
Misc UI refactor

reh-sfl:

* Export: Add confirmation / summary popup
* Settings: move icon to reflect general settings
* Review: toggle to filter out fully outdated assessments
* Move version below logo
* Remove counts in Review sub-sections

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

- Observe change in the navigation bar.
- Go to Export, pick an item then observe the confirmation popup
- Go to Review, use the outdated toggle
- Observe that version is now below logo

## Pull Request Checklist

- [x] T he code compiles and passes all tests

### DIFF
--- a/frontend/src/handlers/assessments.ts
+++ b/frontend/src/handlers/assessments.ts
@@ -42,6 +42,7 @@ type Assessment = {
 export type { Assessment };
 
 type ReviewTimeEstimate = {
+    id: string;
     vuln_id: string;
     variant_id?: string;
     optimistic: number;
@@ -54,6 +55,7 @@ type ReviewTimeEstimate = {
 };
 
 type ReviewCustomCvss = {
+    id: string;
     vuln_id: string;
     variant_id?: string;
     version: string;
--- a/frontend/src/pages/Explorer.tsx
+++ b/frontend/src/pages/Explorer.tsx
@@ -1,7 +1,6 @@
 import { useState, useEffect, useCallback, useRef } from "react";
 import NavigationBar from "../components/NavigationBar";
 import MessageBanner from "../components/MessageBanner";
-import VersionDisplay from "../components/VersionDisplay";
 import type { Package } from "../handlers/packages";
 import type { CVSS, Vulnerability } from "../handlers/vulnerabilities";
 import type { Assessment } from "../handlers/assessments";
@@ -456,9 +455,6 @@ function Explorer({ darkMode, setDarkMode }: Readonly<Props>) {
                 {tab === 'ai' && <AIContext />}
             </div>
             </main>
-            <footer>
-                <VersionDisplay />
-            </footer>
         </div>
     )
 }
--- a/frontend/src/pages/Explorer.tsx
+++ b/frontend/src/pages/Explorer.tsx
@@ -434,7 +434,7 @@ function Explorer({ darkMode, setDarkMode }: Readonly<Props>) {
                 />}
                 {tab === 'scans' && <ScanHistory variantId={currentVariantId} projectId={currentVariantId ? undefined : currentProjectId} onScanComplete={handleScanComplete} />}
                 {tab === 'review' && <Review variantId={currentVariantId} projectId={currentVariantId ? undefined : currentProjectId} onAssessmentChanged={handleAssessmentChanged} />}
-                {tab === 'exports' && <Exports variantId={currentVariantId} projectId={currentProjectId} />}
+                {tab === 'exports' && <Exports variantId={currentVariantId} projectId={currentProjectId} variantIds={currentVariantIds} />}
                 {tab === 'transfer' && <Transfer projectId={currentProjectId} onDataChanged={(message) => {
                     if (message) setLoadingMessage(message);
                     loadData(currentVariantId, currentVariantId ? undefined : currentProjectId, undefined, undefined, currentVariantIds, currentMultiOperation);
--- a/frontend/src/pages/Exports.tsx
+++ b/frontend/src/pages/Exports.tsx
@@ -6,6 +6,11 @@ import {
 import FileTag from "../components/FileTag";
 import PopupExportOptions from "../components/PopupExportOptions";
 import type { Options as PopupOptions } from "../components/PopupExportOptions";
+import ConfirmationModal from "../components/ConfirmationModal";
+import Projects from "../handlers/project";
+import type { Project } from "../handlers/project";
+import Variants from "../handlers/variant";
+import type { Variant } from "../handlers/variant";
 
 
 type ExportDoc = {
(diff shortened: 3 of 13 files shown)


sfl-pe-sync:
test

frontend/src/pages/Exports.tsx
```diff
@@ -152,6 +195,7 @@ function Exports ({ variantId, projectId }: Readonly<{ variantId?: string; proje
         projectId={projectId}
         opened={openDl === doc.id}
         onOpen={() => openDl === doc.id ? setOpenDl(null) : setOpenDl(doc.id)}
+        onRequestDownload={(exportType, url) => setPendingDownload({ exportType, url })}
```
sfl-pe-sync, line 240: The confirmation scope can disagree with the actual download URL. FileTag exports one `variant_id` when present or the entire `project_id` otherwise, while this modal lists `variantIds`; project-level subset selection therefore claims only selected variants but exports every project variant. Derive the confirmation from the effective URL scope, or change the export request to carry the selected variant IDs.

frontend/src/handlers/assessments.ts
```diff
@@ -159,6 +182,18 @@ class Assessments {
         return removeDuplicateAssessments(assessments);
     }
 
+    static async listByVulnerability(vulnId: string, signal?: AbortSignal): Promise<Assessment[]> {
+        const response = await fetch(
+            import.meta.env.VITE_API_URL + `/api/vulnerabilities/${encodeURIComponent(vulnId)}/assessments`,
+            { mode: 'cors', signal },
+        );
+        if (!response.ok) return [];
```
sfl-pe-sync: Returning an empty list for any non-2xx response makes a server or authorization failure indistinguishable from having no assessments. VulnModal then replaces existing summary assessments with this empty result. Throw on non-OK responses so failure preserves current data and can be surfaced to the user.

frontend/src/pages/TableVulnerabilities.tsx
```diff
@@ -666,6 +667,21 @@ function TableVulnerabilities ({ vulnerabilities, filterLabel, filterValue, appe
         setModalVulnSnapshot(prev => prev.map(v => v.id === vulnId ? replace_vuln : v));
     }, [patchVuln]);
 
+    const loadPageDescriptions = useCallback((pageVulns: Vulnerability[]) => {
+        const missingVulns = pageVulns.filter(vuln => !loadedDescriptionIds.current.has(vuln.id));
+        if (missingVulns.length === 0) return;
+        missingVulns.forEach(vuln => loadedDescriptionIds.current.add(vuln.id));
+
+        Vulnerabilities.fetchDescriptions(missingVulns.map(vuln => vuln.id))
+            .then(descriptions => {
```
sfl-pe-sync: This asynchronous callback replaces the vulnerability using the stale `vuln` captured before the request. Any assessment, status, or score update made while descriptions load can be overwritten. Merge only the returned `texts` into the latest record, using a functional updater or a partial-patch API.

src/routes/vulnerabilities.py
```diff
@@ -741,6 +738,29 @@ def _h(v: int | None) -> Iso8601Duration | None:
             case _ as fmt:
                 raise ValueError("Unknown format", fmt)
 
+    @app.post('/api/vulnerabilities/descriptions')
+    def get_vuln_descriptions() -> ResponseReturnValue:
+        """Return primary descriptions for the current vulnerability table page."""
+        payload = request.get_json(silent=True)
+        vuln_ids = payload.get("ids") if isinstance(payload, dict) else None
+        if not isinstance(vuln_ids, list) or not all(isinstance(vuln_id, str) for vuln_id in vuln_ids):
```
sfl-pe-sync: The endpoint accepts an unbounded number and length of IDs, allowing callers to create very large SQL `IN` clauses and responses. Enforce a page-sized maximum, reject oversized IDs, and return 400/413 when limits are exceeded.

sfl-pe-sync (changes requested):
The refactor introduces misleading export-scope confirmation and two paths that can discard or overwrite vulnerability assessment data. These correctness and data-exposure risks should be fixed before merge.

frontend/src/pages/Review.tsx
```diff
@@ -140,6 +196,11 @@ function Review({ variantId, projectId, onAssessmentChanged }: Readonly<Props>)
     const [editVariantPackageMap, setEditVariantPackageMap] = useState<Record<string, string[]>>({});
```
sfl-pe-sync, line 187: Selections persist when `variantId` or `projectId` changes. After selecting records in one scope, navigating to another leaves the hidden IDs selected and allows deleting records from the previous scope. Clear or prune all row-selection states whenever the review scope or backing lists change.

frontend/src/pages/Review.tsx
```diff
@@ -114,6 +156,19 @@ function formatDate(iso: string): string {
     });
```
sfl-pe-sync, line 156: This unions superseded package references across grouped assessments, so a package that is outdated in one variant but current in another makes the whole grouped row appear fully outdated. Evaluate each underlying assessment against its own `superseded_map` and require every underlying assessment to be fully superseded.

frontend/src/pages/Review.tsx
```diff
@@ -635,6 +700,98 @@ function Review({ variantId, projectId, onAssessmentChanged }: Readonly<Props>)
         }
     }, [refreshAssessmentLists, showMessage]);
 
+    const selectedRows = activeTab === 'assessments'
+        ? selectedAssessments
+        : activeTab === 'ai-assessments'
+            ? selectedAiAssessments
+            : activeTab === 'time-estimates'
```
sfl-pe-sync, line 718: Deleting grouped assessments through parallel individual requests is not atomic: if one request fails, earlier deletions remain committed while the UI reports total failure and retains stale state. Add a transactional batch-delete endpoint (and use it for AI groups as well) so the selected operation either fully succeeds or fails.

src/routes/assessments.py
```diff
@@ -422,6 +434,7 @@ def _hours_to_iso(h: int) -> str:
             lik = te.likely or 0
```
sfl-pe-sync, line 626: The response is collapsed by `(vuln_id, scoped_variant)` but exposes only one underlying `TimeEstimate.id`. If multiple findings produce the same logical row, deleting that ID leaves the other records and the row reappears. Return all represented IDs and delete them together, or stop collapsing distinct records.

sfl-pe-sync (changes requested):
The refactor introduces destructive-operation correctness issues: selections can cross review scopes, grouped deletions are non-atomic, and grouped outdated/time-estimate records can be classified or deleted incorrectly.

sfl-zwang (dismissed):

frontend/src/pages/Explorer.tsx
```diff
@@ -349,7 +348,6 @@ function Explorer({ darkMode, setDarkMode }: Readonly<Props>) {
             </div>
             </main>
             <footer>
```
sfl-zwang: No need to keep this `<footer>` element anymore since there is nothing in it.

sfl-zwang (changes requested):

src/routes/assessments.py
```diff
@@ -605,6 +619,38 @@ def review_custom_cvss() -> ResponseReturnValue:
```
sfl-pe-sync, line 710: Both new batch-delete endpoints accept only record UUIDs and never verify that the records belong to the project/variant currently being reviewed. A crafted or stale request can therefore delete time estimates or custom CVSS records from another scope. Require the requested project/variant scope and reject every record outside it before committing; add a cross-project isolation regression test.

sfl-pe-sync (changes requested):
The UI refactor introduces useful shared modal and bulk-selection behavior, but the new backend deletion APIs violate VulnScout's project/variant isolation invariant. The PR should not merge until deletion scope is enforced server-side.

frontend/src/pages/Explorer.tsx
sfl-pe-sync: Agreed. The footer is now empty and should be removed rather than retained as unused markup.

sfl-zwang (dismissed):

ValentinBoudevinSFL (changes requested):
1- Outdated filter don't show entries with a mix of outdated and current assessments

Image

2- Can we remove the count (here 204) in the different tab. As it doesn't make sense anymore with filters

Image

ValentinBoudevinSFL (approved):

PR #470, opened 2026-07-22
Optimize Explorer Initial Load Performance

ValentinBoudevin:

- Defers post-scan background tasks to prevent contention during initial Explorer loading.
- Optimizes package enrichment by splitting expensive database lookups.
- Introduces compact vulnerability responses with details loaded on demand.
- Reduces assessment payload size through compact serialization and compression.
- Preserves full vulnerability and assessment details when opening the vulnerability modal.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

- Load a huge `vulnscout.db` database with and without this pull request. Time to load the UI should be significantly reduced.
- Open or hover over a vulnerability and verify its description, links, and CVSS details appear.
- Verify assessment details remain available in the vulnerability modal.

### DIFF
--- a/frontend/src/handlers/vulnerabilities.ts
+++ b/frontend/src/handlers/vulnerabilities.ts
@@ -74,6 +74,8 @@ type Vulnerability = {
     assessments: Assessment[];
     /** False only for compact list records whose modal-only data is deferred. */
     details_loaded?: boolean;
+    /** Atomic terms confirmed in deferred descriptions by server-side search. */
+    description_search_terms?: string;
 };
 
 export type { Vulnerability, CVSS };
@@ -373,6 +375,43 @@ class Vulnerabilities {
         }
         const parsed = asVulnerability(await response.json());
         return Array.isArray(parsed) ? null : parsed;
+    }
+
+    static async searchDescriptionTerms(
+        vulnerabilityIds: string[],
+        terms: string[],
+        variantId?: string,
+        projectId?: string,
+        signal?: AbortSignal,
+    ): Promise<Record<string, string[]>> {
+        const url = new URL(
+            import.meta.env.VITE_API_URL + '/api/vulnerabilities/search-descriptions',
+            window.location.href,
+        );
+        if (variantId) {
+            url.searchParams.set('variant_id', variantId);
+        } else if (projectId) {
+            url.searchParams.set('project_id', projectId);
+        }
+        const response = await fetch(url.toString(), {
+            method: 'POST',
+            mode: 'cors',
+            headers: { 'Content-Type': 'application/json' },
+            body: JSON.stringify({ vulnerability_ids: vulnerabilityIds, terms }),
+            signal,
+        });
+        if (!response.ok) {
+            throw new Error(`Failed to search vulnerability descriptions (${response.status})`);
+        }
+        const data: unknown = await response.json();
+        if (!data || typeof data !== 'object' || !('matches' in data)) return {};
+        const matches = (data as { matches?: unknown }).matches;
+        if (!matches || typeof matches !== 'object' || Array.isArray(matches)) return {};
+        return Object.fromEntries(
+            Object.entries(matches).flatMap(([term, ids]) =>
+                Array.isArray(ids) ? [[term, asStringArray(ids)]] : []
+            )
+        );
     }
 
     // Re-fetch a single vulnerability in the given (project) scope and return
--- a/frontend/src/handlers/vulnerabilities.ts
+++ b/frontend/src/handlers/vulnerabilities.ts
@@ -72,6 +72,8 @@ type Vulnerability = {
     simplified_status: string;
     status_summary?: StatusSummary;
     assessments: Assessment[];
+    /** False only for compact list records whose modal-only data is deferred. */
+    details_loaded?: boolean;
 };
 
 export type { Vulnerability, CVSS };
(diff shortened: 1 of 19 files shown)


frontend/src/handlers/vulnerabilities.ts
```diff
@@ -306,6 +310,27 @@ class Vulnerabilities {
...
+        projectId?: string,
+        signal?: AbortSignal,
+    ): Promise<Vulnerability | null> {
+        const url = new URL(
+            import.meta.env.VITE_API_URL + `/api/vulnerabilities/${encodeURIComponent(vulnId)}`,
+            window.location.href,
+        );
+        if (variantId) {
+            url.searchParams.set('variant_id', variantId);
+        } else if (projectId) {
```
sfl-pe-sync, line 369: Returning `null` for non-success responses prevents the modal's rejection handler from setting `detailsError`; users instead see silently incomplete vulnerability details. Throw an error containing the response status here so modal loads surface the error and hover loads remain uncached and retryable.

src/routes/packages.py
```diff
@@ -171,47 +171,58 @@ def _multi_pkg_ids_for_variant(variant_uuid: uuid.UUID) -> set[uuid.UUID]:
...
-            rows = db.session.execute(enrich_query).all()
+                document_query = document_query.where(Variant.project_id == uuid.UUID(project_id))
+
+            document_rows = db.session.execute(document_query).all()
+            document_meta = {
+                row.document_id: row
+                for row in document_rows
+            }
+            document_ids = list(document_meta)
+            association_rows = db.session.execute(
```
sfl-pe-sync, line 212: This query loads every package association for every document in scope and only filters to `pkg_ids` afterward in Python. Large documents can therefore create substantial unnecessary database transfer and memory use. Add `SBOMPackage.package_id.in_(pkg_ids)` to the SQL predicate.

sfl-pe-sync (changes requested):
The optimization is directionally sound, but failed detail requests are silently treated as valid empty data, and package enrichment still retrieves potentially large amounts of irrelevant association data.

frontend/src/handlers/vulnerabilities.ts
ValentinBoudevin: Fixed

src/routes/packages.py
ValentinBoudevin: Adding the large package-ID IN predicate was previously benchmarked and regressed project requests from approximately 0.44 s to 5.16 s.

frontend/src/pages/TableVulnerabilities.tsx
```diff
@@ -354,6 +354,69 @@ function TableVulnerabilities ({ vulnerabilities, filterLabel, filterValue, appe
...
+    const [modalDetailsError, setModalDetailsError] = useState(false);
+    const hoverDetailsCache = useRef(new Map<string, Vulnerability>());
+
+    const loadHoverDetails = useCallback(async (summary: Vulnerability): Promise<Vulnerability> => {
+        if (summary.details_loaded !== false) return summary;
+        const cacheKey = `${summary.id}:${variantId ?? ''}:${projectId ?? ''}`;
+        const cached = hoverDetailsCache.current.get(cacheKey);
+        if (cached) return cached;
+        const details = await Vulnerabilities.getDetails(summary.id, variantId, projectId);
+        const resolved = details ? {
```
sfl-pe-sync, line 380: Lazy detail requests preserve only `variantId` or the entire `projectId`, not the compare/multi-variant scope used to build this table. A compare-only vulnerability can therefore load details from the wrong variant, while multi-variant views can include details from unselected project variants. Extend `getDetails` and both callers to pass the same compare/multi-variant selection used by `Vulnerabilities.list`.

src/routes/vulnerabilities.py
```diff
@@ -52,6 +52,19 @@
 _GHSA_RE = re.compile(r'^GHSA-[A-Z0-9]{4}-[A-Z0-9]{4}-[A-Z0-9]{4}$')
 
 
+def _attack_vector_from_vector(vector: str) -> str | None:
```
sfl-pe-sync, line 58: Substring matching can misclassify environmental vectors: for example, a vector containing base `AV:L` and `MAV:N` returns `NETWORK` because `AV:N` occurs inside `MAV:N`. Parse slash-delimited components and match an exact component beginning with `AV:`.

sfl-pe-sync (dismissed):
The performance changes introduce incorrect vulnerability details for compare/multi-variant views and can misclassify attack vectors used by Explorer filters.

src/routes/vulnerabilities.py
```diff
@@ -638,6 +669,12 @@ def _h(v: int | None) -> Iso8601Duration | None:
         _populate_found_by(records, _scope_variant, _scope_project,
                            active_scan_ids=current_scan_ids or None)
         vulns = {r.id: r.to_dict() for r in records}
+        # The frontend consumes the richer ``texts`` field and never reads the
+        # legacy top-level description.  Omitting this duplicate from the list
+        # response saves substantial transfer and JSON parsing on large scans;
+        # the single-vulnerability endpoint still returns the complete record.
+        for vuln in vulns.values():
+            vuln.pop("description", None)
```
reh-sfl, line 714: You need to check if compact mode is requested before you pop the heavy texts like I did in my original PR. This way anyone who needs that data can obtain it.

reh-sfl (changes requested):
Also the search will be broken because we can no longer search inside the descriptions, is this something we want to keep?

src/routes/vulnerabilities.py
ValentinBoudevinSFL: Solved, now search is server side with Enter press to validate the search (same as match condition)

reh-sfl (approved):

PR #471, opened 2026-07-23
Web dashboard: persist project variant selection

reh-sfl:

Creates a base64 encoded config line for the current project / variant(s) so when you restart you are brought to the same context as before.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Open the web dashboard, pick a project / variant through selector and restart, you should see the exact same project you left before.

## Pull Request Checklist

### DIFF
--- a/frontend/src/pages/Explorer.tsx
+++ b/frontend/src/pages/Explorer.tsx
@@ -20,6 +20,9 @@ import AIContext from './AIContext';
 import Assessments, { removeDuplicateAssessments, STATUS_VEX_TO_GRAPH } from '../handlers/assessments';
 import Config from "../handlers/config";
 import type { AppConfig } from "../handlers/config";
+import type { FrontendScope } from "../handlers/config";
+import Projects from '../handlers/project';
+import Variants from '../handlers/variant';
 
 const tabLabels: Record<string, string> = {
         metrics: 'Metrics',
@@ -62,6 +65,7 @@ function Explorer({ darkMode, setDarkMode }: Readonly<Props>) {
         contact_email: "",
         grype_memlimit: "",
     });
+    const [frontendScope, setFrontendScope] = useState<FrontendScope | null>(null);
     const [currentVariantId, setCurrentVariantId] = useState<string | undefined>(undefined);
     const [currentProjectId, setCurrentProjectId] = useState<string | undefined>(undefined);
     const [currentBaseVariantId, setCurrentBaseVariantId] = useState<string | undefined>(undefined);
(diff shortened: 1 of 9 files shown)


ValentinBoudevinSFL (dismissed):

sfl-zwang (changes requested):
This is too complicated for its purpose. Use browser's `localstorage` to store the project and variant states.

ValentinBoudevin (dismissed):

frontend/src/pages/Explorer.tsx
```diff
@@ -122,16 +126,63 @@ function Explorer({ darkMode, setDarkMode }: Readonly<Props>) {
...
+                    ? scope?.compare_base_id
+                    : scope?.mode === 'select' && scope.variant_ids.length === 1
+                        ? scope.variant_ids[0]
+                        : config.variant?.id || undefined;
+                const projectId = scope?.project_id || config.project?.id || undefined;
+                const compareVariantId = compareActive ? scope?.compare_variant_id : undefined;
+                setCurrentVariantId(compareVariantId || variantId);
+                setCurrentProjectId(projectId);
+                setCurrentBaseVariantId(compareActive ? variantId : undefined);
+                setCurrentOperation(compareActive ? scope?.compare_operation : undefined);
```
sfl-pe-sync, line 180: Persisted multi-variant scopes always restore with `union`, regardless of the `multiOperation` originally applied. `FrontendScope` does not store this value, so an intersection selection silently changes semantics after reload. Persist `multi_operation`, restore it here and in the selector, and add an Explorer reload regression test.

frontend/src/pages/Explorer.tsx
```diff
@@ -122,16 +126,63 @@ function Explorer({ darkMode, setDarkMode }: Readonly<Props>) {
...
+                setCurrentVariantIds(multiActive ? scope?.variant_ids : undefined);
+                setCurrentMultiOperation(multiActive ? 'union' : undefined);
+                loadData(
+                    multiActive ? undefined : variantId,
+                    (multiActive || !variantId) ? projectId : undefined,
+                    compareVariantId,
+                    compareActive ? scope?.compare_operation : undefined,
+                    multiActive ? scope?.variant_ids : undefined,
+                    multiActive ? 'union' : undefined,
+                );
```
sfl-pe-sync, line 190: The new scope-validation requests share the `Config.get()` catch handler. If `Projects.list()` or `Variants.list()` transiently fails, this calls `loadData(undefined)` and discards the successfully fetched configured scope, potentially loading unscoped data. Handle validation failures separately and retain the server-configured scope or show an error without issuing an unscoped request.

sfl-pe-sync (changes requested):
The persistence implementation is not ready to merge: it changes multi-variant intersection scopes to unions after reload and can fall back to an unscoped data request when saved-scope validation fails. The changed imports are repository-local and introduce no runtime dependency concern.

frontend/src/pages/Explorer.tsx
```diff
@@ -122,16 +126,73 @@ function Explorer({ darkMode, setDarkMode }: Readonly<Props>) {
...
     useEffect(() => {
+        let cancelled = false;
+
         Config.get()
-            .then(config => {
+            .then(async config => {
+                let scope = Config.getFrontendScope();
+                let discardedScope = false;
+                // Validate the saved scope against the live project/variant
+                // lists. A transient fetch failure here must not discard the
```
sfl-pe-sync, line 136: A fulfilled empty project list is treated as inconclusive whenever the server config has a default project. This preserves an unvalidated saved scope, skips variant validation, and then calls loadData with potentially deleted project/variant IDs. Rejection is already handled by the catch path, so a successful list response should be authoritative: clear the saved scope whenever its project is absent, and update the test that currently codifies this stale-scope behavior.

sfl-pe-sync (changes requested):
The persistence implementation is generally consistent with the existing frontend architecture, uses only repository-local imports, and adds focused restoration tests. However, stale scopes are retained after an authoritative empty project response, allowing unavailable project/variant identifiers to remain active; this should be corrected before merge.

ValentinBoudevin (approved):

Issue #474, opened 2026-07-24
[Need clarification] Import Custom Assessment, change v0.17->v0.18


kili236:
Hello everyone,
I noticed a different behavior between v0.17 and v0.18 when importing custom assessments.

**Environment**
- SPDX SBOM tar generated in July 2026
- CVE data file generated in July 2026, with CVE-XXX marked as "Pending Assessment"
- Custom assessment generated in May 2026, with CVE-XXX marked as "Not Affected"

**Steps**
1. Import the SBOM and CVE data: `./vulnscout --project MyProject --variant MyVariant --add-spdx SBOM.spdx.tar.zst --add-cve-check cve-check.json`
2. Import custom assessment: `./vulnscout --project MyProject --variant MyVariant --import-custom-assessments assessment.json`

**v0.17**
Even if the custom assessment is older than the information contained in cve-check.json, it takes precedence.
Status is "Not Affected".

**v0.18**
The assessment from cve-check.json takes precedence over the custom assessment
Status is "Pending Assessment".

**Discussion**
On the one hand, I understand using `last_updated` or `timestamp` to determine which assessment should take precedence.
On the other hand, this implies that users must keep the full assessment history to obtain meaningful results.

**Use Case: Continuous Integration**
Current workflow is based on storing custom assessments separately and periodically generating new CVE data.
Every weekend, CI Pipeline:
1. Build a new image and generate an new cve-check.json
2. Import the SBOM and CVE data
3. Apply my custom assessments
4. Use --match-condition pending to identify new CVEs and fail the build accordingly.

This workflow appears to work well with v0.17, but fails with v0.18.
Did I miss anything, or is there a recommended way to support this workflow with v0.18?

Thank you very much for your help.

ValentinBoudevinSFL:
Hello, 

You are absolutely right, we made a regression here.

I think both options are valid and correspond to different use-cases (we needed the current v0.18 behavior for a project). 

v0.19 coming at the beginning of the week already address some inconsistencies in the export feature to do a cleanup of the feature. 

<img width="625" height="601" alt="Image" src="https://github.com/user-attachments/assets/432804b7-543a-46ad-94c9-497eca9d7860" />

I will add an option to use/skip the use of `last_updated` or `timestamp` and keep v0.17 behavior before the new release.

ValentinBoudevinSFL:
Version v0.19 is out and the new CLI commands should solve your issue.

```
[...]

PR #477, opened 2026-07-24
Review: add timestamp policies to imports

reh-sfl:

* Allow user to pick between original or current timestamps when importing data in the Review tab

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Open the web dashboard, initiate an import then test the different timestamp options.
For the CLI, refer to the new documentation for testing the extra timestamp args.

### Related Issue

#474 

## Pull Request Checklist

### DIFF
--- a/frontend/src/pages/Review.tsx
+++ b/frontend/src/pages/Review.tsx
@@ -145,6 +145,7 @@ function Review({ variantId, projectId, onAssessmentChanged }: Readonly<Props>) 
     const [transferMode, setTransferMode] = useState<'import' | 'export' | null>(null);
     const [transferVariantIds, setTransferVariantIds] = useState<string[]>([]);
     const [transferFormat, setTransferFormat] = useState<'custom' | 'openvex'>('custom');
+    const [importTimestampPolicy, setImportTimestampPolicy] = useState<'original' | 'current'>('original');
 
     const showMessage = useCallback((message: string, type: "error" | "success") => {
         setBannerMessage(message);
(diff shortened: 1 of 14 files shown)


ValentinBoudevinSFL (dismissed):
SHould work with both OpenVex AND VulnScout format

ValentinBoudevin (changes requested):
Timestamp is not respect as it should with the option with the option "Use original timestamps from file": 

Image

Timestamp is 2026-07-20 but it appear as 7 July and stick with another assessment when imported: 

Image

ValentinBoudevin (approved):
LGTM

frontend/src/pages/Review.tsx
```diff
@@ -438,6 +439,7 @@ function Review({ variantId, projectId, onAssessmentChanged }: Readonly<Props>)
 
     const openTransfer = useCallback((mode: 'import' | 'export') => {
         setTransferFormat('custom');
+        setImportTimestampPolicy('original');
```
sfl-pe-sync, line 442: The UI always defaults imports to `original`, including VulnScout JSON, contradicting the documented and backend/CLI default of `current`. A normal custom-data import therefore unexpectedly backdates assessments. Default custom imports to `current` and OpenVEX imports to `original`, including when the format changes.

sfl-pe-sync (dismissed):
The timestamp-policy plumbing is otherwise coherent and introduces no third-party dependencies, but the frontend's custom-import default changes documented behavior. The PR should be revised before merge.

QuentinL-SFL (changes requested):
It seems to not take the hours correctly : 

Image

Image

It should be 17h17 and not 13h17

reh-sfl:
@QuentinL-SFL the time is correctly registered but it's converted to your local timezone when you display it. This could be fixed in another PR if we want to show time in UTC.

QuentinL-SFL (approved):
LGTM

PR #478, opened 2026-07-24
Fix: Classify outdated packages/variants correctly

ValentinBoudevin:

-Marks package findings as outdated per package/variant combination.
-Exposes historical package findings and their associated variants through the API.
-Displays Outdated badges in vulnerability details and review views.
-Adds authoritative backend validation so assessments cannot target findings that do not belong to the requested package/variant scope

### Status

- [ ] READY
- [ ] HOLD
- [X] WIP (Work-In-Progress)

### How to verify this change

Load a project with a variant which have two SBOM entries with outdated elements

SBOM tab: 

image

VulnModal:

image

image

(Same vulnerability) 

image

Review tab:

image

### DIFF
--- a/frontend/src/components/EditAssessment.tsx
+++ b/frontend/src/components/EditAssessment.tsx
@@ -28,6 +28,7 @@ type Props = {
     defaultSelectedPackages?: string[];
     variantPackageMap?: Record<string, string[]>;
     variantFindingsMap?: Record<string, Array<{ pkg: string; outdated: boolean }>>;
+    findingsLoading?: boolean;
 }
 
 function EditAssessment({
@@ -42,7 +43,8 @@ function EditAssessment({
     availablePackages,
     defaultSelectedPackages,
     variantPackageMap,
-    variantFindingsMap
+    variantFindingsMap,
+    findingsLoading = false
 }: Readonly<Props>) {
     const isImpactStatus = assessment.status === 'not_affected' || assessment.status === 'false_positive';
     const [status, setStatus] = useState(assessment.status || "under_investigation");
@@ -62,20 +64,27 @@ function EditAssessment({
     );
     const [includeOutdatedPackages, setIncludeOutdatedPackages] = useState(false);
     const outdatedPackages = useMemo(() => {
-        const findingStates = new Map<string, boolean>();
+        const packages = new Set<string>();
         for (const finding of Object.values(variantFindingsMap ?? {}).flat()) {
-            findingStates.set(finding.pkg, (findingStates.get(finding.pkg) ?? true) && finding.outdated);
+            if (finding.outdated) packages.add(finding.pkg);
         }
-        return new Set([...findingStates].filter(([, outdated]) => outdated).map(([pkg]) => pkg));
+        return packages;
     }, [variantFindingsMap]);
+    const packagesWithCurrentFindings = useMemo(() => new Set(
+        Object.values(variantFindingsMap ?? {}).flatMap(findings =>
+            findings.filter(finding => !finding.outdated).map(finding => finding.pkg)
+        )
+    ), [variantFindingsMap]);
     const packageOptions = useMemo(() => {
-        const options = new Set((availablePackages ?? []).filter(pkg => !outdatedPackages.has(pkg)));
+        const options = new Set((availablePackages ?? []).filter(pkg =>
+            !outdatedPackages.has(pkg) || packagesWithCurrentFindings.has(pkg)
+        ));
         for (const pkg of defaultSelectedPackages ?? []) options.add(pkg);
         if (includeOutdatedPackages) {
             for (const pkg of outdatedPackages) options.add(pkg);
         }
         return [...options];
-    }, [availablePackages, defaultSelectedPackages, includeOutdatedPackages, outdatedPackages]);
+    }, [availablePackages, defaultSelectedPackages, includeOutdatedPackages, outdatedPackages, packagesWithCurrentFindings]);
     const effectiveVariantPackageMap = useMemo(() => {
         if (!variantPackageMap) return undefined;
         const result: Record<string, string[]> = {};
@@ -325,25 +334,32 @@ function EditAssessment({
             </h3>
 
             {availableVariants && availableVariants.length > 0 && (
-                <div className="mt-2 mb-2 ml-1">
-                    <p className="text-sm font-medium text-gray-300 mb-1">Apply to variants:</p>
-                    <div className="flex flex-wrap gap-x-4 gap-y-1">
+                <div className="mt-3 rounded-lg border border-gray-600 bg-gray-800/40 p-3">
+                    <p className="mb-2 text-sm font-medium text-gray-200">Apply to variants:</p>
+                    <span className="float-right -mt-7 text-xs text-gray-400">{selectedVariantIds.length} selected</span>
+                    <div className="flex flex-wrap gap-2">
                         {availableVariants.map(v => {
                             const incompatible = allowedVariants !== null && !allowedVariants.has(v.id);
+                            const selected = selectedVariantIds.includes(v.id);
                             return (
                             <label
                                 key={v.id}
-                                className={['flex items-center gap-1.5 text-sm select-none', incompatible ? 'cursor-not-allowed opacity-40' : 'cursor-pointer'].join(' ')}
-                                title={incompatible ? 'No selected package is present in this variant' : undefined}
+                                className={[
+                                    'inline-flex items-center rounded-full border px-3 py-1.5 text-sm font-medium transition-colors select-none',
+                                    selected ? 'border-blue-400 bg-blue-500/20 text-blue-100' : 'border-gray-600 bg-gray-700/60 text-gray-300 hover:border-gray-500',
+                                    incompatible ? 'cursor-not-allowed opacity-40' : 'cursor-pointer',
+                                ].join(' ')}
+                                title={incompatible ? 'Not every selected package version applies to this variant' : undefined}
                             >
                                 <input
                                     type="checkbox"
-                                    checked={selectedVariantIds.includes(v.id)}
+                                    checked={selected}
                                     disabled={incompatible}
                                     onChange={(e) => handleVariantToggle(v.id, e.target.checked)}
-                                    className="accent-blue-500"
+                                    className="sr-only"
                                 />
-                                <span className="text-gray-200">{v.name}</span>
+                                <span aria-hidden="true" className="mr-1.5 text-xs">{selected ? '✓' : '+'}</span>
+                                <span>{v.name}</span>
                             </label>
                             );
                         })}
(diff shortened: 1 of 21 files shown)


frontend/src/components/VulnModal.tsx
```diff
@@ -1772,11 +1796,21 @@ type VariantScopedSnapshot = {
                                                 {group.packages.map(pkg => {
                                                     const { nameVersion, supplier } = splitPkgId(pkg);
                                                     const supplierName = extractSupplierName(supplier);
+                                                    const variant = availableVariants.find(v => v.id === firstAssess.variant_id);
+                                                    const finding = firstAssess.variant_id
+                                                        ? variantFindingsMap[firstAssess.variant_id]?.find(item => item.pkg === pkg)
+                                                        : undefined;
                                                     return (
                                                         <span key={pkg} className="inline-flex items-center px-2.5 py-0.5 rounded-full font-medium bg-blue-100 text-blue-800 dark:bg-blue-900 dark:text-blue-300">
```
sfl-pe-sync, line 1803: Every package in this grouped assessment is classified using `firstAssess.variant_id`. Groups can contain assessments from multiple variants, so packages belonging to another variant receive the first variant's label and outdated state. Build tags from each underlying assessment's `(variant_id, package)` pair, as the later `findingTags` logic does.

sfl-pe-sync (changes requested):
The package/variant classification approach is directionally sound, but reset behavior can bypass the outdated-package opt-in and one grouped display still attributes all packages to the first assessment's variant. These correctness regressions should be fixed before merge.

---
- `frontend/src/components/StatusEditor.tsx:262` [warning]: The clear-fields reset uses `availablePackages` rather than the filtered `currentPackages`. When the sole available package is outdated, resetting auto-selects that hidden package while `includeOutdatedPackages` is false, allowing submission without explicit opt-in. Reset from `currentPackages`/`initialPackages` and add a regression test for a vulnerability with only an outdated package.

reh-sfl (commented):
Removed.

src/routes/packages.py
```diff
@@ -239,6 +255,73 @@ def _multi_pkg_ids_for_variant(variant_uuid: uuid.UUID) -> set[uuid.UUID]:
...
+                if finding.vulnerability_id not in item["vulnerability_ids"]:
+                    item["vulnerability_ids"].append(finding.vulnerability_id)
+                if scan_source and scan_source not in item["sources"]:
+                    item["sources"].append(scan_source)
+
+            for item in outdated_rows.values():
+                item["finding_ids"].sort()
+                item["vulnerability_ids"].sort()
+                item["sources"].sort()
+            result.extend(sorted(
```
reh-sfl, line 324: **[MAJOR] `format=dict` key collision can misclassify active packages as outdated.** These outdated rows are appended to `result` after the active rows, and the `format=dict` return below keys by `name + "@" + version`. In project/multi-variant scope, a cross-variant package that shares a `name@version` with an active row overwrites it and is then reported `outdated: true` (in `list` format the pair also appears twice). Suggest keying/deduping by variant as well, and emitting an explicit `outdated: false` on active rows.

src/routes/packages.py
```diff
@@ -159,8 +174,9 @@ def _multi_pkg_ids_for_variant(variant_uuid: uuid.UUID) -> set[uuid.UUID]:
                 pkgs = sorted(pkg_lookup.values(), key=lambda p: p.name)
             active_scan_ids = sbom_ids
         else:
-            pkgs = Package.get_all()
+            pkgs = [] if outdated_only else Package.get_all()
             active_scan_ids = []
+            scope_variant_ids = [variant.id for variant in Variant.get_all()]
```
reh-sfl, line 181: **[MAJOR] Unbounded outdated computation when no scope filter is given.** With `include_outdated`/`outdated_only` and no `variant_id`/`project_id`, `scope_variant_ids` becomes *every* variant, which then drives a per-variant `active_sbom_scan_ids_for_variant(...)` loop (N+1) plus a large 5-table `DISTINCT` historical join across the whole dataset — a scalability / self-DoS cliff on large SBOMs. Suggest requiring a scope (or capping/paginating) for the outdated modes.

frontend/src/pages/TablePackages.tsx
```diff
@@ -231,6 +223,24 @@ function TablePackages({ packages, vulnerabilities = [], onShowVulns }: Readonly
         setMatchConditionError('');
     }, [vulnerabilities]);
 
+    useEffect(() => {
+        if (!showOnlyOutdated || packagesWithOutdated !== null || !onLoadOutdatedPackages) return;
```
reh-sfl, line 235: **[MAJOR] Stale outdated cache never refetches on variant/project change.** This guard (`packagesWithOutdated !== null`) blocks any reload once the cache is set, and the component stays mounted across variant/project switches while `showOnlyOutdated` persists. After changing the variant with the toggle on, it keeps showing the *previous* variant's outdated rows and never refetches. Suggest invalidating the cache (or keying the component) on the variant/project identity.

frontend/src/components/EditAssessment.tsx
```diff
@@ -112,10 +143,10 @@ function EditAssessment({
             : selectedVariantIds.filter(id => id !== variantId);
         setSelectedVariantIds(nextVariants);
 
-        if (!checked && variantPackageMap) {
+        if (!checked && effectiveVariantPackageMap) {
             const stillAllowed = new Set<string>();
             for (const vid of nextVariants) {
-                for (const pkg of variantPackageMap[vid] ?? []) stillAllowed.add(pkg);
+                for (const pkg of effectiveVariantPackageMap[vid] ?? []) stillAllowed.add(pkg);
```
reh-sfl: **[MAJOR] `union` semantics contradict the intersection model.** On unchecking a variant, retained packages are computed as the **union** across remaining variants (`stillAllowed.add(pkg)` for each), while validity elsewhere was redefined as the **intersection** (`allowedPackages`, filtered with `.every(...)`). Unchecking can therefore leave packages present in only *some* remaining variant. Currently masked by the check-side `blocked` gating and caught by backend atomic validation (HTTP 400), but latent. Suggest mirroring the `.every()` intersection logic here. (Same issue in `StatusEditor.tsx`.)

frontend/src/components/StatusEditor.tsx
```diff
@@ -125,10 +148,10 @@ function StatusEditor ({onAddAssessment, progressBar, clearFields: shouldClearFi
             : selectedVariantIds.filter(id => id !== variantId);
         setSelectedVariantIds(nextVariants);
 
-        if (!checked && variantPackageMap) {
+        if (!checked && effectiveVariantPackageMap) {
             const stillAllowed = new Set<string>();
             for (const vid of nextVariants) {
-                for (const pkg of variantPackageMap[vid] ?? []) stillAllowed.add(pkg);
+                for (const pkg of effectiveVariantPackageMap[vid] ?? []) stillAllowed.add(pkg);
```
reh-sfl: **[MAJOR] Same `union`-vs-intersection issue as `EditAssessment.tsx`.** Unchecking a variant retains the union of packages across remaining variants instead of the `.every(...)` intersection used for validity. Suggest aligning to the intersection logic.

reh-sfl (changes requested):
Automated review of the outdated-package/variant work. Requesting changes on the **MAJOR** items below (details left as inline comments on the relevant lines). Minor/nit items and test-coverage gaps are omitted per request.

- packages.py: dict-format key collision mis-flagging active packages as outdated
- packages.py: unbounded outdated computation when no variant/project scope is given
- TablePackages.tsx: stale outdated cache never refetches on variant/project change
- EditAssessment.tsx / StatusEditor.tsx: `union` scope semantics contradict the intersection model

frontend/src/components/StatusEditor.tsx
ValentinBoudevinSFL: Done

frontend/src/components/EditAssessment.tsx
ValentinBoudevinSFL: Done

src/routes/packages.py
ValentinBoudevinSFL: Fixed and added tests. 

The guard itself remains because it prevents duplicate requests within the same scope. The fix was to invalidate the cache when the scope changes.

Made changes in Explorer.tsx and TablePackages to clear outdated rows

frontend/src/pages/TablePackages.tsx
ValentinBoudevinSFL: Two changes prevent an outdated row from replacing an active row:

Active rows now explicitly contain outdated:false in package.py and historical rows include their variant_id

src/routes/packages.py
ValentinBoudevinSFL: Outdated modes now require one of variant_id, project_id or variant_ids

frontend/src/components/EditAssessment.tsx
```diff
@@ -58,52 +71,70 @@ function EditAssessment({
...
+    const packagesWithCurrentFindings = useMemo(() => new Set(
+        Object.values(variantFindingsMap ?? {}).flatMap(findings =>
+            findings.filter(finding => !finding.outdated).map(finding => finding.pkg)
+        )
+    ), [variantFindingsMap]);
+    const packageOptions = useMemo(() => {
+        const options = new Set((availablePackages ?? []).filter(pkg =>
+            !outdatedPackages.has(pkg) || packagesWithCurrentFindings.has(pkg)
+        ));
+        for (const pkg of defaultSelectedPackages ?? []) options.add(pkg);
```
sfl-pe-sync: Turning off the outdated-package toggle does not actually hide or block the originally selected outdated package: `defaultSelectedPackages` is always added to `packageOptions`, and line 104 also keeps it in every applicable variant map. After the toggle clears selections, the user can immediately reselect that historical package and submit without opting in. Only retain an outdated default while `includeOutdatedPackages` is true, and add a regression test covering disabling the toggle while editing an outdated assessment.

frontend/src/components/VulnModal.tsx
```diff
@@ -1800,19 +1834,31 @@ type VariantScopedSnapshot = {
...
-                                        if (a.outdated && a.superseded_map) {
-                                            for (const [ref, versions] of Object.entries(a.superseded_map)) {
-                                                // Union the replacement versions when several assessments
-                                                // in the group flag the same package reference.
-                                                supersededMap[ref] = [...new Set([...(supersededMap[ref] ?? []), ...versions])].sort();
-                                            }
-                                        }
-                                    }
+                                    const groupDateKey = new Date(group.timestamp).toDateString();
+                                    const groupFingerprint = `${firstAssess.simplified_status}|${firstAssess.justification || ''}|${firstAssess.impact_statement || ''}|${firstAssess.status_notes || ''}|${firstAssess.workaround || ''}`;
```
sfl-pe-sync: The expansion from `allVulnAssessments` identifies a batch only by calendar date and assessment-field fingerprint. Two independent assessments for the same CVE created on the same day with common/default fields will therefore be treated as one batch, causing unrelated package/variant tags to be displayed on this group. Match using a stable batch/group identifier or the exact assessment IDs instead of day-level timestamps, with a test for separate same-day assessments having identical content.

sfl-pe-sync (changes requested):
The classification changes are not ready to merge: outdated edit selection can bypass its opt-in control, and the new assessment-tag expansion can misrepresent package/variant scope by conflating independent same-day assessments.

frontend/src/components/VulnModal.tsx
```diff
@@ -322,20 +330,33 @@ type VariantScopedSnapshot = {
...
                 if (!signal.aborted) {
                     setVariantPackageMap(map);
+                    setVariantFindingsMap(findingsMap);
                     setVariantPackageMapLoaded(true);
                 }
             } catch {
                 if (!signal.aborted) {
                     setVariantPackageMap({});
+                    setVariantFindingsMap({});
                     setVariantPackageMapLoaded(true);
```
sfl-pe-sync, line 360: The findings request fails open: on any network, authorization, or parsing failure, the code clears both maps and marks them loaded. Editors then receive no variant/package compatibility map, hide the outdated-package toggle, and allow an existing historical package/variant selection to be submitted without explicit opt-in. Preserve an error state and disable assessment editing until findings load successfully, or conservatively classify the existing assessment as outdated. Add a regression test for a failed findings request while editing an outdated assessment.

sfl-pe-sync (changes requested):
The PR improves package/variant classification, but the new findings-loading path permits outdated assessments when its API request fails. This fail-open behavior should be corrected before merge. Review was limited to the supplied diff; no commands, clean startup, or container validation were performed.

reh-sfl (approved):

PR #479, opened 2026-07-27
Scan: add new progress popup

reh-sfl:

Add new scan progress popup
This allows the user to see progress of queued / active scans within the app. It's especially useful when launching multiple NVD CPE / sbom-cve-check scans for example.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Open the web dashboard, go to Scan then launch a certain number of scans. Observe that:
1. a popup with progress shows up
2. It can be closed
3. You can come back to it by clicking on the scan status in the navbar

## Pull Request Checklist


ValentinBoudevinSFL (changes requested):
Image

I believe Vulnerability "Refresh Vulnerabilities Data" operations can be added in the same queue. 

Right now it leads to errors to start a refresh while some elements are in the queue

ValentinBoudevinSFL (approved):

ValentinBoudevinSFL (dismissed):

frontend/src/handlers/scanStateManager.ts
```diff
@@ -107,9 +121,57 @@ export class ScanStateManager {
...
+        this.rebuildSnapshot();
+    };
+
+    /** Start the next previously queued serial scan batch. */
+    startQueuedScan = async () => {
+        if (!this.serial || this.pendingQueue.length === 0) return;
+
+        this.triggerNextInQueue();
+        if ([...this.states.values()].some((state) => state.status === "running")) {
+            this.startPolling();
```
sfl-pe-sync, line 163: ERROR: `startQueuedScan` starts polling immediately after `triggerNextInQueue()`, but that method only launches `invokeTrigger(...).then(...)` and returns before the POST is accepted. If the trigger request takes longer than the 3-second poll interval, the status endpoint can still report `idle`; the polling logic treats that as terminal, advances the queue, and may start another variant while the first trigger later succeeds. Make `triggerNextInQueue` await and return the trigger result, and do not start polling until a successful trigger has completed (while still advancing synchronously on trigger failure).

frontend/src/pages/ScanHistory.tsx
```diff
@@ -1332,12 +1309,25 @@ function ScanHistory({ variantId, projectId, onScanComplete }: Readonly<Props>)
...
+        await restoreActiveRefreshes();
+        const refreshAheadOfScans = hasActiveRefreshes();
+        await Promise.all(scanQueue.map(async ([scanType, queueScan]) => {
+            if (selectedScanTypes.has(scanType)) await queueScan(variants, opts);
+        }));
+
+        if (refreshAheadOfScans) await waitForRefreshCompletion();
+        for (const [scanType, , startQueuedScan, waitForCompletion] of scanQueue) {
+            if (!selectedScanTypes.has(scanType)) continue;
+            await startQueuedScan();
```
sfl-pe-sync, line 1335: ERROR: This loop serializes only the batches selected by this click; it never waits for an already-running scan in an unselected manager. For example, after Explorer restores an active Grype scan, selecting only NVD queues and starts NVD immediately, so the two backend scans overlap despite the new global queue contract (and refreshes rely on that same contract). Include all pre-existing active scan-manager completion promises in the global scheduler before starting the first newly queued batch, rather than coordinating only the selected managers.

frontend/src/pages/TableVulnerabilities.tsx
```diff
@@ -1926,6 +1926,7 @@ function TableVulnerabilities ({ vulnerabilities, filterLabel, filterValue, filt
                 <span className="h-6 border-l border-gray-400" aria-hidden="true" />
                 <RefreshVulnerabilityData
                     vulnerabilities={vulnerabilities}
+                    getRefreshVulnerabilities={() => Vulnerabilities.list(variantId, projectId, baseVariantId, compareOperation)}
```
sfl-pe-sync: ERROR: The reload callback drops the current multi-variant scope: `variantIds` and `multiOperation` are not forwarded to `Vulnerabilities.list`. In union/intersection/difference views, the queued refresh therefore reloads a different vulnerability set after scans finish and sends the wrong CVE/GHSA IDs to the bulk refresh endpoints. Mirror the full argument list used to load the displayed vulnerability scope, including the selected variant IDs and multi-operation.

sfl-pe-sync (changes requested):
The progress UI is well-contained, but the new orchestration can misreport or overlap scans and can refresh vulnerabilities outside the selected multi-variant view, so the PR is not ready to merge.

sfl-zwang (dismissed):

frontend/src/pages/Explorer.tsx
```diff
@@ -66,6 +74,42 @@ function Explorer() {
...
+        .filter(entry => entry.status === "queued" || entry.status === "running").length;
+
+    useEffect(() => {
+        if (activeScanCount > 0 && !hadActiveScans.current) {
+            setScanProgressOpen(true);
+        }
+        hadActiveScans.current = activeScanCount > 0;
+    }, [activeScanCount]);
+
+    useEffect(() => {
```
sfl-pe-sync, line 98: ERROR: Active-scan restoration is started asynchronously and is not part of the scheduler barrier. Immediately after a reload, a user can launch scans before `getRunningScans()` resolves; `waitForActiveScans()` then sees empty manager snapshots and starts the new batch while a backend scan restored moments later is still running. Expose a shared restoration/readiness promise and await it in scan and refresh queue entry points before snapshotting active work (and handle restoration failure explicitly), so the global queue cannot start until backend activity has been discovered.

sfl-pe-sync (changes requested):
The latest patch fixes the previously reported trigger, pre-existing-manager, and multi-variant-scope defects, but startup restoration still races user-triggered queue operations and can allow overlapping backend scans.

frontend/src/handlers/activeScanQueue.ts
```diff
@@ -0,0 +1,296 @@
...
+    if (ids.length === 0) {
+        setRefreshEntry(type, { status: "done", progress: "No matching vulnerabilities", logs: [`No ${source.noun} were available after scans completed.`] });
+        return;
+    }
+
+    let baseline: RefreshProgress | null = null;
+    try {
+        baseline = await source.getProgress();
+    } catch {
+        // A baseline is optional; active progress still identifies the new cycle.
```
sfl-pe-sync, line 209: ERROR: Treating the baseline as optional makes an accepted fast refresh look as though it never started. If this GET fails and the refresh completes before the first 3-second poll, every observed response is the new terminal `completed` state, but `cycleChanged(null, progress)` can never recognize it; the UI remains `running` for 60 seconds, reports a false failure, and delays later queued sources. Require a reliable pre-trigger baseline or have the trigger return a cycle identifier/timestamp that terminal progress can be matched against, and add a test for baseline failure followed immediately by the new completed state.

sfl-pe-sync (changes requested):
The progress popup and queue are substantially integrated, but refresh-cycle detection can falsely fail and stall a successfully completed queued refresh when baseline polling is unavailable.

PR #484, opened 2026-07-28
feat(ui): unify search interactions

ValentinBoudevin:

Require explicit search application across the SBOM, vulnerabilities, and review tabs through Enter or accessible magnifying-glass buttons. Apply the same interaction to SBOM match conditions and add regression coverage.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Test search bar in SBOM, Vulnerabilities and Review tab

image

### DIFF
--- a/frontend/src/components/ExplicitSearchInput.tsx
+++ b/frontend/src/components/ExplicitSearchInput.tsx
@@ -1,0 +1,71 @@
+import { forwardRef } from 'react';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faMagnifyingGlass } from '@fortawesome/free-solid-svg-icons';
+
+type Props = {
+    id: string;
+    label: string;
+    value: string;
+    onChange: (value: string) => void;
+    onSearch: () => void | Promise<void>;
+    placeholder: string;
+    ariaLabel: string;
+    loading?: boolean;
+    ariaInvalid?: boolean;
+    labelClassName?: string;
+    inputClassName?: string;
+    inputType?: 'search' | 'text';
+};
+
+const ExplicitSearchInput = forwardRef<HTMLInputElement, Readonly<Props>>(function ExplicitSearchInput({
+    id,
+    label,
+    value,
+    onChange,
+    onSearch,
+    placeholder,
+    ariaLabel,
+    loading = false,
+    ariaInvalid = false,
+    labelClassName,
+    inputClassName = 'py-1 px-2 bg-sky-900 focus:bg-sky-950 min-w-[250px] grow max-w-[800px]',
+    inputType = 'search',
+}, ref) {
+    const applySearch = () => void onSearch();
+
+    return <>
+        <label htmlFor={id} className={labelClassName}>{label}</label>
+        <input
+            id={id}
+            ref={ref}
+            value={value}
+            onChange={event => onChange(event.target.value)}
+            onKeyDown={event => {
+                if (event.key === 'Enter') {
+                    event.preventDefault();
+                    applySearch();
+                }
+            }}
+            type={inputType}
+            aria-invalid={ariaInvalid}
+            className={inputClassName}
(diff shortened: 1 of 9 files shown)


frontend/src/components/ExplicitSearchInput.tsx
```diff
@@ -0,0 +1,71 @@
...
+}, ref) {
+    const applySearch = () => void onSearch();
+
+    return <>
+        <label htmlFor={id} className={labelClassName}>{label}</label>
+        <input
+            id={id}
+            ref={ref}
+            value={value}
+            onChange={event => onChange(event.target.value)}
```
reh-sfl, line 42: **[MAJOR] Clearing the input no longer clears the applied results.** `onChange` only updates the draft value; the applied search state (`search`) is untouched until the user presses Enter or clicks the button. Combined with the default `type="search"` (line 49), the browser's native clear (✕) and manual deletion set the draft to `''` but leave the previously-applied results in place until the user explicitly re-searches on an empty field. The handlers this replaces (the debounced `updateSearch` in `Review`/`TablePackages`) auto-cleared results once the term dropped below 2 chars, so this is a behavioral regression. Suggest auto-applying when the field transitions to empty (e.g. call `onSearch()` when `value === ''`), or wiring the native clear event to also apply.

frontend/src/components/ExplicitSearchInput.tsx
```diff
@@ -0,0 +1,71 @@
...
+    loading = false,
+    ariaInvalid = false,
+    labelClassName,
+    inputClassName = 'py-1 px-2 bg-sky-900 focus:bg-sky-950 min-w-[250px] grow max-w-[800px]',
+    inputType = 'search',
+}, ref) {
+    const applySearch = () => void onSearch();
+
+    return <>
+        <label htmlFor={id} className={labelClassName}>{label}</label>
```
reh-sfl, line 37: **[MAJOR] Inconsistent accessible names for one control pair.** The visible `<label>` names the input (e.g. "Search"), while the sibling button's `aria-label`/`title` (line 56-57) is a different phrase ("Search packages", "Search reviews", "Search match condition"). Screen-reader users hear two unrelated names for what is visually a single search widget. Align them — e.g. derive the button's accessible name from the same `label`, or give the input a matching accessible name. Also confirm the icon-only button conveys its purpose without relying solely on color/title (the `title` tooltip is not exposed to all AT).

frontend/tests/unit_tests/test_explicit_search_input.tsx
```diff
@@ -0,0 +1,55 @@
...
+                placeholder="Search packages"
+                ariaLabel="Search packages"
+            />
+        );
+
+        const input = screen.getByRole('searchbox');
+        await user.type(input, '{enter}');
+        await user.click(screen.getByRole('button', {name: 'Search packages'}));
+
+        expect(onSearch).toHaveBeenCalledTimes(2);
```
reh-sfl, line 30: **[MAJOR] The test does not cover the feature's core guarantee.** The whole point of `ExplicitSearchInput` is that typing must *not* apply the search — yet this test only asserts that Enter + click produce 2 `onSearch` calls. It would still pass if the component searched on every keystroke. Add a negative assertion: type a multi-character string and assert `onSearch` was **not** called. Additionally, `value` is hardcoded and `onChange` is never wired back into `value`, so the controlled draft→apply flow is never exercised end-to-end; consider a small controlled wrapper so the test reflects real usage.

reh-sfl (changes requested):
Automated review on #484

frontend/src/components/ExplicitSearchInput.tsx
ValentinBoudevin: Disagree with this take

reh-sfl (approved):

PR #487, opened 2026-07-29
Add outdated data cleanup

reh-sfl:

* Add outdated data cleanup

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

1. Open the web dashboard, go to Settings -> Scan then select "Delete outdated data"

2. Use the CLI:
```bash
./vulnscout --delete-outdated
```

## Pull Request Checklist

### DIFF
--- a/frontend/src/handlers/scans.ts
+++ b/frontend/src/handlers/scans.ts
@@ -167,6 +167,11 @@ type RunningScans = {
 export type { ScanStatusResponse, RunningScanEntry, RunningScans };
 
 type OutdatedDataPreview = {
+    candidate_ids: {
+        observations: string[];
+        assessments: string[];
+        package_pairs: Array<{ package_id: string; variant_id: string }>;
+    };
     packages: Array<{
         project: string;
         variant: string;
@@ -177,7 +182,20 @@ type OutdatedDataPreview = {
     assessments: Array<{ project: string; vulnerability: string; package: string; variant: string }>;
 };
 
-export type { OutdatedDataPreview };
+type EmptyScanPreview = {
+    id: string;
+    description: string;
+    timestamp: string;
+    project: string;
+    variant: string;
+};
+
+type OrphanedVulnerabilityPreview = {
+    id: string;
+    assessments: number;
+};
+
+export type { OutdatedDataPreview, EmptyScanPreview, OrphanedVulnerabilityPreview };
 
 class ScansHandler {
     static async list(variantId?: string, projectId?: string): Promise<Scan[]> {
@@ -343,10 +361,10 @@ class ScansHandler {
         return { ok: false, error: data?.error ?? `HTTP ${response.status}` };
     }
 
-    static async deleteOutdatedData(): Promise<{ ok: boolean; error?: string; summary?: Record<string, number> }> {
+    static async deleteOutdatedData(candidateIds: OutdatedDataPreview["candidate_ids"]): Promise<{ ok: boolean; error?: string; summary?: Record<string, number> }> {
         const response = await fetch(
             import.meta.env.VITE_API_URL + '/api/outdated-data',
-            { method: 'DELETE', mode: 'cors' }
+            { method: 'DELETE', mode: 'cors', headers: { 'Content-Type': 'application/json' }, body: JSON.stringify({ candidate_ids: candidateIds }) }
         );
         const data = await response.json().catch(() => ({}));
         if (response.ok) return { ok: true, summary: data };
@@ -362,6 +380,45 @@ class ScansHandler {
         if (response.ok && Array.isArray(data?.packages) && Array.isArray(data?.assessments)) {
             return { ok: true, preview: data as OutdatedDataPreview };
         }
+        return { ok: false, error: data?.error ?? `HTTP ${response.status}` };
+    }
+
+    static async getEmptyScansPreview(): Promise<{ ok: boolean; error?: string; scans?: EmptyScanPreview[] }> {
+        const response = await fetch(import.meta.env.VITE_API_URL + '/api/empty-scans', { mode: 'cors' });
+        const data = await response.json().catch(() => ({}));
+        if (response.ok && Array.isArray(data?.scans)) return { ok: true, scans: data.scans };
+        return { ok: false, error: data?.error ?? `HTTP ${response.status}` };
+    }
+
+    static async deleteEmptyScans(scanIds: string[]): Promise<{ ok: boolean; error?: string; count?: number }> {
+        const response = await fetch(
+            import.meta.env.VITE_API_URL + '/api/empty-scans',
+            { method: 'DELETE', mode: 'cors', headers: { 'Content-Type': 'application/json' }, body: JSON.stringify({ scan_ids: scanIds }) }
+        );
+        const data = await response.json().catch(() => ({}));
+        if (response.ok) return { ok: true, count: data?.scans_deleted ?? 0 };
+        return { ok: false, error: data?.error ?? `HTTP ${response.status}` };
+    }
+
+    static async getOrphanedVulnerabilitiesPreview(): Promise<{ ok: boolean; error?: string; vulnerabilities?: OrphanedVulnerabilityPreview[] }> {
+        const response = await fetch(import.meta.env.VITE_API_URL + '/api/orphaned-vulnerabilities', { mode: 'cors' });
+        const data = await response.json().catch(() => ({}));
+        if (response.ok && Array.isArray(data?.vulnerabilities)) {
+            return {
+                ok: true,
+                vulnerabilities: data.vulnerabilities,
+            };
+        }
+        return { ok: false, error: data?.error ?? `HTTP ${response.status}` };
+    }
+
+    static async deleteOrphanedVulnerabilities(vulnerabilityIds: string[]): Promise<{ ok: boolean; error?: string; count?: number }> {
+        const response = await fetch(
+            import.meta.env.VITE_API_URL + '/api/orphaned-vulnerabilities',
+            { method: 'DELETE', mode: 'cors', headers: { 'Content-Type': 'application/json' }, body: JSON.stringify({ vulnerability_ids: vulnerabilityIds }) }
+        );
+        const data = await response.json().catch(() => ({}));
+        if (response.ok) return { ok: true, count: data?.vulnerabilities_deleted ?? 0 };
         return { ok: false, error: data?.error ?? `HTTP ${response.status}` };
     }
 }
--- a/frontend/src/handlers/scans.ts
+++ b/frontend/src/handlers/scans.ts
@@ -166,6 +166,19 @@ type RunningScans = {
 
 export type { ScanStatusResponse, RunningScanEntry, RunningScans };
 
+type OutdatedDataPreview = {
+    packages: Array<{
+        project: string;
+        variant: string;
+        package: string;
+        vulnerabilities: string[];
+        linked_data: { observations: number; sbom_packages: number; sbom_observations: number };
+    }>;
+    assessments: Array<{ project: string; vulnerability: string; package: string; variant: string }>;
+};
+
+export type { OutdatedDataPreview };
+
 class ScansHandler {
     static async list(variantId?: string, projectId?: string): Promise<Scan[]> {
         let url: string;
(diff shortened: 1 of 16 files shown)


ValentinBoudevinSFL (changes requested):
Should also have other options. 

1- Delete empty scans (scans which don't bring any information with new CVEs, assessments...)
2- Delete CVEs which aren't included in any variants AND any projects. And delete their corresponding assessments

frontend/src/handlers/scans.ts
```diff
@@ -329,6 +355,67 @@ class ScansHandler {
         const data = await response.json().catch(() => ({}));
         return { ok: false, error: data?.error ?? `HTTP ${response.status}` };
     }
+
+    static async deleteOutdatedData(): Promise<{ ok: boolean; error?: string; summary?: Record<string, number> }> {
```
ValentinBoudevinSFL: ERROR: The DELETE request carries none of the IDs returned by the confirmation preview, so all three maintenance operations recompute their candidates at confirmation time. If an SBOM, scan, or assessment changes after the modal opens, the server can delete newly-outdated packages, newly-empty scans, or newly-orphaned CVEs (including assessments) that the user never approved. Return exact candidate IDs or an opaque snapshot token from each preview and require DELETE to operate only on that snapshot, rejecting drift.

src/helpers/outdated_cleanup.py
```diff
@@ -0,0 +1,516 @@
...
+def empty_scans_preview() -> list[dict[str, str]]:
+    """Return non-initial scans whose complete history diff is empty.
+
+    Reuses the cached scan-history serialisation (scope ``"all"``) that the
+    ``/api/scans`` view warms, so a preview issued after the Scans tab has
+    loaded is served without recomputing every scan diff.
+    """
+    from ..routes._scan_diff import serialize_list_with_diff_cached
+
+    scans = serialize_list_with_diff_cached("all", Scan.get_all())
```
ValentinBoudevinSFL: ERROR: Destructive empty-scan selection uses `serialize_list_with_diff_cached`, whose fingerprint tracks scan-row metadata but not all diff-driving assessment/finding changes. A cached zero-diff result can therefore survive after a scan becomes non-empty, causing `delete_empty_scans()` to delete real history. Use the uncached diff computation for deletion, or make cache invalidation/fingerprinting cover every table that contributes to the diff before relying on it here.

src/routes/scans.py
```diff
@@ -403,6 +403,60 @@ def delete_scan(scan_id: str) -> ResponseReturnValue:
...
+            outdated_data_preview,
+        )
+
+        if flask_request.method == 'GET':
+            return jsonify(outdated_data_preview())
+        try:
+            return jsonify(cleanup())
+        except Exception as exc:
+            db.session.rollback()
+            return jsonify({"error": f"Failed to delete outdated data: {exc}"}), 500
```
ValentinBoudevinSFL: WARNING: Returning `str(exc)` exposes SQLAlchemy/DBAPI exception details to API clients; integrity and statement errors commonly include SQL text and bound parameters. Log the exception server-side and return a fixed, non-sensitive error message for each cleanup endpoint.

ValentinBoudevinSFL (changes requested):
The cleanup feature is not safe to merge yet: confirmation is not tied to the previewed records, and empty-scan deletion can act on stale cached diff data, both permitting unintended deletion. The new error responses can also disclose database details.

src/routes/scans.py
```diff
@@ -403,6 +405,68 @@ def delete_scan(scan_id: str) -> ResponseReturnValue:
...
+        """
+        from ..helpers.outdated_cleanup import (
+            delete_outdated_data as cleanup,
+            outdated_data_preview,
+        )
+
+        if flask_request.method == 'GET':
+            return jsonify(outdated_data_preview())
+        try:
+            return jsonify(cleanup((flask_request.get_json(silent=True) or {}).get("candidate_ids")))
```
ValentinBoudevinSFL: ERROR: An empty or malformed DELETE body still passes `None` to the helper, and `candidate_ids is None` explicitly bypasses snapshot validation, so a direct API request can perform this global destructive cleanup without ever obtaining or confirming a preview. The same bypass exists for `scan_ids` and `vulnerability_ids` below. Require each API DELETE to contain the correctly shaped candidate list/object and return 400 when it is missing; keep the no-preview behavior only in the CLI call path.

src/helpers/outdated_cleanup.py
```diff
@@ -0,0 +1,547 @@
...
+
+def delete_outdated_data(candidate_ids: dict[str, object] | None = None) -> dict[str, int]:
+    """Delete every package observation and custom assessment marked outdated.
+
+    The predicates deliberately mirror ``/api/packages?outdated_only=true``
+    and ``annotate_assessments_outdated``.  Shared records are retained until
+    no scan, assessment, or SBOM relation references them.
+    """
+    with write_lock():
+        stale_observation_ids, stale_package_pairs, stale_finding_ids = _stale_observations(
```
ValentinBoudevinSFL: ERROR: Stale package/variant pairs are derived only from `Observation` rows, so a package that existed in an old SBOM but never had a vulnerability finding produces no pair and its historical `SBOMPackage` link and package row are never cleaned. This leaves exactly the obsolete package evidence the feature promises to remove. Derive stale pairs from historical `SBOMPackage` rows compared with each variant's active SBOM (and union observation-derived pairs as needed), then delete those links and prune unreferenced packages.

src/helpers/outdated_cleanup.py
```diff
@@ -0,0 +1,547 @@
...
+        }
+        for item in scans
+        if not item.get("is_first")
+        and all(item.get(field) == 0 for field in _SCAN_CHANGE_FIELDS)
+    ]
+
+
+def delete_empty_scans(candidate_ids: list[str] | None = None) -> dict[str, int]:
+    """Delete scans selected by :func:`empty_scans_preview`."""
+    current_ids = [scan["id"] for scan in empty_scans_preview()]
```
ValentinBoudevinSFL: ERROR: Empty-scan candidate recomputation, validation, and deletion are not held under `write_lock()`. A concurrent scan/import can add an observation after `empty_scans_preview()` returns but before the delete commits; deleting the `Scan` then cascades to that newly added real data even though the candidate comparison succeeded. Enclose recomputation through commit in the same write lock, as the other destructive cleanup functions do (and use equivalent row/transaction locking on databases where that lock is inactive).

ValentinBoudevinSFL (changes requested):
The cleanup remains unsafe to merge: API callers can bypass preview validation, stale SBOM-only packages are omitted entirely, and empty-scan deletion has a race that can remove newly written scan data.

ValentinBoudevinSFL (changes requested):
Image

Image

I did the three cleans but some vulnerabilities (eg CVE-2001-1267) still exists but should not. 

Here I have two imported SBOM. So everything unique to the previous SBOM should be erase.

src/helpers/outdated_cleanup.py
```diff
@@ -0,0 +1,572 @@
...
+    }
+
+
+_SCAN_CHANGE_FIELDS = (
+    "packages_added",
+    "packages_removed",
+    "packages_upgraded",
+    "findings_added",
+    "findings_removed",
+    "findings_upgraded",
```
ValentinBoudevinSFL, line 470: ERROR: Empty-scan eligibility only checks the listed package/finding/CVE/assessment counters, but deleting a selected SBOM scan cascades its `SBOMDocument` and `SBOMObservation` rows. A replacement SBOM with unchanged package/CVE sets but updated package-less metadata (for example a new Yocto VEX description) therefore has a zero diff, is offered as empty, and deletion discards the new metadata and makes the previous SBOM active again. Exclude SBOM scans whenever document-level content differs, or extend the diff/eligibility check to cover every SBOM child record and metadata field that deletion would remove.

ValentinBoudevinSFL (changes requested):
The cleanup is still unsafe because empty-scan classification ignores SBOM document metadata that is cascaded on scan deletion, allowing a metadata-only replacement scan to be treated as redundant and deleted.

src/routes/scans.py
sfl-pe-sync: WARNING: Agreed. Log the original exception with `app.logger.exception(...)` after rollback and return a fixed endpoint-specific 500 response so SQL text, parameters, and database details never cross the API boundary.

sfl-pe-sync (changes requested):
Request changes: the cleanup endpoints can perform unconfirmed global deletion, outdated cleanup misses SBOM-only historical packages, and empty-scan cleanup can race with imports or delete SBOM metadata not represented by its aggregate diff.

src/routes/scans.py
sfl-pe-sync: ERROR: Agreed. Each DELETE route must reject a missing/non-object JSON body and require its candidate field with the exact expected shape and element types; reserve `None` exclusively for direct CLI calls that intentionally recompute and delete in one operation.

src/helpers/outdated_cleanup.py
sfl-pe-sync: ERROR: Agreed. `SBOMPackage` links are evidence independently of findings, so candidate pairs must originate from historical SBOM package links compared against the active SBOM, then be unioned with observation-derived pairs before pruning unreferenced package rows.

src/helpers/outdated_cleanup.py
sfl-pe-sync: ERROR: Agreed. Hold `write_lock()` across candidate recomputation, candidate comparison, deletion, and commit; otherwise a scan/import can attach data to a previously empty scan between validation and deletion and have it removed by cascade.

src/helpers/outdated_cleanup.py
sfl-pe-sync: ERROR: Agreed. A zero aggregate package/finding/CVE/assessment diff is not sufficient for SBOM scans. Eligibility must compare all `SBOMDocument` and `SBOMObservation` content that would cascade, or conservatively exclude SBOM scans whenever document-level payload differs.

src/routes/scans.py
sfl-zwang: Please fix this

src/helpers/outdated_cleanup.py
sfl-zwang: Please fix this.

sfl-zwang (dismissed):

src/helpers/outdated_cleanup.py
```diff
@@ -0,0 +1,593 @@
...
+) -> set[StalePackagePair]:
+    """Return historical SBOM package links absent from each variant's active SBOM."""
+    pairs: set[StalePackagePair] = set()
+    for package_id, name, version, variant_id in db.session.execute(
+        db.select(SBOMPackage.package_id, Package.name, Package.version, Scan.variant_id)
+        .join(Package, Package.id == SBOMPackage.package_id)
+        .join(SBOMDocument, SBOMDocument.id == SBOMPackage.sbom_document_id)
+        .join(Scan, Scan.id == SBOMDocument.scan_id)
+    ):
+        if (name, version) not in active_identities_by_variant.get(variant_id, set()):
```
sfl-pe-sync, line 131: WARNING: Package identity here ignores `supplier`, even though `Package.find_or_create`, equality, and `string_id` treat `(name, version, supplier)` as the package identity. If an old SBOM contains `foo@1::SupplierA` and the active SBOM contains `foo@1::SupplierB`, the old link is considered active and is never cleaned. Include normalized supplier in `PackageIdentity` and in both the active-identity and historical-link queries.

src/helpers/outdated_cleanup.py
```diff
@@ -0,0 +1,593 @@
...
+from ..models.time_estimate import TimeEstimate
+from ..models.variant import Variant
+from ..models.vuln_refresh import VulnRefresh
+from ..models.vulnerability import Vulnerability
+from .assessment_staleness import annotate_assessments_outdated
+
+PackageIdentity = tuple[str | None, str | None]
+StalePackagePair = tuple[uuid.UUID, uuid.UUID]
+UNKNOWN_PROJECT = "Unknown project"
+UNKNOWN_VARIANT = "Unknown variant"
```
sfl-pe-sync, line 42: WARNING: This global preview performs two SQL queries per variant (`active_sbom_scan_ids_for_variant` plus the package query), so request cost is `2N+1` queries before the subsequent whole-database cleanup queries. Databases with many projects/variants can make the Settings preview slow or time out. Batch-fetch the latest SBOM scan and its package identities for all variants in a constant number of queries, as `assessment_staleness` already does.

sfl-pe-sync (approved):
The destructive-preview validation and empty-scan locking fixes are correct; two non-blocking cleanup completeness and scaling concerns remain.

src/routes/scans.py
sfl-pe-sync: Addressed in d74d955: all three HTTP DELETE routes now reject missing or malformed candidate payloads with 400, while CLI helpers retain the intentional `None` path.

src/helpers/outdated_cleanup.py
sfl-pe-sync: Addressed in d74d955: empty-scan recomputation, candidate comparison, deletion, and commit now execute within the same `write_lock()` scope.

ValentinBoudevinSFL (approved):
LGTM

PR #488, opened 2026-07-29
Add tooltip hints for SBOM and vulnerabilities tables

reh-sfl:

Add tooltip hints for SBOM and vulnerabilities tables

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Try tooltips in the table header of SBOM and Vulnerabilities.

## Pull Request Checklist

### DIFF
--- a/frontend/src/components/HintButton.tsx
+++ b/frontend/src/components/HintButton.tsx
@@ -1,0 +1,35 @@
+import { useState } from 'react';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faCircleQuestion } from '@fortawesome/free-solid-svg-icons';
+
+type Props = {
+    heading: string;
+    text: string;
+    ariaLabel: string;
+};
+
+export default function HintButton({ heading, text, ariaLabel }: Readonly<Props>) {
+    const [open, setOpen] = useState(false);
+
+    return (
+        <div className="relative">
+            <button
+                type="button"
+                aria-label={ariaLabel}
+                title={`View ${heading.toLowerCase()}`}
+                className="text-white hover:text-blue-300 transition-colors"
+                onClick={() => setOpen(!open)}
+            >
+                <FontAwesomeIcon icon={faCircleQuestion} />
+            </button>
+            {open && (
+                <div
+                    className="absolute top-full right-0 mt-1 bg-sky-900 border border-sky-700 rounded-lg shadow-lg p-4 z-50 w-[400px] text-sm"
+                >
+                    <h3 className="font-bold text-white mb-3">{heading}</h3>
+                    <div className="text-gray-100">{text}</div>
+                </div>
+            )}
+        </div>
+    );
+}
--- a/frontend/src/components/RefreshVulnerabilityData.tsx
+++ b/frontend/src/components/RefreshVulnerabilityData.tsx
@@ -1,6 +1,7 @@
 import { useEffect, useRef, useState } from "react";
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faArrowsRotate, faChevronDown } from '@fortawesome/free-solid-svg-icons';
+import HintButton from './HintButton';
 import type { Vulnerability } from "../handlers/vulnerabilities";
 import { BulkNvdRefreshCancelHandler, BulkEpssRefreshCancelHandler, BulkGhsaRefreshCancelHandler, BulkEuvdRefreshCancelHandler } from "../handlers/bulkRefresh";
 import type { NVDProgress } from "../handlers/nvd_progress";
@@ -26,6 +27,7 @@ function RefreshVulnerabilityData({ vulnerabilities, getRefreshVulnerabilities, 
     const [refreshMenuOpen, setRefreshMenuOpen] = useState(false);
     const [refreshMode, setRefreshMode] = useState<'complete' | 'custom'>('complete');
     const [selectedRefreshTypes, setSelectedRefreshTypes] = useState<Set<RefreshType>>(new Set(['nvd', 'epss', 'ghsa', 'euvd']));
+    const [nvdRefreshMode, setNvdRefreshMode] = useState<'local' | 'api'>('local');
     const [cancelling, setCancelling] = useState<Set<RefreshType>>(new Set());
     const refreshMenuRef = useRef<HTMLDivElement>(null);
     const cveIds = vulnerabilities.map(vulnerability => vulnerability.id).filter(id => id.toUpperCase().startsWith('CVE-'));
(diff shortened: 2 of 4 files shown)


ValentinBoudevinSFL (changes requested):
Image

Would be nice to add it too for CPE, PURL, SUPPLIER, ACTION (explain it adds a filter in "packages" and redirect to Vulnerabilities tab). 

image

Could be cool to have more tooltips in Scan tab to explain the different tab of scanners and the difference between NVD API and NVD GIT

Also add one for Vulnerability tab "Refresh Vulnerability Data"

image

reh-sfl:
@ValentinBoudevinSFL I added tooltips to the SBOM table.

I don't agree that we should add tooltips everywhere if we can simply have a Book icon that redirects to extensive documentation. I don't like the idea of taking the user by hand on each small feature.

If the concern is that the user is not going to think about the existance of a docs or something like that, I would suggest adding a Welcome popup in the first fontend app run (that can be controller with local storage as well, so it shouldn't be too complex).

ValentinBoudevinSFL:
> Could be cool to have more tooltips in Scan tab to explain the different tab of scanners and the difference between NVD API and NVD GIT
> 
> Also add one for Vulnerability tab "Refresh Vulnerability Data"

Not for every feautres.

Just the difference between NVD API and NVD GIT is not obvious to understand and I guess it deserve one explanation in Scan tab and "Refresh Vulnerability Data"

frontend/src/pages/TablePackages.tsx
```diff
@@ -413,6 +437,14 @@ function TablePackages({ packages, vulnerabilities = [], onShowVulns, onLoadOutd
             columnHelper.accessor('source', {
                 id: 'source',
                 header: () => <div className="flex items-center justify-center">Sources</div>,
+                HintText: <>
+                    <h3 className="font-bold text-white mb-2">Sources</h3>
+                    <div className="space-y-1 text-gray-100">
+                        <p>Lists the SBOM sources that reported this package.</p>
```
ValentinBoudevinSFL, line 427: WARNING: This describes the column as SBOM-only, but `Packages.enrich_with_vulns()` replaces the displayed `source` value with the union of the package's `sources` and every matched vulnerability's `found_by` values. A package can therefore show Grype, NVD, OSV, or other vulnerability sources even when those sources did not report the package, so the hint misleads users about provenance. Either describe the column as the combined package/vulnerability sources or stop merging `found_by` into this field and expose the two provenance sets separately.

ValentinBoudevinSFL (dismissed):
The tooltip additions are generally consistent with the existing table hint mechanism, but the SBOM Sources hint incorrectly describes a column that also contains vulnerability-discovery sources.

ValentinBoudevinSFL (changes requested):
Dont see any new tooltips in Scan or Refresh Vulnerability Data

frontend/src/components/HintButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+    heading: string;
+    text: string;
+    ariaLabel: string;
+};
+
+export default function HintButton({ heading, text, ariaLabel }: Readonly<Props>) {
+    const [open, setOpen] = useState(false);
+
+    return (
+        <div className="relative">
```
ValentinBoudevinSFL, line 15: WARNING: This button opens a persistent popover without exposing its expanded state or relationship to the content, and keyboard users cannot dismiss it with Escape; each hint must be toggled closed manually and multiple hints can remain open. Implement this as an accessible disclosure/popover (for example, `aria-expanded` plus `aria-controls`, an identified popup, Escape handling, and focus/outside-click dismissal) so assistive-technology and keyboard users can operate it predictably.

ValentinBoudevinSFL (changes requested):
The hints introduce an accessibility defect in the shared popover component: opened content is not programmatically associated with its trigger and lacks standard keyboard dismissal.

ValentinBoudevinSFL (dismissed):

frontend/src/pages/TablePackages.tsx
sfl-pe-sync: WARNING: Agreed. The current enrichment path explicitly unions each package's SBOM `sources` with matched vulnerabilities' `found_by`, so this hint cannot accurately call the displayed values SBOM sources. The copy should describe both provenance sets, or the UI/data model should expose them separately.

sfl-pe-sync (changes requested):
The tooltip content and interaction need correction before merge: package provenance is described inaccurately, and the new persistent popover lacks required keyboard and assistive-technology behavior.

frontend/src/components/HintButton.tsx
sfl-pe-sync: WARNING: Agreed. Because this is a click-persistent popup rather than a transient tooltip, the trigger should expose `aria-expanded`/`aria-controls`, the popup needs an ID and appropriate semantics, and Escape plus focus/outside-click dismissal should be implemented and covered by keyboard tests.

sfl-zwang:
> @ValentinBoudevinSFL I added tooltips to the SBOM table.
> 
> I don't agree that we should add tooltips everywhere if we can simply have a Book icon that redirects to extensive documentation. I don't like the idea of taking the user by hand on each small feature.
> 
> If the concern is that the user is not going to think about the existance of a docs or something like that, I would suggest adding a Welcome popup in the first fontend app run (that can be controller with local storage as well, so it shouldn't be too complex).

I agree with @reh-sfl. We should not add too many tooltips. I would prefer to leave the tutorials in the vulnscout documentation site.

ValentinBoudevinSFL (changes requested):
As we discussed, I guess too much tooltips is not intuitive

As guess in SBOM we can keep CPE/PURL/Supplier/Actions and remove "Sources"
Vulnerability let's remove "status" and "sources"

Everything else LGTM

frontend/src/pages/TablePackages.tsx
```diff
@@ -397,6 +421,14 @@ function TablePackages({ packages, vulnerabilities = emptyVulnerabilities, onSho
...
```
sfl-pe-sync, line 427: WARNING: This hint says the column contains SBOM sources, but `Packages.enrich_with_vulns()` replaces `source` with the union of the package's `sources` and matched vulnerabilities' `found_by` values. The table can therefore display NVD, OSV, Grype, or other vulnerability sources that did not report the package, making the provenance guidance incorrect. Describe the column as combined package/vulnerability sources, or expose the two provenance sets separately.

frontend/src/components/RefreshVulnerabilityData.tsx
```diff
@@ -213,6 +221,23 @@ function RefreshVulnerabilityData({ vulnerabilities, getRefreshVulnerabilities,
                                     >{cancelling.has(source) ? 'Cancelling…' : 'Cancel'}</button>
```
sfl-pe-sync, line 221: WARNING: The NVD source selector is rendered inside the existing custom-refresh source list, so it is absent during the normal `complete` mode and complete refreshes always use the default local repository. Users cannot select the REST API while refreshing all available sources. Render the NVD mode control whenever NVD is part of `refreshTypes` (including complete mode), rather than only inside the custom-source mapping.

sfl-pe-sync (approved):
The hints are generally integrated with the existing table metadata, but the package provenance copy is inaccurate and the new NVD source selector is unavailable for complete refreshes; the standalone popup also retains the unresolved keyboard and assistive-technology accessibility issue.

ValentinBoudevinSFL (approved):

PR #489, opened 2026-07-29
Save tables user options into local storage

reh-sfl:

Save tables user options into local storage:
This means each time the user does a filter, sort, match condition, it gets saved in browser local storage so that they can find the same context when they come back later.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Pick a project / variant. go to the vulns or SBOM tables then do any number of filters / sorting / match condition, then either refresh the webpage or switch project / variant and come back to the one you desire to work on.
Observe that the actions are saved then restored.

## Pull Request Checklist

### DIFF
--- a/frontend/src/handlers/localStorage.ts
+++ b/frontend/src/handlers/localStorage.ts
@@ -1,0 +1,27 @@
+import { Dispatch, SetStateAction, useEffect, useState } from 'react';
+
+function readStoredValue<T>(key: string | null, initialValue: T): T {
+    if (key === null) return initialValue;
+    try {
+        const storedValue = window.localStorage.getItem(key);
+        return storedValue === null ? initialValue : JSON.parse(storedValue) as T;
+    } catch {
+        return initialValue;
+    }
+}
+
+
+export function useLocalStorageState<T>(key: string | null, initialValue: T): [T, Dispatch<SetStateAction<T>>] {
+    const [value, setValue] = useState<T>(() => readStoredValue(key, initialValue));
+
+    useEffect(() => {
+        if (key === null) return;
+        try {
+            window.localStorage.setItem(key, JSON.stringify(value));
+        } catch {
+            // Storage can be unavailable or full; keep the in-memory state usable.
+        }
+    }, [key, value]);
+
+    return [value, setValue];
+}
--- a/frontend/src/pages/TableVulnerabilities.tsx
+++ b/frontend/src/pages/TableVulnerabilities.tsx
@@ -75,6 +75,7 @@ function useRefreshProgressEffect(
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faFilter, faCaretDown, faCircleQuestion, faSync, faCircleInfo, faBook } from '@fortawesome/free-solid-svg-icons';
 import ExplicitSearchInput from '../components/ExplicitSearchInput';
+import { useLocalStorageState } from '../handlers/localStorage';
 import RangeSlider from "../components/RangeSlider";
 
 type Props = {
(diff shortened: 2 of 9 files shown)


ValentinBoudevinSFL (dismissed):
LGTM

frontend/src/pages/TableVulnerabilities.tsx
```diff
@@ -1436,7 +1440,13 @@ function TableVulnerabilities ({ vulnerabilities, filterLabel, filterValue, filt
         } finally {
             if (!controller.signal.aborted) setDescriptionSearchLoading(false);
         }
-    }, [draftSearch, vulnerabilities, variantId, projectId]);
+    }, [draftSearch, vulnerabilities, variantId, projectId, setSearch]);
+
+    useEffect(() => {
```
sfl-pe-sync, line 1436: ERROR: The one-shot `hasRestoredSearch` guard marks restoration complete on the first render, even though Explorer loads/replaces `vulnerabilities` asynchronously (and can initially render the previous scope's rows). A persisted search containing description terms is therefore evaluated against empty or stale IDs once; when the current scope's vulnerabilities arrive, `applySearch` changes and this effect runs again but exits at the guard, leaving incorrect `descriptionMatches` and an incorrectly filtered table. Re-run the restored search when the scoped vulnerability data becomes ready/changes (while aborting the prior request), or defer setting the restored flag until the current scope's data has loaded.

sfl-pe-sync (changes requested):
The persistence implementation has a blocking restoration defect: description-based vulnerability searches can be restored against empty or previous-scope data and are not recomputed when the selected scope finishes loading.

sfl-zwang (dismissed):

frontend/src/pages/TableVulnerabilities.tsx
```diff
@@ -1416,7 +1419,24 @@ function TableVulnerabilities ({ vulnerabilities, filterLabel, filterValue, filt
...
+    const applySearchRef = useRef(applySearch);
+    applySearchRef.current = applySearch;
+
+    // Re-run the persisted/restored search whenever the scoped vulnerability
+    // data changes. Explorer loads vulnerabilities asynchronously and may
+    // initially render the previous scope's rows, so restoring only once on
+    // mount would evaluate the search against stale or empty IDs and leave an
+    // incorrectly filtered table. Re-running on data changes keeps
+    // descriptionMatches consistent with the current scope; applySearch aborts
+    // any in-flight description request before starting a new one. Depending on
```
sfl-pe-sync, line 1433: WARNING: This restoration effect applies `draftSearch`, so text that the user typed but deliberately did not submit becomes active after a refresh. For example, after applying `foo`, editing the input to `bar`, and refreshing, the stored applied search is still `foo` but this effect calls `applySearch`, overwrites it with `bar`, and filters on the uncommitted draft. Re-evaluate the persisted `search` value on data changes (for example via a helper that accepts the query), while keeping `draftSearch` only for explicit button/Enter submissions.

frontend/src/handlers/localStorage.ts
```diff
@@ -0,0 +1,27 @@
+import { Dispatch, SetStateAction, useEffect, useState } from 'react';
+
+function readStoredValue<T>(key: string | null, initialValue: T): T {
+    if (key === null) return initialValue;
+    try {
+        const storedValue = window.localStorage.getItem(key);
+        return storedValue === null ? initialValue : JSON.parse(storedValue) as T;
+    } catch {
```
sfl-pe-sync, line 8: WARNING: Successfully parsed JSON is asserted to `T` without validating its shape, so stale or malformed-but-valid storage can crash a table during render; for example, a stored `null` pagination value reaches `pagination.pageIndex`, and a stored non-array filter reaches `.includes`. Accept a runtime validator/decoder per preference (or validate each persisted state before use), remove invalid entries, and fall back to the initial value when the decoded shape is incompatible.

sfl-pe-sync (changes requested):
The persistence feature has two correctness gaps: restoration can apply an unsubmitted vulnerability-search draft, and structurally invalid stored values can crash table rendering.

PR #490, opened 2026-07-30
Select scans to perform after the upload of an SBOM

reh-sfl:

* Select scans to perform after the upload of an SBOM

### Status

- [ ] READY
- [ ] HOLD
- [X] WIP (Work-In-Progress)

### How to verify this change

Open the web dashboard, go to settings tab then upload an SBOM, you will be able top pick which scans you want to perform once it's processed.

## Pull Request Checklist

### DIFF
--- a/src/routes/settings.py
+++ b/src/routes/settings.py
@@ -51,6 +51,7 @@ class _CrudController(Protocol[_C]):
 # Tracks in-progress SBOM uploads: upload_id → {status, message, ts}
 _upload_status: dict[str, dict] = {}
 _UPLOAD_STATUS_TTL = 3600  # seconds – entries older than this are pruned
+_REFRESH_SOURCES = {"epss", "nvd", "euvd", "ghsa"}
 
 
 def _prune_upload_status() -> None:
@@ -167,9 +168,11 @@ def _extract_spdx_archive(archive_path: str, filename: str) -> list[tuple[str, s
 
 def _process_sbom_background(
     app: Flask, upload_id: str, file_paths: list[str],
-    scan_id: uuid.UUID, variant_id: uuid.UUID,
+    scan_id: uuid.UUID, variant_id: uuid.UUID, refresh_sources: set[str] | None = None,
 ) -> None:
     """Run SBOM parsing in a background thread for one or more files."""
+    if refresh_sources is None:
+        refresh_sources = {"epss"}
     with app.app_context():
         try:
             _upload_status[upload_id] = {"status": "processing", "message": "Parsing SBOM file(s)..."}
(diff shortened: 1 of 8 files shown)


src/routes/settings.py
```diff
@@ -1102,6 +1126,16 @@ def upload_sbom() -> ResponseReturnValue:
 
         project_id = request.form.get('project_id', '').strip()
         variant_id = request.form.get('variant_id', '').strip()
+        requested_refresh_sources = request.form.getlist('refresh_sources')
+        refresh_sources = {
+            source.strip().lower() for source in requested_refresh_sources if source.strip()
+        } if requested_refresh_sources else {"epss"}
+
```
ValentinBoudevinSFL, line 1154: ERROR: When the user unchecks every source, the frontend appends no `refresh_sources` fields, so `getlist()` returns an empty list and this fallback silently re-enables EPSS. The UI therefore cannot request “no refresh” and performs work the user explicitly deselected. Add an explicit presence/sentinel field (or another representation that distinguishes an empty selection from a legacy request that omitted the parameter) and cover the all-unchecked case end to end.

src/routes/settings.py
```diff
@@ -194,12 +195,35 @@ def _process_sbom_background(
                 else ScanModel.get_by_id(uuid.UUID(str(scan_id)))
             populate_observations(scan, vulnCtrl, log_prefix="settings/upload")
 
-            # Run EPSS enrichment
+            encountered = vulnCtrl._encountered_this_run
+            all_vulnerabilities = vulnCtrl.vulnerabilities
             try:
```
ValentinBoudevinSFL, line 203: ERROR: This `try` wraps the entire provider loop, so an exception from an earlier selected provider prevents every later provider from running; for example, an EPSS HTTP failure skips selected NVD/EUVD/GHSA refreshes and the upload is still reported as `done`. Isolate error handling per source and continue with the remaining selections, while reporting partial failures accurately.

src/routes/settings.py
```diff
@@ -167,7 +168,7 @@ def _extract_spdx_archive(archive_path: str, filename: str) -> list[tuple[str, s
 
 def _process_sbom_background(
     app: Flask, upload_id: str, file_paths: list[str],
```
ValentinBoudevinSFL, line 170: ERROR: Making `refresh_sources` a required positional argument leaves the existing direct callers in `TestProcessSBOMBackground` unchanged. The current backend CI run fails four tests with this contract change, so the branch cannot pass the required gate. Update all callers/tests to pass the intended set, or provide a backward-compatible default that preserves the previous EPSS behavior.

.github/skills/do-vulnscout-release/SKILL.md
```diff
@@ -0,0 +1,122 @@
...
+- `doc/source/conf.py`
+- `doc/source/container-entrypoint.md`
+- `doc/source/vulnscout-script.md`
+- `frontend/package-lock.json`
+- `frontend/package.json`
+
+For JSON files, use a JSON-aware edit and keep the lockfile's package version fields consistent with the package manifest. Confirm that the old version is not unintentionally left in release metadata. Review the diff, then commit only these version updates:
+
+```sh
+git add Dockerfile README.adoc doc/source/conf.py \
```
ValentinBoudevinSFL: WARNING: The procedure says to bump `vulnscout`, but this staging command omits it, so the launcher’s image version remains outside the version commit and the subsequent clean-worktree completion check cannot succeed. Add `vulnscout` to this `git add` command and update the completion check’s “seven files” count to eight.

ValentinBoudevinSFL (changes requested):
The scan-selection implementation is not ready to merge: an empty selection still runs EPSS, one provider failure silently skips later selections, and the required-argument change currently breaks the backend test gate. The added release skill also omits the launcher from its version commit.

ValentinBoudevinSFL (changes requested):
LGTM but should be merge after the queue system. 

Also, should also concern the Scan features in scan tab (e.g sbom-cve-check)

ValentinBoudevinSFL (dismissed):
No actionable defects found.

src/routes/settings.py
```diff
@@ -194,16 +197,48 @@ def _process_sbom_background(
...
+                        vuln_id: all_vulnerabilities[vuln_id]
+                        for vuln_id in encountered
+                        if vuln_id.startswith(prefix) and vuln_id in all_vulnerabilities
+                    }
+                    if not vulnCtrl.vulnerabilities:
+                        continue
+                    label = source.upper()
+                    _upload_status[upload_id] = {
+                        "status": "processing",
+                        "message": f"Refreshing {label} data...",
```
sfl-pe-sync, line 218: ERROR: This aggregation assumes each provider raises when its refresh fails, but `fetch_nvd_data()` explicitly swallows per-vulnerability failures and `fetch_euvd_data()` catches per-record and final-commit failures; EUVD download failures can also degrade to an empty mapping. A total NVD, GHSA, or EUVD failure therefore leaves `failed_sources` empty and reports “SBOM imported successfully,” even though the selected refresh did not complete. Have providers return structured success/failure results (or raise when the source could not be refreshed), aggregate those results here, and test the providers' real failure contracts rather than only mocks that raise at the top level.

sfl-pe-sync (changes requested):
The selection and backward-compatible sentinel handling are implemented, but refresh failures can still be silently reported as success because the provider methods suppress the errors this new aggregation relies on.

src/routes/settings.py
sfl-pe-sync: Resolved in this patchset: the frontend now sends an explicit `none` sentinel and the backend distinguishes it from an omitted field while preserving the legacy EPSS default.

src/routes/settings.py
sfl-pe-sync: Resolved in this patchset: `refresh_sources` now has a backward-compatible `None` default that preserves EPSS behavior for existing direct callers, with coverage for the no-argument path.

.github/skills/do-vulnscout-release/SKILL.md
sfl-pe-sync: Resolved by removing the unrelated release-skill changes from this patchset.

sfl-pe-sync (approved):
No actionable defects found.

src/routes/settings.py
sfl-pe-sync: Resolved: each selected source now has its own exception boundary, so a failure is recorded while later sources continue running.

ValentinBoudevinSFL (approved):

PR #491, opened 2026-07-31
Feature/configurable upload limit

Emmmery:

* Since v0.19, the web API upload size limit is hardcoded at ~5MB, which prevents
  larger uploads from going through.
* This PR makes the limit configurable via the new `VULNSCOUT_MAX_UPLOAD_MB`
  environment variable (value in MB, default `500`).
* Adds a small typed `get_int_env()` helper in `src/helpers/env_vars.py` to
  read integer environment variables safely.
* Documents the new variable in `README.adoc` and
  `doc/source/vulnscout-script.md`.

### Status

- [ ] READY
- [ ] HOLD
- [x] WIP (Work-In-Progress)

### How to verify this change

1. Set a small limit and restart in dev mode (so the local `src/` runs):
   ```bash
   ./vulnscout config VULNSCOUT_MAX_UPLOAD_MB 30
   ./vulnscout stop && ./vulnscout --dev --serve
   ```
2. Check the value reached the container:
   ```bash
   docker exec vulnscout printenv VULNSCOUT_MAX_UPLOAD_MB   # -> 30
   ```
3. Upload a file above the configured limit → the request is rejected with an
   HTTP 413 (Payload Too Large) error; below the limit → it succeeds.
4. Removing the variable (or setting `500`) restores the previous default.

### Additional notes

The default stays at 500 MB — maybe the default limit should be lower? 
## Pull Request Checklist

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -417,6 +417,10 @@ The following environment variables can be set via `vulnscout config` or exporte
 |`USER_GID`
 |GID used to write output files
 |current group
+
+|`VULNSCOUT_MAX_UPLOAD_MB`
+|Maximum upload/request body size accepted by the web API, in MB
+|`500`
 |===
 
 .Scan & enrichment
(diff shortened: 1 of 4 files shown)


ValentinBoudevinSFL (changes requested):
@reh-sfl IMO should be also a parameters present in Settings tab. Do you agree?

reh-sfl:
> @reh-sfl IMO should be also a parameters present in Settings tab. Do you agree?

Correct, we use `config.py` which handles custom user settings that can be adjusted both through the `config.env` file and in the web dashboard through Settings.

PR #493, opened 2026-08-03
fix(import): reject foreign variant_id then fall back to variant name

reh-sfl:

A variant_id from another VulnScout instance's export never matches a variant in this database since each instance mints its own UUIDs. Trusting it as-is silently attached the assessment to a variant_id no local query would ever match, so it disappeared from the Review page while remaining in the DB.

Now the import falls back to the human-readable variant name when the variant_id is unknown locally, and reports an error if neither resolves to an existing variant.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Import a Vulnscout JSON into a new instance where the exported variant does not exist. You should get an error indicating that some of the variants are absent from your current DB, you can create those in the Settings tab then try again.

### Related Issue

#474

## Pull Request Checklist

### DIFF
--- a/doc/source/api.md
+++ b/doc/source/api.md
@@ -630,7 +630,9 @@ Accepts either:
 - `multipart/form-data` with a `file` field containing a `.json` file.
 - `application/json` body with the custom-data payload directly.
 
-The records keep the variant identifiers embedded in the file.
+The JSON body must include the destination `project_id` UUID. Variant IDs in
+the file are resolved only within that project; a foreign ID falls back to its
+variant name in the selected project.
 
 **Response:**
 ```json
--- a/src/helpers/assessment_io.py
+++ b/src/helpers/assessment_io.py
@@ -601,6 +601,8 @@ def import_custom_data(
         ).all()
         return [variant_id for (variant_id,) in rows]
 
+    known_variant_ids = {v.id for v in variant_by_name.values()}
+
     def _resolve_variant(raw_item: dict) -> "_uuid.UUID | None":
         if variant_id is not None:
             return variant_id
@@ -613,12 +615,25 @@ def import_custom_data(
             return None
 
         try:
-            return _uuid.UUID(str(variant_token))
+            parsed = _uuid.UUID(str(variant_token))
         except (ValueError, TypeError):
-            mapped_variant = variant_by_name.get(str(variant_token))
-            if mapped_variant is None:
-                return None
-            return mapped_variant.id
+            parsed = None
+
+        # A variant_id copied from another VulnScout instance's export never
+        # matches a variant in this database (each instance generates its own
+        # UUIDs), so it must not be trusted as-is: doing so silently attaches
+        # the record to a variant_id that no local query will ever match.
+        # Fall back to the human-readable name in that case.
+        if parsed is not None and parsed in known_variant_ids:
+            return parsed
+
+        name_token = raw_item.get("variant")
+        if name_token in (None, ""):
+            name_token = variant_token
+        mapped_variant = variant_by_name.get(str(name_token))
+        if mapped_variant is None:
+            return None
+        return mapped_variant.id
 
     def _import_assessments(
         key: str,
@@ -650,6 +665,16 @@ def import_custom_data(
 
             # Determine which variant to attach to
             target_variant_id = _resolve_variant(a)
+
+            variant_token = a.get("variant_id")
+            if variant_token in (None, ""):
+                variant_token = a.get("variant")
+            if target_variant_id is None and variant_token not in (None, ""):
+                result["errors"].append({
+                    "vuln_id": vuln_name,
+                    "error": f"Variant '{variant_token}' not found",
+                })
+                continue
 
             justification = a.get("justification", "")
             impact_statement = a.get("impact_statement", "")
--- a/src/routes/assessments.py
+++ b/src/routes/assessments.py
@@ -815,7 +815,7 @@ def init_app(app: Flask) -> None:
         Accepts either:
 
         * ``multipart/form-data`` with a ``file`` field containing a ``.json``
-          file.
+                    file and a destination ``project_id`` field.
         * ``application/json`` body with the custom-data payload directly.
 
         OpenAPI:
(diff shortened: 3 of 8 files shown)


src/helpers/assessment_io.py
```diff
@@ -613,12 +615,25 @@ def _resolve_variant(raw_item: dict) -> "_uuid.UUID | None":
...
+        # UUIDs), so it must not be trusted as-is: doing so silently attaches
+        # the record to a variant_id that no local query will ever match.
+        # Fall back to the human-readable name in that case.
+        if parsed is not None and parsed in known_variant_ids:
+            return parsed
+
+        name_token = raw_item.get("variant")
+        if name_token in (None, ""):
+            name_token = variant_token
+        mapped_variant = variant_by_name.get(str(name_token))
```
sfl-pe-sync, line 633: ERROR: The web import builds `variant_by_name` from every project, so this new fallback can attach a foreign export to an arbitrary project when common names such as `default` exist in more than one project; `build_variant_by_name_map()` overwrites duplicate keys and the imported assessment/CVSS/time estimate is then persisted under whichever variant won. Scope web imports to an explicit project, or export and resolve project identity alongside the variant name and reject ambiguous names; add a regression test with two projects sharing the same variant name.

sfl-pe-sync (changes requested):
The foreign-UUID fallback introduces a cross-project attachment path in the web importer because variant names are resolved through a global, collision-prone map. Static review only; shell validation was unavailable under the review-session constraints.

src/helpers/assessment_io.py
ValentinBoudevinSFL: Feature looks good but this feedback looks valid to me

src/routes/assessments.py
```diff
@@ -847,11 +847,24 @@ def import_review_custom_data() -> ResponseReturnValue:
         if not isinstance(data, dict) or "version" not in data:
             return {"error": "Invalid custom-data format. Expected {version, assessments, ...}"}, 400
 
+        project_id = data.get('project_id')
```
sfl-pe-sync: ERROR: Multipart imports now read `project_id` only from the uploaded JSON object, but `build_custom_data_export()` does not emit a destination project ID. Therefore the documented `multipart/form-data` workflow cannot upload an unmodified VulnScout export anymore: every such request reaches this check without `project_id` and returns 400. Accept the destination project from a separate multipart form field (or another explicit request parameter) while retaining the JSON-body path, and add a regression test that uploads an unmodified export with that field.

sfl-pe-sync (changes requested):
The project-scoped fallback resolves the prior cross-project ambiguity, but the new required field breaks the documented multipart import contract for normal exported files.

src/helpers/assessment_io.py
sfl-pe-sync: The latest patch resolves this concern: the endpoint now requires a destination `project_id`, builds the variant map only for that project, and adds duplicate-name coverage across projects.

src/routes/assessments.py
ValentinBoudevinSFL: Looks valid to me @reh-sfl

src/helpers/assessment_io.py
```diff
@@ -651,6 +666,16 @@ def _import_assessments(
             # Determine which variant to attach to
```
sfl-pe-sync, line 666: ERROR: This unresolved-variant guard is only applied inside `_import_assessments`. CVSS and time-estimate entries also call `_resolve_variant`, but an explicit foreign `variant_id` with no resolvable name still becomes `None` there and can follow their existing unscoped/inferred-variant path instead of being rejected, so metadata from a foreign variant can be attached to local variants for the same vulnerability. Make resolution distinguish "no variant supplied" from "an explicit variant failed to resolve", reject the latter consistently in all three import loops, and add CVSS/time-estimate regression cases.

doc/source/api.md
```diff
@@ -630,7 +630,9 @@ Accepts either:
 - `multipart/form-data` with a `file` field containing a `.json` file.
 - `application/json` body with the custom-data payload directly.
 
-The records keep the variant identifiers embedded in the file.
+The JSON body must include the destination `project_id` UUID. Variant IDs in
```
sfl-pe-sync, line 633: WARNING: The endpoint now requires `project_id` for multipart imports too, but this documentation says only the JSON body must include it and the multipart bullet above still lists only the `file` field. A client following the documented multipart contract will receive `400 project_id is required`; document `project_id` as a required multipart form field as well.

sfl-pe-sync (changes requested):
The project-scoped assessment fallback is improved, but unresolved foreign variant IDs are still not rejected consistently for CVSS and time-estimate records, allowing incorrect variant association. The multipart API documentation also omits its newly required project field.

src/routes/assessments.py
sfl-pe-sync: The latest patch addresses this: multipart requests now read `project_id` from `request.form`, and the new round-trip regression test uploads an unmodified export with that form field.

src/routes/assessments.py
```diff
@@ -837,21 +838,35 @@ def import_review_custom_data() -> ResponseReturnValue:
                 data = _json.load(uploaded.stream)
             except Exception:
                 return {"error": "Invalid JSON file"}, 400
+            project_id = request.form.get('project_id')
         elif request.content_type and 'application/json' in request.content_type:
             data = request.get_json(silent=True)
             if data is None:
```
sfl-pe-sync, line 844: WARNING: `request.get_json()` may return any JSON type, so a valid body such as `[]` reaches `data.get('project_id')` and raises `AttributeError` before the existing custom-data format check, turning malformed client input into a 500 response. Check `isinstance(data, dict)` before reading `project_id` (or move the format validation ahead of this access) and add a non-object JSON regression test expecting 400.

sfl-pe-sync (approved):
The project-scoped fallback is improved, but the new JSON project lookup regresses validation for non-object JSON bodies by raising before the format check.

ValentinBoudevinSFL (approved):

PR #496, opened 2026-08-04
Remove variants filter and unscope assessment history

reh-sfl:

* Remove variants filter in vulnerabilities table
*  Unscope assessment history in vulnmodal

### Status

- [ ] READY
- [ ] HOLD
- [X] WIP (Work-In-Progress)

### How to verify this change

1. Open the web dashboard, go to vulnerabilities table, observe there is no variants filter anymore.
2. Pick a vuln, then observe that you are able to work with all variants from the current project

## Pull Request Checklist

### DIFF
--- a/frontend/src/components/VulnModal.tsx
+++ b/frontend/src/components/VulnModal.tsx
@@ -184,12 +184,13 @@ type VariantScopedSnapshot = {
         return () => controller.abort();
     }, [vuln.id, projectId]);
 
-    // Fetch ALL assessments for this vuln (unfiltered) so variant tags are
-    // complete even when a variant filter is active in the explorer.
+    // Fetch all assessments for this vulnerability in the current project so
+    // its history remains complete when the Explorer is variant-scoped.
     useEffect(() => {
         const controller = new AbortController();
         setAllVulnAssessments([]);
-        fetch(import.meta.env.VITE_API_URL + `/api/vulnerabilities/${encodeURIComponent(vuln.id)}/assessments`, { mode: 'cors', signal: controller.signal })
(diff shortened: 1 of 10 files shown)


ValentinBoudevinSFL (changes requested):
Variants frontend filter should be removed from both Review and SBOM tab too. Also, commit message is empty

frontend/src/components/VulnModal.tsx
```diff
@@ -911,7 +912,13 @@ type VariantScopedSnapshot = {
             .sort((a, b) => new Date(b.timestamp).getTime() - new Date(a.timestamp).getTime());
     };
 
-    const groupedAssessments = groupAssessments(vuln.assessments);
+    // Prefer the project-wide response so history includes every variant. The
+    // scoped vulnerability rows remain a fallback for unavailable or empty
```
sfl-pe-sync, line 916: ERROR: This now treats `allVulnAssessments` as the history even when it was fetched without a project scope. `Review.tsx` receives `projectId` but invokes `VulnModal` without passing it, so that call site requests `/api/vulnerabilities/{id}/assessments` with no `project_id` and renders assessments from every project's variants. Pass `projectId` from every call site (preferably make the prop required), or only use this response when it is keyed to the current project; add a Review-page regression test with the same CVE assessed in two projects.

sfl-pe-sync (changes requested):
The project-scoped endpoint is sound, but the modal's new history source creates a blocking cross-project assessment leak from the Review page because that existing call site does not pass its available project ID.

sfl-pe-sync (approved):
No actionable defects found.

ValentinBoudevinSFL (approved):

PR #498, opened 2026-08-04
Settings: redesign and change workflow

reh-sfl:

* Redesign the serttings page to have a better view of the projects / variants context.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Open the web dashboard, go to Settings then explore the new view.

## Pull Request Checklist

### DIFF
--- a/frontend/src/pages/Settings.tsx
+++ b/frontend/src/pages/Settings.tsx
@@ -12,6 +12,10 @@ import {
   faXmark,
   faPenToSquare,
   faBug,
+  faChevronDown,
+  faChevronRight,
+  faFolder,
+  faGear,
 } from "@fortawesome/free-solid-svg-icons";
 import Projects from "../handlers/project";
 import type { Project } from "../handlers/project";
@@ -30,62 +34,11 @@ type Props = {
   onLoadingMessage?: (message: string | null) => void;
 };
 
-type SettingsTab = "general" | "projects" | "variants" | "scan";
-
-type OutdatedPackagePlan = {
-  package: string;
-  vulnerabilities: string[];
-  assessments: string[];
-  linkedData: { observations: number; sbomPackages: number; sbomObservations: number };
-};
-
-type OutdatedVariantPlan = { name: string; packages: Map<string, OutdatedPackagePlan> };
-type OutdatedProjectPlan = { name: string; variants: Map<string, OutdatedVariantPlan> };
+type SettingsTab = "general" | "projects" | "variants";
+type FeedbackMsg = { text: string; type: "success" | "error" } | null;
 type AdditionalCleanup =
   | { kind: "empty-scans"; scans: EmptyScanPreview[] }
   | { kind: "orphaned-vulnerabilities"; vulnerabilities: OrphanedVulnerabilityPreview[] };
-
-function buildOutdatedDataPlan(preview: OutdatedDataPreview): OutdatedProjectPlan[] {
-  const projects = new Map<string, OutdatedProjectPlan>();
-  const ensurePackage = (projectName: string, variantName: string, packageName: string) => {
-    let project = projects.get(projectName);
-    if (!project) {
-      project = { name: projectName, variants: new Map() };
-      projects.set(projectName, project);
-    }
-    let variant = project.variants.get(variantName);
-    if (!variant) {
-      variant = { name: variantName, packages: new Map() };
-      project.variants.set(variantName, variant);
-    }
-    let packagePlan = variant.packages.get(packageName);
-    if (!packagePlan) {
-      packagePlan = {
-        package: packageName,
-        vulnerabilities: [],
-        assessments: [],
-        linkedData: { observations: 0, sbomPackages: 0, sbomObservations: 0 },
-      };
-      variant.packages.set(packageName, packagePlan);
-    }
-    return packagePlan;
-  };
-
-  for (const item of preview.packages) {
-    const packagePlan = ensurePackage(item.project, item.variant, item.package);
-    packagePlan.vulnerabilities.push(...item.vulnerabilities);
-    packagePlan.linkedData = {
-      observations: item.linked_data.observations,
-      sbomPackages: item.linked_data.sbom_packages,
-      sbomObservations: item.linked_data.sbom_observations,
-    };
-  }
-  for (const item of preview.assessments) {
-    const packagePlan = ensurePackage(item.project, item.variant, item.package);
-    packagePlan.assessments.push(item.vulnerability);
-  }
-  return [...projects.values()];
-}
 
 function Settings({ onDataChanged, onLoadingMessage }: Readonly<Props>) {
   // ---- Active category tab ----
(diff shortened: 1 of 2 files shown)


ValentinBoudevinSFL (approved):
LGTM

ValentinBoudevinSFL (changes requested):
Missing tests but changes LGTM

frontend/src/pages/Settings.tsx
```diff
@@ -138,6 +145,80 @@ function Settings({ onDataChanged, onLoadingMessage }: Readonly<Props>) {
...
+  };
+
+  const startNewProject = () => {
+    clearProjectMsgs();
+    setRenameProjectId("");
+    setRenameProjectName("");
+    setActiveTab("projects");
+  };
+
+  const selectVariantFromTree = (projectId: string, variant: Variant) => {
```
sfl-pe-sync, line 218: ERROR: Switching the selected project/variant does not clear `importFiles`, so files queued for variant A remain selected and `handleUploadSBOM` can upload them into variant B after the user changes the tree selection. Reset the variant-scoped import state whenever the target changes (including the new-variant flow), or bind the queued files to the target IDs and reject mismatches before upload.

frontend/src/pages/Settings.tsx
```diff
@@ -320,65 +410,56 @@ function Settings({ onDataChanged, onLoadingMessage }: Readonly<Props>) {
   const handleCreateVariant = async () => {
     if (!newVariantName.trim() || !variantProjectId) return;
     setCreateVariantBusy(true);
-    setVariantMsg(null);
+    setAddVariantMsg(null);
     try {
-      await Variants.create(variantProjectId, newVariantName.trim());
+      const created = await Variants.create(variantProjectId, newVariantName.trim());
```
sfl-pe-sync: ERROR: This async create flow reads `variantProjectId` again after `await`. If the user starts creation under project A and selects project B before the request resolves, the returned variant belongs to A while the form retains B as its project; a subsequent SBOM upload sends mismatched project/variant IDs. Snapshot the originating project ID before awaiting and apply/reload the result only for that target (or prevent target changes while the operation is pending).

frontend/src/pages/Settings.tsx
```diff
@@ -204,23 +285,26 @@ function Settings({ onDataChanged, onLoadingMessage }: Readonly<Props>) {
...
+  const [deleteProjectMsg, setDeleteProjectMsg] = useState<FeedbackMsg>(null);
 
   const handleRenameProject = async () => {
     if (!renameProjectId || !renameProjectName.trim()) return;
     setRenameProjectBusy(true);
-    setProjectMsg(null);
+    setRenameProjectMsg(null);
     try {
-      await Projects.rename(renameProjectId, renameProjectName.trim());
+      const updated = await Projects.rename(renameProjectId, renameProjectName.trim());
```
sfl-pe-sync, line 436: WARNING: A late project-rename response unconditionally writes its name into the currently displayed form. If the user selects another project while the request is pending, that other project receives the first project's name in the input and can then be renamed accidentally. Capture the target ID and ignore stale responses, or disable tree navigation until the rename finishes; apply the same guard to the variant rename flow.

frontend/src/pages/Settings.tsx
```diff
@@ -138,6 +145,80 @@ function Settings({ onDataChanged, onLoadingMessage }: Readonly<Props>) {
     loadProjects();
   }, [loadProjects]);
 
+  useEffect(() => {
+    let cancelled = false;
+
+    Promise.all(projects.map(async (project) => [project.id, await Variants.list(project.id)] as const))
```
sfl-pe-sync, line 172: WARNING: The full-tree variant request replaces the entire `projectVariants` map without guarding against stale responses. A slower load started before a create/rename/delete refresh can resolve afterward and overwrite the newer per-project list, leaving the sidebar stale. Add request generation/abort handling so an older tree load cannot commit over mutation refreshes.

sfl-pe-sync (changes requested):
The redesigned settings workflow introduces two blocking target-consistency defects that can upload an SBOM to the wrong variant/project context, plus stale async responses can overwrite the active form and refreshed tree data.

frontend/src/pages/Settings.tsx
ValentinBoudevinSFL: Looks valid to me

frontend/src/pages/Settings.tsx
ValentinBoudevinSFL: Looks valid to me

frontend/src/pages/Settings.tsx
```diff
@@ -138,6 +145,82 @@ function Settings({ onDataChanged, onLoadingMessage }: Readonly<Props>) {
...
+  };
+
+  const startNewProject = () => {
+    clearProjectMsgs();
+    setRenameProjectId("");
+    setRenameProjectName("");
+    setActiveTab("projects");
+  };
+
+  const selectVariantFromTree = (projectId: string, variant: Variant) => {
```
sfl-pe-sync, line 218: ERROR: Changing the selected project or variant leaves `importFiles` queued, so files chosen for variant A can be uploaded into variant B after navigating the tree. Clear all variant-scoped import state whenever the target changes (including the new-variant flow), or bind queued files to the target IDs and reject an upload when they no longer match.

frontend/src/pages/Settings.tsx
```diff
@@ -204,23 +287,26 @@ function Settings({ onDataChanged, onLoadingMessage }: Readonly<Props>) {
...
+  const [deleteProjectMsg, setDeleteProjectMsg] = useState<FeedbackMsg>(null);
 
   const handleRenameProject = async () => {
     if (!renameProjectId || !renameProjectName.trim()) return;
     setRenameProjectBusy(true);
-    setProjectMsg(null);
+    setRenameProjectMsg(null);
     try {
-      await Projects.rename(renameProjectId, renameProjectName.trim());
+      const updated = await Projects.rename(renameProjectId, renameProjectName.trim());
```
sfl-pe-sync, line 436: WARNING: The rename response unconditionally updates the currently displayed form after the request resolves. If the user selects another node while a project or variant rename is pending, the new selection receives the previous node's name and can then be renamed accidentally. Capture the target ID and ignore stale responses, or prevent navigation until the mutation completes; guard both project and variant rename paths.

frontend/src/pages/Settings.tsx
```diff
@@ -138,6 +145,82 @@ function Settings({ onDataChanged, onLoadingMessage }: Readonly<Props>) {
     loadProjects();
   }, [loadProjects]);
 
+  useEffect(() => {
+    let cancelled = false;
+
+    Promise.all(projects.map(async (project) => [project.id, await Variants.list(project.id)] as const))
```
sfl-pe-sync, line 172: WARNING: The full-tree variant request replaces the entire `projectVariants` map without stale-response protection. A slower initial load can resolve after a create, rename, or delete refresh and overwrite the newer per-project list, leaving the sidebar inconsistent with the server. Abort superseded requests or use a request generation so older loads cannot commit over mutation refreshes.

sfl-pe-sync (changes requested):
The settings redesign has a blocking target-switch bug that can upload a queued SBOM into the wrong variant, plus stale async responses can overwrite newer selection and tree state.

frontend/src/pages/Settings.tsx
reh-sfl: This is irrelevant as user can decide to change where the files needs to be targeted.

PR #499, opened 2026-08-06
Add CPE to the synced NVD data 

reh-sfl:

* Add CPE as part of the vulnerability metadata obtained through the NVD refresh feature

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Run a vulnerabilities refresh with NVD selected, observe that CPEs now appear in VulnModal as a collapsible metadata. 

## Pull Request Checklist

### DIFF
--- a/src/controllers/nvd_extract.py
+++ b/src/controllers/nvd_extract.py
@@ -12,6 +12,36 @@ from __future__ import annotations
 import datetime
 
 
+def extract_affected_cpes(cve: dict) -> list[str]:
+    """Return deduplicated vulnerable CPE criteria from NVD configurations."""
+    configurations = cve.get("configurations", [])
+    if isinstance(configurations, dict):
+        configurations = [configurations]
+    if not isinstance(configurations, list):
+        return []
+
+    affected_cpes: list[str] = []
+
+    def visit_nodes(nodes: object) -> None:
+        if not isinstance(nodes, list):
+            return
+        for node in nodes:
+            if not isinstance(node, dict):
+                continue
+            for match in node.get("cpeMatch", []):
+                if not isinstance(match, dict) or not match.get("vulnerable"):
+                    continue
+                criteria = match.get("criteria")
+                if isinstance(criteria, str) and criteria and criteria not in affected_cpes:
+                    affected_cpes.append(criteria)
+            visit_nodes(node.get("nodes", []))
+
+    for configuration in configurations:
+        if isinstance(configuration, dict):
+            visit_nodes(configuration.get("nodes", []))
+    return affected_cpes
+
+
 def extract_cve_details(cve: dict) -> dict:
     """Extract description, severity, links, weaknesses from an NVD CVE dict.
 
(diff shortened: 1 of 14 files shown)


src/migrations/versions/u7d8e9f0a1b2_ensure_codebase_path_column.py
```diff
@@ -0,0 +1,33 @@
...
+Revises: t6c7d8e9f0a1
+Create Date: 2026-08-04 00:00:00.000000
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+revision = 'u7d8e9f0a1b2'
+down_revision = 't6c7d8e9f0a1'
```
sfl-pe-sync, line 17: ERROR: This reuses the already-published `u7d8e9f0a1b2` revision ID for different migration code. Any deployed database stamped at the original deduplication revision is therefore considered to have run this new repair migration, so Alembic skips `upgrade()` and the missing `variant_context.codebase_path` column remains missing; later ORM access can still fail after upgrading to the new head. Keep the existing `u7d8e9f0a1b2` migration and revision unchanged, assign this repair a new revision after it, and make the CPE migration depend on that new revision.

sfl-pe-sync (changes requested):
The CPE extraction and display path is incomplete, and the migration chain contains a blocking revision-ID reuse that prevents the intended schema repair on already-upgraded databases.

---
- `src/controllers/nvd_extract.py:150` [warning]: WARNING: The web/CLI NVD scan path also consumes `extract_cve_details()` through `persist_nvd_cve()`, but that persistence helper does not pass `details["cpes"]` when creating a vulnerability or update it on an existing one. Consequently CPEs obtained during those NVD syncs are silently discarded and only a later refresh can populate them. Wire `cpes` into both persistence branches and add coverage for new and existing records.

sfl-pe-sync (dismissed):
No actionable defects found.

src/controllers/nvd_extract.py
```diff
@@ -12,6 +12,36 @@
...
+                continue
+            for match in node.get("cpeMatch", []):
+                if not isinstance(match, dict) or not match.get("vulnerable"):
+                    continue
+                criteria = match.get("criteria")
+                if isinstance(criteria, str) and criteria and criteria not in affected_cpes:
+                    affected_cpes.append(criteria)
+            visit_nodes(node.get("nodes", []))
+
+    for configuration in configurations:
```
sfl-pe-sync, line 39: WARNING: Deduplicating with `criteria not in affected_cpes` makes extraction quadratic in the number of unique matches. Large NVD configuration trees can contain thousands of distinct CPE criteria, so every NVD refresh now performs millions of string comparisons per such CVE and can substantially delay bulk refreshes. Track seen criteria in a `set` while retaining the list for deterministic output order, and add a large-input regression test.

sfl-pe-sync (approved):
The CPE flow is generally coherent, but extraction introduces avoidable quadratic work for CVEs with large NVD configuration trees.

ValentinBoudevinSFL (approved):

PR #502, opened 2026-08-07
Create scans wizard including refresh options

reh-sfl:

* Create a new Run Scan Wizard to guide users
* Add options to pick EPSS / EUVD refresh to be done on newly added vulns

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Go to the web dashboard, open the Scan tab then click on Run scans, follow the wizard with its different options.

## Pull Request Checklist

### DIFF
--- a/frontend/src/components/RunScansWizard.tsx
+++ b/frontend/src/components/RunScansWizard.tsx
@@ -1,0 +1,248 @@
+import { useEffect, useRef, useState } from "react";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faBug, faCircleQuestion, faCrosshairs, faLeaf, faPlay, faShieldHalved, faXmark } from "@fortawesome/free-solid-svg-icons";
+import type { Variant } from "../handlers/variant";
+import { refreshSourcesForScans } from "../helpers/refreshSources";
+
+type Props = {
+    isOpen: boolean;
+    variants: Variant[];
+    selectedVariantIds: Set<string>;
+    selectedScanTypes: Set<string>;
+    selectedRefreshTypes: Set<string>;
+    refreshMode: "complete" | "custom";
+    excludeKernel: boolean;
+    onClose: () => void;
+    onToggleVariant: (variantId: string) => void;
+    onToggleScanType: (scanType: string) => void;
+    onToggleRefreshType: (refreshType: string) => void;
+    onRefreshModeChange: (mode: "complete" | "custom") => void;
+    onSelectAllVariants: () => void;
+    onSelectNoVariants: () => void;
+    onExcludeKernelChange: (exclude: boolean) => void;
+    onLaunch: () => void;
+};
+
+const scanTypes = [
+    { key: "grype", label: "Grype", description: "Matches SBOM packages against Grype vulnerability databases.", icon: faBug, iconClass: "text-purple-400" },
+    { key: "nvd", label: "NVD CPE", description: "Matches package CPEs against National Vulnerability Database records.", icon: faShieldHalved, iconClass: "text-orange-400" },
+    { key: "osv", label: "OSV", description: "Queries the Open Source Vulnerabilities database for affected package versions.", icon: faLeaf, iconClass: "text-green-400" },
+    { key: "scc", label: "sbom-cve-check", description: "Checks the SBOM against CVE data using sbom-cve-check.", icon: faCrosshairs, iconClass: "text-sky-400" },
+] as const;
+
+const refreshTypes = [
+    { key: "nvd", label: "FKIE", description: "Refresh local NVD-FKIE data for newly discovered CVEs." },
+    { key: "epss", label: "EPSS", description: "Refresh exploit prediction scores for newly discovered CVEs." },
+    { key: "euvd", label: "ENISA EUVD", description: "Refresh EU vulnerability data for newly discovered CVEs." },
+] as const;
+
+const stepLabels = ["Scans", "Variants", "Kernel", "Refresh", "Review"] as const;
+const reviewStep = stepLabels.length;
+
+export default function RunScansWizard({
+    isOpen,
+    variants,
+    selectedVariantIds,
+    selectedScanTypes,
+    selectedRefreshTypes,
+    refreshMode,
+    excludeKernel,
+    onClose,
+    onToggleVariant,
+    onToggleScanType,
+    onToggleRefreshType,
+    onRefreshModeChange,
+    onSelectAllVariants,
+    onSelectNoVariants,
+    onExcludeKernelChange,
+    onLaunch,
+}: Readonly<Props>) {
+    const [step, setStep] = useState(1);
+    const [showKernelHelp, setShowKernelHelp] = useState(false);
+    const dialogRef = useRef<HTMLDivElement>(null);
+
+    useEffect(() => {
+        if (!isOpen) return;
+        setStep(1);
+        setShowKernelHelp(false);
+        dialogRef.current?.focus();
+    }, [isOpen]);
+
+    useEffect(() => {
+        if (!isOpen) return;
+        function handleEscape(event: KeyboardEvent) {
+            if (event.key === "Escape") onClose();
+        }
+        document.addEventListener("keydown", handleEscape);
+        return () => document.removeEventListener("keydown", handleEscape);
+    }, [isOpen, onClose]);
+
+    if (!isOpen) return null;
+
+    const availableRefreshTypes: Set<string> = refreshSourcesForScans(selectedScanTypes);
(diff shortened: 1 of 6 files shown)


ValentinBoudevinSFL (changes requested):
Image

The 4th step needs to be change. 

1- In Vulnerability tab we used "Complete refresh" and "Custom refresh" to avoid complex choice. Should be the same

Image

2- EUVD/EPSS are not sufficient. Let's say I do a Grype scan which finds CVEs, then the refresh via FKIE should be triggered too. 

As a summary:

Grype triggers FKIE, EUVD, EPSS
SCC triggers EUVD, EPSS
NVD triggers FIKE, EUVD, EPSS
OSV triggers nothing 

3- Do not trigger in the queue FKIE, EUVD or EPSS if nothing was found

reh-sfl:
@ValentinBoudevinSFL I can understand that Grype might have some level of latency in data freshness (though minimal) which requires a full FKIE refresh, but I don't agree on the need to add it on an NVD CPE scan because it pulls data exactly from the same `local_database` cache. That said it will not be part of the mapping.

ValentinBoudevinSFL:
> @ValentinBoudevinSFL I can understand that Grype might have some level of latency in data freshness (though minimal) which requires a full FKIE refresh, but I don't agree on the need to add it on an NVD CPE scan because it pulls data exactly from the same `local_database` cache. That said it will not be part of the mapping.

True for both, ignore my feedback 2) @reh-sfl. But 1 and 3 still valid

frontend/src/pages/ScanHistory.tsx
```diff
@@ -1334,6 +1333,20 @@ function ScanHistory({ variantId, projectId, onScanComplete }: Readonly<Props>)
...
             await waitForCompletion();
         }
+        const applicableRefreshTypes = refreshSourcesForScans(selectedScanTypes);
+        const refreshTypes = (refreshMode === 'complete' ? [...applicableRefreshTypes] : [...selectedRefreshTypes])
+            .filter(type => applicableRefreshTypes.has(type));
+        if (refreshTypes.length > 0) {
+            queueVulnerabilityRefresh({
+                refreshTypes,
+                nvdMode: 'local',
+                loadVulnerabilities: () => Vulnerabilities.list(variantId, projectId),
```
sfl-pe-sync, line 1343: ERROR: This reloads vulnerabilities from the page's `variantId`/`projectId`, not from the variants selected in the wizard, and it returns every current vulnerability rather than only findings introduced by these scans. On a project page, selecting one variant therefore sends CVEs from every project variant to all chosen bulk refreshes; on a variant page, selecting additional variants omits their CVEs. Capture the selected variants' vulnerability IDs before scanning and refresh only the post-scan set difference (or collect IDs from the completed scan results), using the multi-variant query parameters so the backend request matches the wizard selection.

sfl-pe-sync (changes requested):
The wizard UI is coherent, but the new refresh path violates its selected-variant and newly-added-vulnerability contract, causing unrelated existing CVEs to be refreshed while some selected variants can be omitted.

frontend/src/pages/ScanHistory.tsx
```diff
@@ -1329,11 +1328,35 @@ function ScanHistory({ variantId, projectId, onScanComplete }: Readonly<Props>)
...
         for (const [scanType, , startQueuedScan, waitForCompletion] of scanQueue) {
             if (!selectedScanTypes.has(scanType)) continue;
             await startQueuedScan();
             await waitForCompletion();
         }
+        const applicableRefreshTypes = refreshSourcesForScans(selectedScanTypes);
+        const refreshTypes = (refreshMode === 'complete' ? [...applicableRefreshTypes] : [...selectedRefreshTypes])
+            .filter(type => applicableRefreshTypes.has(type));
+        if (refreshTypes.length > 0) {
+            queueVulnerabilityRefresh({
```
sfl-pe-sync: ERROR: This ignores the boolean result from `queueVulnerabilityRefresh`. The function rejects a new queue while any refresh is active, and another refresh can start during these potentially long scans after the one-time `refreshAheadOfScans` check. In that race, the wizard closes and reports no error, but none of the selected post-scan refreshes run. Wait for active refreshes again immediately before enqueueing, or handle a `false` return by retrying/awaiting or surfacing an explicit failure.

sfl-pe-sync (changes requested):
The revised flow fixes the previously reported variant/new-finding selection issue, but a concurrent refresh can still cause the wizard's requested post-scan refreshes to be silently discarded.

frontend/src/helpers/refreshSources.ts
```diff
@@ -0,0 +1,12 @@
+import type { RefreshType } from "../handlers/activeScanQueue";
+
+// Refresh sources that can be applied to the vulnerabilities a given set of scans produces.
+export function refreshSourcesForScans(selectedScans: Set<string>): Set<RefreshType> {
+    const sources = new Set<RefreshType>();
+    if (selectedScans.has("grype")) sources.add("nvd");
+    if (selectedScans.has("grype") || selectedScans.has("nvd") || selectedScans.has("scc")) {
+        sources.add("epss");
```
sfl-pe-sync, line 9: ERROR: OSV scans can create CVE findings from each OSV record's `aliases` (the scan route persists every `CVE-*` alias), but an OSV-only selection reaches this condition without adding any refresh source. The wizard therefore claims OSV cannot produce refreshable CVEs and silently skips EPSS/EUVD for newly discovered CVE aliases. Include `osv` in the CVE-producing scan set (and expose each applicable source) and update the test that currently codifies the incorrect empty result.

frontend/src/pages/ScanHistory.tsx
```diff
@@ -1329,11 +1328,39 @@ function ScanHistory({ variantId, projectId, onScanComplete }: Readonly<Props>)
...
+        const applicableRefreshTypes = refreshSourcesForScans(selectedScanTypes);
+        const refreshTypes = (refreshMode === 'complete' ? [...applicableRefreshTypes] : [...selectedRefreshTypes])
+            .filter(type => applicableRefreshTypes.has(type));
+        const targetVariantIds = variants.map(variant => variant.id);
+        const loadSelectedVulnerabilities = () => targetVariantIds.length > 1
+            ? Vulnerabilities.list(undefined, projectId, undefined, undefined, targetVariantIds, 'union')
+            : Vulnerabilities.list(targetVariantIds[0], projectId);
+        // An unavailable baseline must never block the scans: fall back to refreshing the whole selected scope.
+        const existingVulnerabilityIds = new Set(refreshTypes.length === 0 ? [] : (
+            await loadSelectedVulnerabilities().catch(() => [])
```
sfl-pe-sync, line 1341: WARNING: Treating a failed baseline request as an empty baseline makes every existing vulnerability in the selected variants look newly discovered. A transient API failure can therefore launch large NVD/EPSS/EUVD refreshes over the whole scope, contrary to the wizard's promise to refresh only scan additions. Surface the baseline failure and skip/cancel post-scan refresh (while still allowing scans to run), rather than silently substituting `[]`.

sfl-pe-sync (changes requested):
The wizard misses refreshes for CVE aliases produced by OSV and can refresh an entire selected scope when its baseline request fails, so the new post-scan refresh behavior is not yet reliable.

frontend/src/components/RunScansWizard.tsx
```diff
@@ -0,0 +1,248 @@
...
+            className="fixed inset-0 z-50 flex items-center justify-center bg-black/60 p-4"
+            onMouseDown={(event) => {
+                if (event.target === event.currentTarget) onClose();
+            }}
+        >
+            <div ref={dialogRef} tabIndex={-1} role="dialog" aria-modal="true" aria-labelledby="scan-wizard-title" className="w-full max-w-xl rounded-lg border border-sky-700/60 bg-neutral-900 shadow-xl outline-none">
+                <div className="border-b border-neutral-700 px-6 py-4">
+                    <div className="flex items-center justify-between gap-4">
+                        <div>
+                            <h2 id="scan-wizard-title" className="text-lg font-semibold text-white">Run scans</h2>
```
sfl-pe-sync, line 100: WARNING: This modal only moves focus into the dialog when it opens; it neither traps Tab/Shift+Tab nor restores focus to the Run Scans trigger when it closes. Keyboard users can therefore move into and operate the still-active page behind an `aria-modal` dialog, and lose their place after closing it. Use the project's accessible dialog primitive or add a focus trap/background inert handling and restore the previously focused element on close.

sfl-pe-sync (approved):
The scan wizard and refresh orchestration are otherwise coherent, but the new modal needs complete keyboard focus management.

PR #504, opened 2026-08-07
AI review for user assessments

sfl-zwang:

This PR adds a new skill to review a user/custom assessment and a section below the reviewed assessment to display the review results. New tools are also added to the vulnscout-mcp to assist AI agents with the review: https://github.com/savoirfairelinux/vulnscout-mcp/pull/6

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

To review a single assessment, assess a CVE as usual and use an AI agent to invoke the skill `/assessment-review` with an assessment ID, multiple assessment IDs, or a group ID, which can be obtained from the Review page's assessment tab by clicking the copy button in the action cell.

image

Note that an assessment targeting multiple variants or packages copies a group ID instead of a single assessment ID. It is essential to indicate the ID type in the prompt when invoking the review skill.

Example prompts:
```shell
# assessment review for provided ID(s)
/assesment-review assessment ID: 51226a16-22dd-45f0-9b17-069b5f483c46 069b5f483c46-51226a16-22dd-45f0-9b17

# review all custom assessments
/assesment-review 
```

The review is appended to the corresponding assessment on the vulnerability modal:

image

The reviews are stacked when the assessment is a group:

image

Like the AI assessments, there is a button to discard the AI review, but it is only shown in edit mode.

image

### Additional notes

* The review will only target the user/custom assessments. Assessments of other origins will be skipped in all cases.
* The review verdict can be `differs`  when the independently derived review does not exactly match the stored assessment on any of the compared assessment fields:

    * status 
    * justification 
    * impact_statement 
    * workaround 
    * responses 

    So it’s not only status; a match in status can still be `differs` if one of those other compared fields is different.
Also, the confidence gate can force reviewed status to  under_investigation  (MEDIUM in strict mode, LOW always), which can also produce `differs`.

* There won't be a button to accept a review and overwrite the corresponding assessment, for now.
* At most 1 review is allowed per assessment per target at a time. This means a multi-target assessment can have more than one review. 
* The new review will always overwrite the old one.
* When a user edits an assessment that already has a review, the review will be marked as stale.
### Related Issue

## Pull Request Checklist


ValentinBoudevin (changes requested):
1- Frontend VulnModal

It is not obvious which ID correspond to which variants/packages in VulnModal.

Image

Maybe this point doesn't need to be address based on the next issue

2- Backend/DB consistency issue 

Review tab one row should be one ID

Image

IMO database assessment should produce a row with a variants JSON and a packages JSON containing multiples variants/packages. 

It doesn't make sense to have a frontend which display a single assessment but to store it under multiples ID (which are then merge) in the frontend.

3- Review tab frontend Issue

With a basic 100 percent zoom I don't see the column "Actions" anymore

image

4- AI tab discard the project text

Not sure this bug was introduced with this PR, but selecting a variant clear/remove the project context

Step 1 - Enter a project description

image

Step 2 - Select a variant 

image

And the project text is gone

src/routes/assessment_reviews.py
```diff
@@ -0,0 +1,201 @@
...
+        if variant_uuid is None:
+            return None, ({"error": "Internal error"}, 500)
+        return [variant_uuid], None
+    if project_id:
+        project_uuid, err = parse_uuid_or_400(project_id, "project_id")
+        if err:
+            return None, err
+        if project_uuid is None:
+            return None, ({"error": "Internal error"}, 500)
+        return [v.id for v in Variant.get_by_project(project_uuid)], None
```
sfl-pe-sync, line 49: ERROR: A syntactically valid `project_id` with no variants (including a nonexistent project UUID) resolves to `[]`, but both `Assessment.get_by_origin` and `AssessmentReview.get_for_variants` only apply their filter when the list is truthy. These project-scoped endpoints therefore fall back to returning every project's assessments/reviews. Reject unknown projects and preserve an empty scope as an empty result (or make the query helpers distinguish `None` from `[]`), with a regression test for a zero-variant/unknown project.

src/routes/assessment_reviews.py
```diff
@@ -0,0 +1,201 @@
...
+            return {"error": "rationale is required"}, 400
+
+        justification = data.get("justification") or None
+        if justification is not None and justification not in VALID_JUSTIFICATION_OPENVEX:
+            return {
+                "error": f"justification must be one of {', '.join(VALID_JUSTIFICATION_OPENVEX)}"
+            }, 400
+
+        responses = data.get("responses")
+        if responses is not None and not isinstance(responses, list):
```
sfl-pe-sync, line 109: ERROR: This only verifies that `responses` is a list, so payloads such as `["update", {}]` are committed to the JSON column. When the stored review otherwise matches its assessment, `verdict()` sorts that mixed list and raises `TypeError`; the PUT fails after the commit and subsequent review-list requests keep failing on the poisoned row. Validate every element as a supported string response before calling `upsert` (and reject non-object JSON payloads/invalid field types before persistence).

frontend/src/handlers/assessmentReviews.ts
```diff
@@ -0,0 +1,127 @@
...
+        else if (projectId) url.searchParams.set("project_id", projectId);
+
+        try {
+            const response = await fetch(url.toString(), { mode: "cors" });
+            if (!response.ok) return {};
+            return await response.json();
+        } catch {
+            return {};
+        }
+    }
```
sfl-pe-sync, line 109: WARNING: HTTP and network failures are converted into `{}`, which is also the valid “no reviews” result. The Review page then renders every row as not reviewed and the Reviewed filter can hide all rows during a transient API failure, giving analysts incorrect state with no error indication. Let this method reject on failure and have callers retain prior data or render an explicit unavailable/error state.

sfl-pe-sync (changes requested):
The assessment-review feature has blocking backend defects: an empty project scope leaks reviews and assessments from all projects, and insufficient response validation can persist a review that breaks subsequent API serialization. The frontend also misrepresents fetch failures as an absence of reviews.

ValentinBoudevinSFL (changes requested):
1- We should group analysis which are the same

Image

Here we have the exact same conclusion displayed twice. With ten variants it would display 10 times the same text. 

It is just one result which concern two variants. 

Same as VulnModal assessments system which should regroup and an analysis can concern several packages/variants

Image

2- UI issues. 

-Review tab rows are larger 
-Warning flag not center
-Copied text breaks the UI

Image

Image

3- Missing filter on Review tab to get CVEs with 

-AI Review approved
-AI Review differs
-No AI Review

sfl-zwang:
@ValentinBoudevinSFL 

Centered the content:
image

The icon will be swapped only, no more text:
image

Replaced the toggle with a dropdown:
image

src/models/assessment_review.py
```diff
@@ -0,0 +1,257 @@
...
+    def get_for_variants(variant_ids: list[uuid.UUID] | None = None) -> list["AssessmentReview"]:
+        query = db.select(AssessmentReview).join(
+            Assessment, AssessmentReview.assessment_id == Assessment.id
+        ).options(
+            contains_eager(AssessmentReview.assessment),
+            joinedload(AssessmentReview.finding).joinedload(Finding.package),
+        )
+        if variant_ids:
+            query = query.where(
+                Assessment.target_rows.any(AssessmentTarget.variant_id.in_(variant_ids))
```
sfl-pe-sync: ERROR: This scopes reviews by whether the parent assessment has any target in `variant_ids`, not by the review's own `variant_id`. For a multi-target assessment spanning variants, `/api/assessment-reviews?variant_id=A` therefore returns reviews belonging to variant B as well; the frontend then renders and filters on out-of-scope review data. Filter on `AssessmentReview.variant_id.in_(variant_ids)` (and add a regression test with independently reviewed targets in two variants).

.github/skills/assessment-review/SKILL.md
```diff
@@ -0,0 +1,299 @@
...
+
+**Large scopes.** `limit` defaults to 50. If the caller's scope plausibly exceeds
+that, page with `offset` rather than raising `limit` past 100 — each assessment
+costs a full CVE research pass, so confirm with the caller before starting a run
+of more than 25 assessments.
+
+### Re-review policy
+
+By default, **skip assessments where every target already carries a review**,
+by passing `has_review=false`. Re-running over a broad scope should not redo
```
sfl-pe-sync, line 98: ERROR: Passing `has_review=false` does not implement “skip assessments where every target already carries a review”: the route excludes an assessment as soon as any target has a review. A broad review run will therefore never see or review the remaining targets of a partially reviewed multi-target assessment, despite the following instructions saying to inspect `target_reviews`. Fetch all rows and apply the all-target predicate client-side, or change the API filter to return assessments having at least one unreviewed target.

sfl-pe-sync (changes requested):
The per-target review design has two blocking scope/completeness defects: scoped review reads leak sibling-target reviews from other variants, and the skill's default listing filter skips partially reviewed assessments before their missing targets can be processed.

.github/skills/assessment-review/SKILL.md
ValentinBoudevin: Fixed

src/models/assessment_review.py
ValentinBoudevin: Fixed

src/models/assessment_review.py
```diff
@@ -0,0 +1,254 @@
...
+    __tablename__ = "assessment_reviews"
+    __table_args__ = (
+        UniqueConstraint(
+            "assessment_id", "variant_id", "finding_id", name="uq_assessment_review_target"
+        ),
+    )
+
+    id: Mapped[uuid.UUID] = mapped_column(primary_key=True, default=uuid.uuid4)
+    assessment_id: Mapped[uuid.UUID] = mapped_column(
+        ForeignKey("assessments.id", ondelete="CASCADE"), nullable=False, index=True
```
sfl-pe-sync, line 79: ERROR: Reviews are keyed like `AssessmentTarget` but are only linked to `Assessment`, `Variant`, and `Finding` independently. The normal reconcile path removes a deselected target with `assessment.target_rows.remove(target_row)` without deleting its review, so a partial target edit leaves an orphan review that still appears in bulk/modal responses and cannot be deleted through `/review` because `_resolve_target` now rejects it as no longer belonging to the assessment. Delete the matching review whenever reconcile removes a target (or enforce a composite target FK with reliable cascading), and cover removing one reviewed target while retaining another.

src/routes/assessment_reviews.py
```diff
@@ -0,0 +1,315 @@
...
+                if any(
+                    key not in reviews_by_target
+                    for key in scoped_target_keys(assessment)
+                )
+            ]
+
+        rows = []
+        for a in assessments[offset:offset + limit]:
+            row = a.to_dict()
+            row["has_review"] = a.id in reviewed_ids
```
sfl-pe-sync: ERROR: A variant-scoped custom-assessment listing still serializes every target of a matching multi-target assessment. For an assessment spanning variants A and B, `?variant_id=A` returns B in both the assessment payload and `target_reviews`; because `reviews_by_target` is now correctly scoped to A, an existing B review is even reported as absent. The skill is instructed to derive every returned target, so a request scoped to A can re-review and overwrite B. Filter both `targets` and `target_reviews` to `variant_ids` (or return an explicitly scoped representation), with an A/B regression.

frontend/src/pages/Review.tsx
```diff
@@ -110,39 +112,51 @@ function toReviewRow(
...
+const aiReviewList = [
+    AI_REVIEW_LABELS.agrees,
+    AI_REVIEW_LABELS.differs,
+    AI_REVIEW_LABELS.stale,
+    AI_REVIEW_LABELS.none,
+];
+
+/** The AI review label a row carries. An assessment without a review counts
+ *  as "No AI review"; one with reviews on several targets carries the
+ *  worst-case label across them. */
```
sfl-pe-sync: WARNING: Collapsing all target reviews to one `worstVerdictOf` label makes the filter disagree with the per-target column. A partially reviewed assessment never matches “No AI review”, and an assessment with one agreeing and one differing target cannot be found via “AI review agreed”. Build the label set by matching each `row.targets` entry to its review, including `none` for missing targets, so every selected per-target state can match.

sfl-pe-sync (changes requested):
The per-target review feature still has blocking target-lifecycle and variant-scope defects: target edits can orphan undeletable reviews, and variant-scoped skill listings can drive reviews of targets outside the requested variant. The Review page also has stale/filter inconsistencies for multi-target data.

---
- `frontend/src/pages/Review.tsx:355` [warning]: WARNING: Reviews are fetched only when `variantId` or `projectId` changes. Editing an assessment from this page can make its review stale, but the successful reconcile path refreshes assessments without refetching reviews, leaving the AI-review column and stale filter incorrect until the page is remounted or scope changes. Reuse a review refresh callback after successful assessment mutations, as the modal now does.

src/models/assessment_review.py
sfl-pe-sync: Confirmed: `get_for_variants` now filters on `AssessmentReview.variant_id`, and the added regression covers the cross-variant target case.

.github/skills/assessment-review/SKILL.md
sfl-pe-sync: Confirmed: `has_review=false` now retains assessments with at least one unreviewed in-scope target, including partially reviewed assessments.

frontend/src/pages/Review.tsx
ValentinBoudevin: Fixed

src/routes/assessment_reviews.py
ValentinBoudevin: Fixed

src/models/assessment_review.py
ValentinBoudevin: Fixed

.github/skills/assessment-review/SKILL.md
```diff
@@ -0,0 +1,299 @@
...
+vulnscout-write_assessment_review(
+    assessment_id=<id>,
+    variant_id=<the target's variant_id>,
+    package=<the target's package>,
+    status=<derived or under_investigation>,
+    rationale=<why>,
+    status_notes=<proposed notes + confidence level: ...>,
+    justification=<only for not_affected>,
+    impact_statement=<only for not_affected>,
+    workaround=<only when one exists>,
```
sfl-pe-sync, line 252: ERROR: The workflow requires `responses` to be independently derived and compared (line 195), but the prescribed write call never sends that field. The backend consequently stores `[]`, so an otherwise identical review of any assessment with responses such as `update` or `will_not_fix` is falsely reported as `differs`, and a genuinely proposed response is lost. Pass `responses=<derived responses>` in Phase 4 (including an explicit empty list when none apply) and cover the matching-nonempty-response path.

frontend/src/components/VulnModal.tsx
```diff
@@ -2050,35 +2095,91 @@ type VariantScopedSnapshot = {
...
+                                                            {hasStatusNotes && <>{row.status_notes}<br/></>}
+                                                            {hasWorkaround && row.workaround}
                                                         </p>
                                                     )}
                                                 </div>
                                             </div>
+                                            {row.origin === "custom" && (() => {
+                                                const assessmentId = row.id;
+                                                const rowReviews = reviews[assessmentId] ?? [];
+                                                const multiTarget = rowReviews.length > 1;
```
sfl-pe-sync: WARNING: `multiTarget` is inferred from the number of reviews rather than the number of assessment targets. When a multi-target assessment has only one reviewed target, the sole card omits its package/variant label, so users cannot tell which target the result covers. Base this condition on `row.targets` (for example `isMultiTarget(targets)`) and keep the target label visible for partially reviewed rows.

frontend/src/components/VulnModal.tsx
```diff
@@ -2050,35 +2095,91 @@ type VariantScopedSnapshot = {
...
+                                                                                Discard review
+                                                                            </button>
+                                                                        )}
+                                                                    </div>
+                                                                    <p className="text-sm text-gray-200 whitespace-pre-line">
+                                                                        <strong>{review.status}</strong>
+                                                                        {review.justification && <> · {review.justification}</>}<br/>
+                                                                        {review.impact_statement && <>{review.impact_statement}<br/></>}
+                                                                        {review.status_notes && <>{review.status_notes}<br/></>}
+                                                                        {review.workaround && <>{review.workaround}<br/></>}
```
sfl-pe-sync, line 2165: WARNING: The review card renders every compared VEX field except `responses`. A review can therefore be marked `differs` solely because its proposed response tags differ while the UI shows no proposed value, preventing the analyst from understanding the discrepancy. Render `review.responses` alongside the other review fields.

sfl-pe-sync (changes requested):
The target-level scoping and cleanup fixes are correct, but the skill currently loses derived response tags and can produce incorrect review verdicts; the modal also needs to expose target identity and response differences for partially reviewed multi-target assessments.

src/models/assessment_review.py
sfl-pe-sync: Confirmed: reconciliation now removes the matching review through the delete-orphan relationship before removing the target, and the regression preserves the sibling review.

src/routes/assessment_reviews.py
sfl-pe-sync: Confirmed: the custom-assessment response now rebuilds packages, variants, targets, and target review metadata from only the requested scope.

frontend/src/pages/Review.tsx
sfl-pe-sync: Confirmed: filtering now derives the full set of per-target states, including missing reviews, instead of collapsing to one worst verdict.

frontend/src/components/VulnModal.tsx
ValentinBoudevin: Fixed

frontend/src/components/VulnModal.tsx
ValentinBoudevin: Fixed

.github/skills/assessment-review/SKILL.md
ValentinBoudevin: Fixed

src/helpers/outdated_cleanup.py
```diff
@@ -553,9 +569,11 @@ def delete_outdated_data(candidate_ids: dict[str, object] | None = None) -> dict
             if remove_target(assessment_id, variant_id, finding_id):
                 targets_removed += 1
         # Bulk DELETE bypasses the ORM's cascade="all, delete-orphan" on
-        # Assessment.target_rows (sqlite foreign_keys stay off), so the
-        # target rows for these assessments are cleared explicitly first —
-        # otherwise they'd be left dangling, pointing at a deleted assessment.
+        # Assessment.target_rows/.review (sqlite foreign_keys stay off), so
```
sfl-pe-sync, line 572: ERROR: This only deletes reviews for `fully_stale_ids`. Earlier in the same cleanup, `remove_target(...)` can remove one stale target from a multi-target assessment that remains alive; because reviews are independently keyed by `(assessment_id, variant_id, finding_id)` and are not children of `AssessmentTarget`, that target's review survives and continues to appear in `/api/assessment-reviews`, while the target-scoped DELETE endpoint can no longer resolve it. Delete the matching review whenever this cleanup removes an individual target (before `remove_target`), and add a regression for one stale reviewed target plus one retained target.

frontend/src/pages/Review.tsx
```diff
@@ -343,6 +367,19 @@ function Review({ variantId, projectId, onAssessmentChanged }: Readonly<Props>)
         return () => { cancelled = true; };
     }, [editVariants, editingRow]);
```
sfl-pe-sync, line 369: WARNING: This mutation-triggered refresh has no cancellation or scope/version guard. If a user edits or deletes an assessment and switches project/variant before the review request completes, the old scope's response can resolve after the new scope's effect and overwrite `reviews`, making the new page show/filter rows as unreviewed. Use an `AbortController` or monotonically increasing request token tied to the current scope before calling `setReviews`; the analogous `refreshReviews` callback in `VulnModal` needs the same protection.

sfl-pe-sync (changes requested):
The latest fixes address the three open review-display issues, but target cleanup can still leave inaccessible review rows, and mutation refreshes can overwrite a newer scope with stale data.

.github/skills/assessment-review/SKILL.md
sfl-pe-sync: Confirmed: the skill now always passes the independently derived `responses`, including an explicit empty list, and the regression covers a matching non-empty response list.

frontend/src/components/VulnModal.tsx
sfl-pe-sync: Confirmed: multi-target labeling now derives from the assessment targets, so a partially reviewed assessment labels its sole review card.

frontend/src/components/VulnModal.tsx
sfl-pe-sync: Confirmed: proposed response tags are now rendered in the review card and covered by a frontend test.

src/helpers/outdated_cleanup.py
ValentinBoudevin: The feedback is not relevant to the current code

delete_outdated_data() already removes partial stale targets through remove_target(). That helper explicitly:

-Finds the review matching (assessment_id, variant_id, finding_id).
-Deletes that review.
-Removes the target row.
-Deletes the assessment only when no targets remain.

This behavior was introduced in commit 570ff69e when reviews became target-scoped. The reviewer appears to have inspected only the later bulk deletion of reviews for fully stale assessments and overlooked the cleanup inside remove_target().

.github/skills/assessment-review/SKILL.md
```diff
@@ -0,0 +1,302 @@
...
+For each distinct `variant_id` in the scope, call once and cache for the whole
+run:
+
+```
+context = vulnscout-get_merged_context(project_id=<id>, variant_id=<id>)
+```
+
+`get_custom_assessment` returns the assessment's `variant_id`;
+`list_custom_assessments` returns it per row. Resolve `project_id` with
+`vulnscout-find_project_id` when the caller gave a project name, or from the
```
sfl-pe-sync: ERROR: Assessment-ID and unscoped review flows cannot resolve the `project_id` required by `get_merged_context`. `get_custom_assessment` and unscoped `list_custom_assessments` expose target `variant_id`s but no project identity, while the only project resolver requires a project name that these prompts do not provide. The agent therefore cannot load the project/variant context for the primary documented scopes and may derive reviews against the wrong/default source context. Include `project_id` with each returned target (or add a variant-to-project resolver / variant-only context tool) and document that path here.

.github/skills/assessment-review/SKILL.md
```diff
@@ -0,0 +1,302 @@
...
+by passing `has_review=false`. Re-running over a broad scope should not redo
+settled work. The filter returns assessments with at least one unreviewed target
+in the requested scope, including partially reviewed multi-target assessments.
+`has_review` on a listing row is still true when *any* target has a review, so
+check `target_reviews` (each target's own `has_review`/`is_stale`) and derive
+only the still-unreviewed targets normally.
+
+Two exceptions:
+- The caller explicitly asks to re-review everything → omit `has_review`.
+- A target's existing review is stale (`is_stale=true` on that target's entry,
```
sfl-pe-sync, line 108: ERROR: The stated stale-review exception is unreachable for broad project/all-assessment runs. Those runs pass `has_review=false`, but the API treats a stale review as an existing review, so a single-target stale assessment (or one whose every target is stale) is filtered out before `target_reviews` can be inspected and is never re-reviewed. Either fetch all rows and client-filter fresh reviews versus stale/unreviewed targets, or make the listing filter treat stale targets as pending review, with a regression for an all-stale assessment.

src/routes/assessment_reviews.py
```diff
@@ -0,0 +1,344 @@
...
+        try:
+            limit = int(request.args.get('limit', 50))
+            offset = int(request.args.get('offset', 0))
+        except ValueError:
+            return {"error": "limit and offset must be integers"}, 400
+        if limit < 1 or offset < 0:
+            return {"error": "limit must be >= 1 and offset >= 0"}, 400
+
+        # get_by_origin already orders by timestamp descending.
+        assessments = Assessment.get_by_origin(variant_ids, origin="custom")
```
sfl-pe-sync, line 274: WARNING: Paging is not stable because `get_by_origin` orders only by `timestamp`, yet VulnScout deliberately creates batches with identical timestamps. Separate `offset` requests can therefore reorder tied rows and skip or repeat assessments during a broad review run. Add a unique tie-breaker such as `Assessment.id` in the requested direction (and cover equal-timestamp rows across page boundaries).

sfl-pe-sync (changes requested):
The target-scoped review model is substantially improved, but the documented broad/ID workflows still cannot reliably load context or re-review stale assessments, and pagination can miss tied rows. These core workflow defects should be fixed before merge.

src/helpers/outdated_cleanup.py
sfl-pe-sync: You're right; I overlooked the target-level cleanup inside `remove_target()`. It explicitly deletes the matching `(assessment_id, variant_id, finding_id)` review before removing the target, so the reported orphan path is already handled.

.github/skills/assessment-review/SKILL.md
ValentinBoudevin: Fixed

.github/skills/assessment-review/SKILL.md
ValentinBoudevin: Fixed

src/routes/assessment_reviews.py
```diff
@@ -0,0 +1,366 @@
...
+
+    if (variant_uuid, finding_uuid) not in set(assessment.targets):
+        return None, None, ({"error": "Not a target of this assessment"}, 400)
+    return variant_uuid, finding_uuid, None
+
+
+def _scoped_variant_ids() -> "tuple[list[UUID] | None, ResponseReturnValue | None]":
+    """Resolve variant_id / project_id query args into a variant ID list."""
+    variant_id = request.args.get('variant_id')
+    project_id = request.args.get('project_id')
```
sfl-pe-sync, line 78: ERROR: When both `project_id` and `variant_id` are supplied, this branch returns the variant scope without parsing or checking `project_id`. A request such as `?project_id=A&variant_id=<variant from B>` therefore returns project B assessments/reviews despite explicitly claiming project A scope. Parse both identifiers and reject a variant whose `project_id` does not match the requested project (or intersect the two scopes), with regressions for both listing endpoints.

src/routes/assessment_reviews.py
```diff
@@ -0,0 +1,366 @@
...
+
+        try:
+            limit = int(request.args.get('limit', 50))
+            offset = int(request.args.get('offset', 0))
+        except ValueError:
+            return {"error": "limit and offset must be integers"}, 400
+        if limit < 1 or offset < 0:
+            return {"error": "limit must be >= 1 and offset >= 0"}, 400
+
+        # get_by_origin already orders by timestamp descending.
```
sfl-pe-sync, line 273: WARNING: `limit` and `offset` are applied only after `get_by_origin` materializes every custom assessment and `get_for_variants` materializes every review in scope. Thus even `limit=1` performs work and retains objects proportional to the full database, which makes the documented broad/all-assessment workflow increasingly expensive. Push stable ordering, review-presence filtering, offset, and limit into the SQL query (and fetch reviews only for the selected assessment IDs).

sfl-pe-sync (changes requested):
The latest patchset fixes the two open-thread issues, but project/variant scope can still be bypassed when both filters are supplied, so the PR is not ready to merge. The new listing endpoint also does unbounded in-memory pagination.

.github/skills/assessment-review/SKILL.md
sfl-pe-sync: Confirmed: the current payloads now expose `project_id` both at assessment level and on each target, so assessment-ID and unscoped runs can resolve merged context.

.github/skills/assessment-review/SKILL.md
sfl-pe-sync: Confirmed: `has_review=false` now includes targets whose existing review is stale, and the added regression covers an all-stale assessment.

src/routes/assessment_reviews.py
ValentinBoudevin: FIxed

src/models/assessment_review.py
```diff
@@ -0,0 +1,254 @@
...
+        review.workaround = workaround
+        review.responses = list(responses or [])
+        review.reviewed_fingerprint = fingerprint_assessment(
+            db.session.get(Assessment, assessment_id)
+        )
+        review.timestamp = datetime.now(timezone.utc)
+        db.session.commit()
+        return review
+
+    def delete(self) -> None:
```
sfl-pe-sync, line 242: ERROR: This fingerprints the assessment at review-write time, not the version the agent actually reviewed. The documented workflow fetches the assessment, performs a potentially long research pass, and only then writes; if an analyst edits the assessment in that interval, this code hashes the new content and records the old review as fresh even though its comparison and rationale were derived from the previous fields. Return a content fingerprint/version with the read payload and require it on PUT, rejecting the write with 409 when it no longer matches (or persist that supplied fingerprint), with a regression covering an edit between read and write.

sfl-pe-sync (changes requested):
The combined project/variant scope fix is correct, but review staleness can still be defeated by an assessment edit during the AI research window, allowing an outdated review to be recorded as fresh.

src/routes/assessment_reviews.py
sfl-pe-sync: Confirmed: the current patch parses both IDs and rejects a variant whose project does not match the supplied project scope.

src/models/assessment_review.py
ValentinBoudevin: Fixed

src/routes/assessment_reviews.py
```diff
@@ -0,0 +1,388 @@
...
+        status = data.get("status")
+        rationale = data.get("rationale")
+
+        if status not in VALID_STATUS_OPENVEX:
+            return {"error": f"status must be one of {', '.join(VALID_STATUS_OPENVEX)}"}, 400
+        if not rationale or not str(rationale).strip():
+            return {"error": "rationale is required"}, 400
+
+        justification = data.get("justification") or None
+        if justification is not None and justification not in VALID_JUSTIFICATION_OPENVEX:
```
sfl-pe-sync, line 187: ERROR: The review endpoint accepts `status="not_affected"` with no justification, even though VulnScout's normal assessment write path rejects that invalid VEX combination. This lets the AI persist an incomplete proposed assessment (and potentially report it as agreeing when the stored assessment is similarly malformed). Apply the same conditional validation used by `payload_to_assessment`/`is_justification_required`, and add a PUT regression for a missing justification.

sfl-pe-sync (changes requested):
The concurrency-token fix is sound, but the new write endpoint can still persist an invalid not-affected review without the required VEX justification, so the PR is not ready to merge.

src/models/assessment_review.py
sfl-pe-sync: Confirmed: the PUT now requires the fingerprint returned by the read API, rejects a mismatch with `assessment_changed`, and persists the supplied read-time fingerprint.

src/routes/assessment_reviews.py
ValentinBoudevin: Fixed

frontend/src/components/VulnModal.tsx
```diff
@@ -189,6 +193,25 @@ type VariantScopedSnapshot = {
...
+    // current content, so an edit/delete/approve leaves the previously
+    // fetched review stale-flag stale (pun intended) until this refetches.
+    const refreshReviews = useCallback(() => {
+        AssessmentReviews.fetchForScope(variantId, projectId)
+            .then(data => setReviews(data))
+            .catch(() => setReviews({}));
+    }, [variantId, projectId]);
+
+    useEffect(() => {
+        let cancelled = false;
```
sfl-pe-sync, line 208: WARNING: Every modal open fetches `/api/assessment-reviews` for the entire variant/project even though this component renders reviews only for the current vulnerability's assessment rows. Because that endpoint is unpaginated and serializes every matching review, opening or navigating the modal costs O(all reviews in scope) network, query, and serialization work and can become slow on production projects. Fetch `/api/assessments/<id>/reviews` for the displayed assessment IDs, or add an assessment/vulnerability filter to the collection endpoint and use it here.

sfl-pe-sync (dismissed):
The target-level review behavior is largely coherent, but the vulnerability modal introduces an unbounded project-wide review fetch on every open/navigation, which should be narrowed to the current vulnerability's assessments.

src/routes/assessment_reviews.py
sfl-pe-sync: Confirmed: the endpoint now rejects `not_affected` reviews without a justification before persistence, and the regression verifies that no review row is created.

frontend/src/components/VulnModal.tsx
ValentinBoudevin: Fixed

src/routes/assessment_reviews.py
```diff
@@ -0,0 +1,393 @@
...
+            return {"error": "responses must be a list"}, 400
+
+        expected_fingerprint = data.get("expected_assessment_fingerprint")
+        if not isinstance(expected_fingerprint, str) or not expected_fingerprint:
+            return {"error": "expected_assessment_fingerprint is required"}, 400
+        current_fingerprint = fingerprint_assessment(assessment)
+        if expected_fingerprint != current_fingerprint:
+            return {
+                "error": "Assessment changed after it was read; review it again",
+                "code": "assessment_changed",
```
sfl-pe-sync, line 203: ERROR: The concurrency token is checked before `AssessmentReview.upsert()` starts and commits its write, so the check is still subject to a TOCTOU race. An assessment reconcile can change the authored fields or remove this target after `_resolve_target`/`fingerprint_assessment` succeeds but before the review insert commits; the resulting review can be marked fresh against content the agent never reviewed, or can recreate an orphan review for a removed target because there is no composite FK to `assessment_targets`. Perform target resolution, fingerprint comparison, and upsert atomically under the same write serialization/transaction used by assessment mutations (or use a database version/conditional write), then add a concurrent-write regression.

sfl-pe-sync (changes requested):
The latest patchset fixes the previously reported modal over-fetching, but the stale-write protection remains non-atomic and can still persist a fresh-looking or orphaned review during a concurrent assessment edit.

frontend/src/components/VulnModal.tsx
sfl-pe-sync: Confirmed: the modal now sends `vulnerability_id`, and the backend filters reviews by the joined finding vulnerability, so opening it no longer retrieves all reviews in the project/variant scope.

PR #505, opened 2026-08-10
feat(assessments): remove false positive option from new assessment status

ValentinBoudevin:

Restrict new assessment statuses to pending assessment, affected, not applicable and fixed. Drop the now-unreachable false_positive branches in StatusEditor and remove the obsolete tests.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

False positive not available anymore in "Add a new assessment" menu: 

image

### DIFF
--- a/frontend/src/components/StatusEditor.tsx
+++ b/frontend/src/components/StatusEditor.tsx
@@ -209,14 +209,6 @@ function StatusEditor ({onAddAssessment, progressBar, clearFields: shouldClearFi
             }
             return;
         }
-        if (status == "false_positive" && impact == '') {
-            if (triggerBanner) {
-                triggerBanner("You must provide an impact statement for false positive status", "error");
-            } else {
-                internalTriggerBanner("You must provide an impact statement for false positive status", "error");
-            }
-            return;
-        }
         if (variants && variants.length > 0 && selectedVariantIds.length === 0) {
             if (triggerBanner) {
                 triggerBanner("You must select at least one variant", "error");
@@ -238,7 +230,7 @@ function StatusEditor ({onAddAssessment, progressBar, clearFields: shouldClearFi
             justification: status == "not_affected" ? justification : undefined,
             status_notes: statusNotes,
             workaround,
-            impact_statement: (status == "not_affected" || status == "false_positive") ? impact : undefined,
+            impact_statement: status == "not_affected" ? impact : undefined,
             variant_ids: selectedVariantIds.length > 0 ? selectedVariantIds : undefined,
             packages: selectedPackages.length > 0 ? selectedPackages : (availablePackages ?? [])
         });
@@ -286,7 +278,6 @@ function StatusEditor ({onAddAssessment, progressBar, clearFields: shouldClearFi
                 <option value="affected">Affected / exploitable</option>
                 <option value="fixed">Fixed / patched</option>
                 <option value="not_affected">Not applicable</option>
-                <option value="false_positive">False positive</option>
             </select>
             {status == "not_affected" && <>
                 Justification:
@@ -394,7 +385,7 @@ function StatusEditor ({onAddAssessment, progressBar, clearFields: shouldClearFi
                 </div>
             </div>
         )}
-        {(status == "not_affected" || status == "false_positive") && <>
+        {status == "not_affected" && <>
             <textarea
                 value={impact}
                 onChange={(event: React.ChangeEvent<HTMLTextAreaElement>) => setImpact(event.target.value)}
--- a/frontend/src/components/StatusEditor.tsx
+++ b/frontend/src/components/StatusEditor.tsx
@@ -14,6 +14,18 @@ type PostAssessment = {
     variant_ids?: string[]
 }
 
+// Statuses that can be created through this editor (must match the <option>
+// values rendered in the Status <select>).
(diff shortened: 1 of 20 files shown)


frontend/src/components/StatusEditor.tsx
```diff
@@ -238,7 +230,7 @@ function StatusEditor ({onAddAssessment, progressBar, clearFields: shouldClearFi
             justification: status == "not_affected" ? justification : undefined,
```
sfl-pe-sync, line 242: ERROR: `false_positive` is still reachable through the `defaultStatus` prop even though its option is hidden. `MultiEditBar` passes the selected vulnerabilities' `uniformStatus`, and `VulnModal` also derives a default from existing assessments, so opening the add-assessment editor on legacy false-positive data leaves React state as `false_positive`; clicking Add can post a new false-positive assessment (now without the impact validation or payload) despite the UI restriction. Normalize unsupported `defaultStatus` values to an allowed creation status before initializing/resetting `status`, and add a regression test covering `defaultStatus="false_positive"`.

sfl-pe-sync (changes requested):
The hidden option does not fully enforce the new-assessment restriction because legacy false-positive status can still enter through existing `defaultStatus` callers and be submitted.

sfl-pe-sync (changes requested):
The intended status restriction is implemented in the selector, but the changed behavior has no direct regression coverage.

---
- `frontend/tests/unit_tests/test_status_editor.tsx` [warning]: WARNING: This patch removes every `false_positive` creation test but adds no assertion that the status selector now excludes that value. The PR's only intended behavior can therefore regress without any test failing; add a test that inspects the selector options and asserts the creation-status set is exactly `under_investigation`, `affected`, `fixed`, and `not_affected`.

frontend/src/components/StatusEditor.tsx
ValentinBoudevin: Fixed

sfl-pe-sync (dismissed):
No actionable defects found.

frontend/src/components/StatusEditor.tsx
```diff
@@ -14,6 +14,18 @@ type PostAssessment = {
     variant_ids?: string[]
 }
 
+// Statuses that can be created through this editor (must match the <option>
+// values rendered in the Status <select>).
+const CREATABLE_STATUSES = ["under_investigation", "affected", "fixed", "not_affected"] as const;
```
reh-sfl, line 19: I think this is a UI fix that doesn't eliminate the False Positive option itself from the system, you can look at the statuses mapping in the frontend and go from there to remove any hardcoded False Positive (in the API for example).

reh-sfl (changes requested):

frontend/src/components/StatusEditor.tsx
ValentinBoudevinSFL: Done

src/models/assessment.py
```diff
@@ -243,8 +241,14 @@ def add_package(self, package: str | Package) -> bool:
     def set_status(self, status: str) -> bool:
         """Validate and set the assessment status (OpenVEX or CDX VEX).
 
+        ``false_positive`` was removed as a distinct status; any incoming value
+        (for example from an imported CycloneDX VEX document) is absorbed into
+        ``not_affected`` so it is never stored as a first-class status.
+
         Return ``True`` if the status was accepted, ``False`` otherwise.
         """
+        if status == "false_positive":
```
sfl-pe-sync, line 250: ERROR: `set_status()` is also the validator used by the create/update API routes, so this normalization makes `status: "false_positive"` a successful request that is silently stored as `not_affected`; the backend therefore does not actually restrict new assessment statuses to the four advertised values and clients cannot detect invalid input. Reject `false_positive` in the API/model validation path and perform the legacy CycloneDX conversion explicitly at the importer boundary instead.

src/views/cyclonedx.py
```diff
@@ -44,7 +44,6 @@ def __init__(self, controllers: ControllersCache):
         "resolved_with_pedigree": ImpactAnalysisState.RESOLVED_WITH_PEDIGREE,
         "exploitable": ImpactAnalysisState.EXPLOITABLE,
         "in_triage": ImpactAnalysisState.IN_TRIAGE,
-        "false_positive": ImpactAnalysisState.FALSE_POSITIVE,
         "not_affected": ImpactAnalysisState.NOT_AFFECTED,
```
sfl-pe-sync, line 47: ERROR: Existing database rows are loaded directly and never pass through `Assessment.set_status()`, but this patch removes every read/export classification for their persisted `false_positive` value without adding a data migration. Such a row can become `[invalid status]` in the frontend, disappear from ignored/inactive report filters, and export through `str_to_vex_status()` as its fallback `in_triage`, changing an inactive vulnerability into pending. Backfill persisted rows to `not_affected` during upgrade (or retain legacy read mappings until they are normalized) before removing these branches, with an upgrade regression test.

sfl-pe-sync (changes requested):
The creation UI is restricted, but backend validation still accepts the removed value and existing persisted assessments are not migrated, causing status misclassification and incorrect CycloneDX exports. These compatibility and contract regressions block merge.

PR #506, opened 2026-08-10
Clarify vulnerability table empty states and missing-data banner

ValentinBoudevin:

* Replaced the EU KEV / published-date specific banner messages with a single generic message telling users to use the "Refresh vulnerability data" button, so the banner no longer depends on which data source is missing.

*euvd_fetched_at is now stamped for every CVE processed during an EUVD sync (both bulk refresh and initial enrichment), not only CVEs matched to an EUVD entry. This lets the EU KEV column show — for vulnerabilities that were synced but are not known-exploited, versus Requires Refresh Vulnerability Data for vulnerabilities never synced.

*The Published Date column now shows an em-dash — only when an NVD or GHSA refresh has actually run but returned no date, and Requires Refresh Vulnerability Data when no refresh has run yet. 

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Look at Vulnerability tab:

- Add vulnerabilities (e.g via sbom-cve-check) and new Vulnerabilities will display "Requires Refresh Vulnerability Data" until a new sync is performed
- Sync data now display properly "-" if not found/not known exploitable

### DIFF
--- a/frontend/src/pages/TableVulnerabilities.tsx
+++ b/frontend/src/pages/TableVulnerabilities.tsx
@@ -578,14 +578,8 @@ function TableVulnerabilities ({ vulnerabilities, filterLabel, filterValue, filt
         if (missingPublishedDateDataBannerDismissed === undefined) setLocalMissingPublishedDateDataBannerDismissed(dismissed);
     }, [missingPublishedDateDataBannerDismissed, onMissingPublishedDateDataBannerDismissedChange]);
 
-    // The EU KEV column renders a badge only when a vulnerability is flagged
-    // known_exploited; every other row shows an empty placeholder. The backend
-    // also always serialises an `euvd` object (often with just an alias id and
-    // known_exploited=false), so object presence does not indicate KEV data.
-    // Mirror the column: KEV data "exists" only when at least one vulnerability
-    // is actually known-exploited.
     const hasAnyEuvdData = useMemo(
-        () => vulnerabilities.some(v => v.euvd?.known_exploited === true),
+        () => vulnerabilities.some(v => typeof v.euvd_fetched_at === "string" && v.euvd_fetched_at !== ""),
         [vulnerabilities]
     );
     const previousHasAnyEuvdData = useRef(hasAnyEuvdData);
--- a/frontend/src/pages/TableVulnerabilities.tsx
+++ b/frontend/src/pages/TableVulnerabilities.tsx
@@ -603,11 +603,7 @@ function TableVulnerabilities ({ vulnerabilities, filterLabel, filterValue, filt
 
     const shouldShowMissingDataBanner = shouldShowMissingEuvdDataBanner || shouldShowMissingPublishedDateDataBanner;
 
-    const missingDataBannerMessage = shouldShowMissingEuvdDataBanner && shouldShowMissingPublishedDateDataBanner
-        ? <><strong className="font-bold">EU KEV data</strong> and <strong className="font-bold">published date data</strong> need updating. Use the "Refresh vulnerability data" button to update them.</>
-        : shouldShowMissingEuvdDataBanner
-            ? <><strong className="font-bold">EU KEV data</strong> needs updating. Use the "Refresh vulnerability data" button to update it.</>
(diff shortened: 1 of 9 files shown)


frontend/src/pages/TableVulnerabilities.tsx
```diff
@@ -574,11 +574,7 @@ function TableVulnerabilities ({ vulnerabilities, filterLabel, filterValue, filt
 
     const shouldShowMissingDataBanner = shouldShowMissingEuvdDataBanner || shouldShowMissingPublishedDateDataBanner;
```
sfl-pe-sync, line 599: ERROR: The banner still derives `hasAnyEuvdData` solely from `v.euvd?.known_exploited === true`, so a successful EUVD refresh in which none of the table's CVEs are known-exploited stamps every `euvd_fetched_at` and renders dashes in the cells, but leaves this generic "incomplete" banner visible indefinitely. Base banner completeness on the new per-row `euvd_fetched_at` signal (and add the refreshed-with-no-KEV regression case) rather than requiring a positive KEV match.

src/routes/bulk_refresh.py
```diff
@@ -520,6 +520,11 @@ def _run() -> None:
                                 matched += 1
                                 if known_exploited:
                                     kev_matched += 1
+                            else:
+                                # No EUVD entry for this CVE, but still record that it
+                                # was checked during this refresh so the UI can tell
+                                # "synced, not on KEV list" apart from "never refreshed".
+                                rec.update_record(euvd_fetched_at=now, commit=False)
```
sfl-pe-sync: ERROR: The newly handled unmatched path only advances `euvd_fetched_at`; it does not clear prior EUVD/KEV values. If a CVE that was previously marked known-exploited is later absent from both current mappings, the refresh records success but the serializer retains `euvd_known_exploited=True`, and the frontend prioritizes that stale flag over the timestamp and continues showing "Known Exploited". Clear the stale EUVD/KEV fields in this branch, and make the equivalent correction in `fetch_euvd_data`, with a transition test from previously matched to now unmatched.

sfl-pe-sync (changes requested):
The new empty-state distinction is not reliable yet: successful no-match refreshes keep the missing-data banner active, and unmatched refreshes can preserve a stale known-exploited result.

src/routes/bulk_refresh.py
ValentinBoudevin: Fixed

frontend/src/pages/TableVulnerabilities.tsx
```diff
@@ -549,14 +549,8 @@ function TableVulnerabilities ({ vulnerabilities, filterLabel, filterValue, filt
...
 
-    // The EU KEV column renders a badge only when a vulnerability is flagged
-    // known_exploited; every other row shows an empty placeholder. The backend
-    // also always serialises an `euvd` object (often with just an alias id and
-    // known_exploited=false), so object presence does not indicate KEV data.
-    // Mirror the column: KEV data "exists" only when at least one vulnerability
-    // is actually known-exploited.
     const hasAnyEuvdData = useMemo(
-        () => vulnerabilities.some(v => v.euvd?.known_exploited === true),
+        () => vulnerabilities.some(v => typeof v.euvd_fetched_at === "string" && v.euvd_fetched_at !== ""),
```
sfl-pe-sync, line 582: WARNING: Using `some` marks EUVD data complete as soon as any row has been refreshed. After a table is refreshed and a new CVE is imported, the older rows carry `euvd_fetched_at`, the new row renders “Requires Refresh Vulnerability Data,” but this banner is suppressed. Base the missing state on whether any relevant row lacks `euvd_fetched_at` (or require every non-empty row to have it), and add a mixed refreshed/unrefreshed regression test.

sfl-pe-sync (approved):
The stale EUVD fields are now cleared in both refresh paths, but mixed refreshed and newly imported vulnerabilities can still suppress the incomplete-data banner.

reh-sfl (approved):

PR #507, opened 2026-08-10
fix(scope): silently fall back to default scope on DB switch

ValentinBoudevin:

When a saved frontend scope no longer matches the loaded database (e.g. after loading a different DB), fall back to the server default scope without surfacing a red error banner. The scope is still validated and cleared as before; only the misleading error banner is removed.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Look at the code, this banner does not exist anymore

image


sfl-pe-sync (approved):
No actionable defects found.

reh-sfl (approved):

PR #508, opened 2026-08-10
fix(epss): remove CVE-count cap that blocked bulk EPSS refresh

ValentinBoudevin:

The bulk EPSS refresh rejected any request with more than _MAX_CVE_IDS (1000) CVEs with a 400, but the frontend sends every CVE in the loaded database. On any database with >1000 CVEs the refresh always failed to start ("Failed to start EPSS refresh").

The background job already chunks the CVE list into _EPSS_BATCH_SIZE batches against the FIRST.org API and tolerates per-batch failures, so the hard cap is unnecessary. Remove it and accept arbitrarily large lists. Update the route test to assert a >1000 CVE request now returns 202 instead of 400.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

EPSS refresh does not fail anymore

image

### DIFF
--- a/src/routes/bulk_refresh.py
+++ b/src/routes/bulk_refresh.py
@@ -239,8 +239,9 @@ def init_app(app: Flask) -> None:
         cve_ids = [c for c in cve_ids if _CVE_RE.match(c)]
         if not cve_ids:
             return jsonify({"error": "cve_ids must contain valid CVE identifiers (e.g. CVE-2024-1234)"}), 400
-        if len(cve_ids) > _MAX_CVE_IDS:
-            return jsonify({"error": f"cve_ids must contain at most {_MAX_CVE_IDS} entries"}), 400
+        # No hard cap here: the refresh below chunks the CVE list into batches
+        # of _EPSS_BATCH_SIZE against the FIRST.org API, so an arbitrarily large
+        # database (which the frontend sends in full) is handled gracefully.
 
         total = len(cve_ids)
         if not EPSSProgressTracker.start_if_idle("bulk_epss_refresh"):
--- a/src/routes/bulk_refresh.py
+++ b/src/routes/bulk_refresh.py
@@ -47,6 +47,11 @@ _GHSA_SLEEP_INTERVAL = 1.0
 # EUVD enrichment annotates from already-cached ENISA dumps (no per-CVE network
 # call), so it is fast; commit in larger batches than the network-bound refreshes.
 _EUVD_COMMIT_EVERY = 200
+# Chunk size for the "which of these CVE IDs exist?" lookup. Kept well under
+# SQLite's 999 bound-parameter limit so the intersection can be done with
+# indexed ``WHERE id IN (...)`` queries bounded by the request size, instead of
+# materializing the whole vulnerabilities table.
+_DB_LOOKUP_CHUNK = 500
 
 
 def _nvd_sleep_interval() -> float:
@@ -220,9 +225,16 @@ def init_app(app: Flask) -> None:
 
         Body: ``{"cve_ids": ["CVE-A", "CVE-B", ...]}``
 
+        The submitted IDs are deduplicated and intersected with the CVEs stored
+        in the database, so the background work is bounded by the database size
+        regardless of how large the request body is. The singleton refresh is
+        reserved *before* any database access, so a request that arrives while a
+        refresh is already running returns 409 without querying the database.
+
         Returns 202 immediately and runs the refresh in a background thread.
         Returns 409 if a refresh is already in progress.
-        Returns 400 if cve_ids is empty or contains no valid CVE identifiers.
+        Returns 400 if cve_ids is empty, contains no valid CVE identifiers, or
+        none of the submitted CVEs exist in the database.
 
         OpenAPI:
         body JsonObject optional JSON body containing cve_ids.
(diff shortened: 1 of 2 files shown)


src/routes/bulk_refresh.py
```diff
@@ -239,8 +239,9 @@ def bulk_epss_refresh() -> ResponseReturnValue:
         cve_ids = [c for c in cve_ids if _CVE_RE.match(c)]
         if not cve_ids:
             return jsonify({"error": "cve_ids must contain valid CVE identifiers (e.g. CVE-2024-1234)"}), 400
-        if len(cve_ids) > _MAX_CVE_IDS:
-            return jsonify({"error": f"cve_ids must contain at most {_MAX_CVE_IDS} entries"}), 400
+        # No hard cap here: the refresh below chunks the CVE list into batches
```
sfl-pe-sync: ERROR: Removing the only CVE-count bound makes this unauthenticated, permissive-CORS endpoint able to enqueue work proportional to the entire JSON body. `_EPSS_BATCH_SIZE` limits each FIRST.org request, not aggregate work: the app-wide request limit is about 5 MiB, so one request can contain hundreds of thousands of valid or repeated IDs and keep the singleton refresh occupied for thousands of sequential API calls (each with a 30-second timeout), preventing legitimate refreshes. Preserve a server-side work bound; to support large databases, deduplicate and derive/validate the IDs against database records on the server, or process bounded pages/jobs rather than accepting an arbitrary client-supplied list without a cap.

sfl-pe-sync (changes requested):
The intended >1000-CVE refresh is not safe as implemented because removing the aggregate work bound exposes a long-lived background-job denial of service despite per-request batching.

src/routes/bulk_refresh.py
```diff
@@ -239,8 +244,10 @@ def bulk_epss_refresh() -> ResponseReturnValue:
         cve_ids = [c for c in cve_ids if _CVE_RE.match(c)]
         if not cve_ids:
             return jsonify({"error": "cve_ids must contain valid CVE identifiers (e.g. CVE-2024-1234)"}), 400
-        if len(cve_ids) > _MAX_CVE_IDS:
-            return jsonify({"error": f"cve_ids must contain at most {_MAX_CVE_IDS} entries"}), 400
+        known_ids = {row[0] for row in db.session.query(Vulnerability.id).all()}
```
sfl-pe-sync: ERROR: This replaces the request-count amplification with an unbounded full-table read before the singleton guard: even a one-ID request executes `.all()` for every vulnerability and materializes the entire ID set, and requests arriving while a refresh is already running repeat that scan before returning 409. Because this endpoint is unauthenticated and CORS-enabled, concurrent tiny requests can multiply database work and Python memory by the total vulnerability count, causing worker/SQLite contention on the large databases this change targets. Atomically reserve/check `EPSSProgressTracker` before database work, and avoid loading the whole table in the request handler (for example, run a server-derived, streamed/chunked CVE query in the reserved background job, or intersect through bounded indexed queries); add a regression test proving a busy request does not query vulnerabilities.

sfl-pe-sync (changes requested):
The known-CVE intersection addresses the previously reported unbounded background workload, but it introduces a blocking unauthenticated full-table-read amplification before the refresh lock. CI checks were still in progress, and no local commands were run in this review session.

src/routes/bulk_refresh.py
ValentinBoudevin: Fixed

src/routes/bulk_refresh.py
```diff
@@ -235,16 +247,38 @@ def bulk_epss_refresh() -> ResponseReturnValue:
...
+        # time. Bound the work to CVEs that actually exist by intersecting the
+        # requested IDs with the database through chunked, primary-key-indexed
+        # ``IN`` lookups. Cost is bounded by the (deduplicated) request size, not
+        # the size of the whole vulnerabilities table, so it scales to the large
+        # databases this endpoint targets. Release the reservation if nothing
+        # remains so a later request is not wrongly rejected with 409.
+        known_ids: set[str] = set()
+        for i in range(0, len(cve_ids), _DB_LOOKUP_CHUNK):
+            chunk = cve_ids[i:i + _DB_LOOKUP_CHUNK]
+            rows = db.session.query(Vulnerability.id).filter(Vulnerability.id.in_(chunk)).all()
```
sfl-pe-sync: ERROR: The tracker has already been reserved when this new database lookup runs, but an `OperationalError` or any other query failure escapes without calling `complete()`/`error()`. The request returns 500 while `in_progress` remains true indefinitely, so every later EPSS refresh is rejected with 409 until the process restarts. Wrap the intersection in an exception path that rolls back the session and releases/marks the tracker as errored before returning or re-raising, and add a regression test for a failing lookup.

sfl-pe-sync (changes requested):
The revised patch addresses the previously reported busy-request and client-controlled background-work issues, but a database failure during the newly introduced preflight lookup permanently wedges the singleton tracker and blocks all subsequent EPSS refreshes.

src/routes/bulk_refresh.py
ValentinBoudevin: Fixed

src/routes/bulk_refresh.py
```diff
@@ -235,16 +247,47 @@ def bulk_epss_refresh() -> ResponseReturnValue:
...
+        try:
+            for i in range(0, len(cve_ids), _DB_LOOKUP_CHUNK):
+                chunk = cve_ids[i:i + _DB_LOOKUP_CHUNK]
+                rows = db.session.query(Vulnerability.id).filter(Vulnerability.id.in_(chunk)).all()
+                known_ids.update(row[0] for row in rows)
+        except Exception as exc:
+            # The tracker is already reserved, so a failing lookup must release
+            # it; otherwise ``in_progress`` would stay true and every later EPSS
+            # refresh would be rejected with 409 until the process restarts.
+            db.session.rollback()
```
sfl-pe-sync, line 281: WARNING: Storing `str(exc)` in the progress tracker can expose SQLAlchemy query text, parameters, database paths, or driver details through the EPSS progress/status response after a lookup failure; the immediate 500 response is generic, but the tracker state is not. Log the detailed exception server-side and pass a fixed, user-safe message such as `Failed to look up CVEs in the database` to `EPSSProgressTracker.error()`.

sfl-pe-sync (approved):
The bulk-refresh flow now reserves the singleton and limits refresh work to known, deduplicated CVEs, but the new lookup-failure path can disclose internal database error details through tracker status.

reh-sfl (approved):

PR #516, opened 2026-08-11
feat(exports): add guided export workflow

reh-sfl:

* Add guided export workflow
* Move Custom Reports to Settings tab
* Add exports as an async queue job

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Open web dashboard, go to the Export tab and follow the new workflow, observe that it creates a new job in the Operations Queue.
Go to Settings and observe the new Custom reports tab.

## Pull Request Checklist

### DIFF
--- a/doc/source/api.md
+++ b/doc/source/api.md
@@ -690,6 +690,34 @@ Generate a report from a template or export an SBOM in the requested format.
 
 **Response:** File download with appropriate `Content-Type` and `Content-Disposition` headers. Supports JSON, XML, PDF, and HTML outputs depending on the template.
 
+### Export Multiple Documents
+
+```
+POST /api/documents/export
+```
+
+Renders selected document formats for every variant in one project. `consolidated` creates one project-wide file per selection; `per_variant` creates one file per selection and variant. Both modes return a ZIP archive.
+
+**Request body:**
+
+```json
+{
+  "project_id": "11111111-1111-1111-1111-111111111111",
+  "variant_ids": [
+    "22222222-2222-2222-2222-222222222222"
+  ],
+  "mode": "per_variant",
+  "documents": [
+    { "name": "summary.adoc", "extension": "pdf" },
+    { "name": "CycloneDX 1.6", "extension": "json" }
+  ]
+}
+```
+
+`variant_ids` selects one or more variants belonging to the project. When omitted, all project variants are exported. Reports support either output mode. SBOM selections require `per_variant` mode and cannot be mixed with reports in the same request. Document and extension pairs must match non-asset entries returned by `GET /api/documents`. Custom assets are resources used by report templates and cannot be exported directly.
+
+**Response:** ZIP file download (`Content-Type: application/zip`).
+
 ---
 
 ## Scans
--- a/frontend/src/handlers/exportQueue.ts
+++ b/frontend/src/handlers/exportQueue.ts
@@ -1,0 +1,144 @@
+import type { ScanEntryState, ScanManagerSnapshot } from "./scanStateManager";
+
+export type ExportRequest = {
+    project_id: string;
+    variant_ids: string[];
+    mode: "consolidated" | "per_variant";
+    documents: Array<{ name: string; extension: string }>;
+};
+
+type ExportStatus = {
+    status: "running" | "done" | "error";
+    current: number;
+    total: number;
+    progress: string;
+    logs: string[];
+    error: string | null;
+};
+
+const POLL_INTERVAL_MS = 1000;
+const MAX_POLL_FAILURES = 3;
+const states = new Map<string, ScanEntryState>();
+const listeners = new Set<() => void>();
+let snapshot: ScanManagerSnapshot = [];
+let nextOperationId = 1;
+
+function publish() {
+    snapshot = [...states.values()];
+    listeners.forEach(listener => listener());
+}
+
+function downloadBlob(blob: Blob, filename: string) {
+    const url = URL.createObjectURL(blob);
+    const anchor = document.createElement("a");
+    anchor.href = url;
+    anchor.download = filename;
+    document.body.appendChild(anchor);
+    anchor.click();
+    anchor.remove();
+    URL.revokeObjectURL(url);
+}
+
+function responseFilename(response: Response, fallback: string): string {
+    const disposition = response.headers.get("Content-Disposition") ?? "";
+    return disposition.match(/filename="?([^";]+)/)?.[1] ?? fallback;
+}
+
+function updateEntry(localId: string, update: Partial<ScanEntryState>) {
+    const current = states.get(localId);
+    if (!current) return;
+    states.set(localId, { ...current, ...update });
+    publish();
+}
+
+async function pollExport(
+    localId: string,
+    jobId: string,
+    fallbackFilename: string,
+    failedAttempts = 0,
+): Promise<void> {
+    try {
+        const response = await fetch(`${import.meta.env.VITE_API_URL}/api/documents/export/${jobId}`, { mode: "cors" });
+        if (!response.ok) throw new Error(`Failed to check export progress (${response.status})`);
+        const status = await response.json() as ExportStatus;
+        updateEntry(localId, {
+            status: status.status === "error" ? "error" : status.status === "done" ? "done" : "running",
+            error: status.error,
+            progress: status.progress,
+            logs: status.logs,
+            total: status.total,
+            doneCount: status.status === "done" ? status.total : Math.max(0, status.current - 1),
+        });
+
+        if (status.status === "running") {
+            setTimeout(() => void pollExport(localId, jobId, fallbackFilename), POLL_INTERVAL_MS);
+            return;
+        }
+        if (status.status === "error") return;
+
+        const download = await fetch(`${import.meta.env.VITE_API_URL}/api/documents/export/${jobId}/download`, { mode: "cors" });
+        if (!download.ok) throw new Error(`Failed to download export (${download.status})`);
+        downloadBlob(await download.blob(), responseFilename(download, fallbackFilename));
+    } catch (reason) {
+        const message = reason instanceof Error ? reason.message : String(reason);
+        if (failedAttempts < MAX_POLL_FAILURES) {
+            updateEntry(localId, { status: "running", error: null, progress: "Reconnecting to export", logs: [message] });
+            setTimeout(
+                () => void pollExport(localId, jobId, fallbackFilename, failedAttempts + 1),
+                POLL_INTERVAL_MS,
+            );
+            return;
+        }
+        updateEntry(localId, { status: "error", error: message, progress: "Export failed", logs: [message] });
+    }
+}
+
+export const subscribe = (listener: () => void): (() => void) => {
+    listeners.add(listener);
+    return () => listeners.delete(listener);
+};
+
+export const getSnapshot = (): ScanManagerSnapshot => snapshot;
+
+export const dismiss = (localId: string) => {
+    const state = states.get(localId);
+    if (!state || state.status === "running" || state.status === "queued") return;
+    states.delete(localId);
+    publish();
+};
+
+export async function queueExport(request: ExportRequest, projectName: string): Promise<void> {
+    const localId = `export-${nextOperationId++}`;
+    const scopeCount = request.mode === "per_variant" ? request.variant_ids.length : 1;
+    states.set(localId, {
+        variantId: localId,
+        variantName: projectName,
+        status: "running",
+        error: null,
+        progress: "Queued",
+        logs: [],
+        total: scopeCount * request.documents.length,
+        doneCount: 0,
+    });
+    publish();
+
+    try {
+        const response = await fetch(import.meta.env.VITE_API_URL + "/api/documents/export", {
+            method: "POST",
+            mode: "cors",
+            headers: { "Content-Type": "application/json" },
+            body: JSON.stringify({ ...request, async: true }),
+        });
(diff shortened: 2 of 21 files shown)


src/routes/documents.py
```diff
@@ -319,6 +442,86 @@ def upload_asset() -> ResponseReturnValue:
...
+                for variant, scope in scopes:
+                    for doc_name, extension in selections:
+                        rendered = _render_document(app, doc_name, extension, scope)
+                        if rendered.status_code >= 400:
+                            message = rendered.get_json(silent=True) or {}
+                            return {"error": message.get("error", f"Failed to export {doc_name}")}, rendered.status_code
+                        fallback = f"{_safe_archive_name(doc_name)}.{_safe_archive_name(extension)}"
+                        filename = _download_filename(rendered, fallback)
+                        if variant is not None:
+                            filename = f"{_safe_archive_name(variant.name)}/{filename}"
```
sfl-pe-sync: ERROR: Archive entry paths can collide after sanitization: distinct variant names such as `prod us` and `prod_us` produce the same folder, and different selected documents may also render the same filename. `ZipFile.writestr()` permits duplicate names, so common extractors overwrite one export and return an incomplete result. Track generated archive paths and deterministically disambiguate or reject collisions (ideally include a variant UUID/document identifier).

frontend/src/pages/Exports.tsx
```diff
@@ -1,351 +1,82 @@
...
-        const extension = files[0].name.split(".").pop()?.toLowerCase() ?? "";
-        if (tab === "assets" || (tab === "all" && assetExtensions.has(extension))) {
-                uploadAsset(files[0]);
-            } else {
-                uploadTemplate(files[0]);
-            }
+        if (!projectId) {
+            setVariants([]);
+            return;
         }
```
sfl-pe-sync, line 55: WARNING: This request is not tied to the `projectId` that initiated it. If project A's slower response resolves after switching to project B, it overwrites `variants` with A's variants while the wizard submits B's project ID; the backend then rejects every export until another refresh. Clear variants on each project change and abort stale requests or guard the resolution against the current project ID.

sfl-pe-sync (changes requested):
The guided workflow has a blocking archive-path collision that can silently omit selected outputs, plus a frontend request-ordering issue when switching projects quickly.

src/routes/documents.py
```diff
@@ -319,6 +555,142 @@ def upload_asset() -> ResponseReturnValue:
...
+                    app.logger.exception("Export job %s failed", job_id)
+                    with _export_jobs_lock:
+                        _export_jobs[job_id].update({
+                            "status": "error",
+                            "error": str(exc),
+                            "progress": "Export failed",
+                            "finished_at": time.monotonic(),
+                        })
+
+            threading.Thread(target=generate_archive, name=f"document-export-{job_id}", daemon=True).start()
```
sfl-pe-sync: ERROR: Every async request starts an unbounded daemon thread, and each completed job retains the entire ZIP as a `bytes` object for up to an hour. Repeated requests can therefore bypass the web server's worker limits and exhaust CPU and memory, especially for PDF/report exports. Run exports through a bounded worker queue, reject or defer work when the queue is full, spool archives to size-limited temporary files/object storage, and guarantee cleanup.

src/routes/documents.py
```diff
@@ -319,6 +555,142 @@ def upload_asset() -> ResponseReturnValue:
...
+                            "error": message,
+                            "progress": "Export failed",
+                            "finished_at": time.monotonic(),
+                        })
+                except Exception as exc:
+                    app.logger.exception("Export job %s failed", job_id)
+                    with _export_jobs_lock:
+                        _export_jobs[job_id].update({
+                            "status": "error",
+                            "error": str(exc),
```
sfl-pe-sync: WARNING: Unexpected exception text is stored verbatim and returned by the public status endpoint, so converter, filesystem, database, or template failures can disclose internal paths or SQL details. Keep `app.logger.exception(...)` for diagnostics, but expose a generic error message and a stable error code to clients.

frontend/src/handlers/exportQueue.ts
```diff
@@ -0,0 +1,130 @@
...
+            status: status.status === "error" ? "error" : status.status === "done" ? "done" : "running",
+            error: status.error,
+            progress: status.progress,
+            logs: status.logs,
+            total: status.total,
+            doneCount: status.status === "done" ? status.total : Math.max(0, status.current - 1),
+        });
+
+        if (status.status === "running") {
+            setTimeout(() => void pollExport(localId, jobId, fallbackFilename), POLL_INTERVAL_MS);
```
sfl-pe-sync, line 74: WARNING: Any single failed status request immediately marks the export failed and permanently stops polling, while the backend job continues and its result becomes unreachable from this client. Use bounded consecutive-failure retries/backoff (as `scanStateManager` already does) and only transition to `error` after the retry budget is exhausted.

sfl-pe-sync (changes requested):
The guided workflow is not ready to merge: archive-path collisions can silently lose selected output, and the async implementation permits unbounded thread and memory consumption. The status path also leaks internal exception details and abandons jobs after one transient polling failure; frontend CI currently misses its coverage gate.

frontend/src/handlers/exportQueue.ts
```diff
@@ -0,0 +1,130 @@
...
+        if (!response.ok || typeof body.job_id !== "string") {
+            throw new Error(body.error || `Export failed (${response.status})`);
+        }
+        const suffix = request.mode === "consolidated" ? "consolidated" : "by_variant";
+        const fallbackFilename = `${projectName}_${suffix}_export.zip`;
+        void pollExport(localId, body.job_id, fallbackFilename);
+    } catch (reason) {
+        const message = reason instanceof Error ? reason.message : String(reason);
+        updateEntry(localId, { status: "error", error: message, progress: "Export failed", logs: [message] });
+        throw reason;
```
sfl-pe-sync, line 142: WARNING: The job ID exists only in this in-memory polling closure. Reloading the page resets the module state while the backend export continues, so the completed archive cannot be rediscovered or downloaded and remains on disk until pruning. Persist active job IDs and restore polling after reload, or add a backend endpoint that lists recoverable jobs for the client.

sfl-pe-sync (changes requested):
The guided workflow is not ready to merge: archive-name collisions can lose outputs, and both synchronous exports and retained completed jobs still permit unbounded aggregate resource use. Job recovery and server-path disclosure also need correction.

frontend/src/handlers/exportQueue.ts
```diff
@@ -0,0 +1,144 @@
...
+            doneCount: status.status === "done" ? status.total : Math.max(0, status.current - 1),
+        });
+
+        if (status.status === "running") {
+            setTimeout(() => void pollExport(localId, jobId, fallbackFilename), POLL_INTERVAL_MS);
+            return;
+        }
+        if (status.status === "error") return;
+
+        const download = await fetch(`${import.meta.env.VITE_API_URL}/api/documents/export/${jobId}/download`, { mode: "cors" });
```
sfl-pe-sync, line 79: ERROR: The download retry path cannot recover from an interrupted response after the server starts sending the archive: `/download` removes the job and unlinks its file before `send_file` finishes, so `blob()` can reject after the only copy has already been consumed and every retry then receives 404. Keep completed archives available until TTL/explicit acknowledgement (and make download idempotent), then retry the download directly without destroying the job on first access.

sfl-pe-sync (changes requested):
The guided export workflow still has a blocking one-shot download failure mode: a transient interruption can permanently discard a completed archive despite the new retry logic.

ValentinBoudevin (changes requested):
1- Bug when selecting SBOM

Image

Also why do I have to check JSON as a second checkbox as , except for SPDX2, it is the only mandatory option.

Same for the reports, a lot of them only support one output format

2-Frontend, can we make the menu bigger? Right not it only take a part of the webpage and font as very small

3- Can we keep the drag-and-drop feature for custom report (in addition to the one in Settings tab)

ValentinBoudevin (dismissed):
Nothing changed since last review

src/routes/documents.py
```diff
@@ -319,6 +644,178 @@ def upload_asset() -> ResponseReturnValue:
 
         return {"name": safe_name, "message": "Asset uploaded."}, 201
 
+    @app.route('/api/documents/export', methods=['POST'])
+    def export_documents() -> ResponseReturnValue:
+        """Render selected documents for every variant of one project."""
+        body = request.get_json(silent=True) or {}
```
sfl-pe-sync, line 650: WARNING: A truthy JSON value that is not an object (for example, `["x"]`) reaches the following `body.get(...)` calls and raises `AttributeError`, turning invalid client input into a 500 response. Validate that `request.get_json()` returned a dictionary and return 400 before reading request fields.

src/routes/documents.py
```diff
@@ -180,11 +201,343 @@ def guess_mime_type(doc_name: Optional[str]) -> Optional[str]:
...
+        value = item[1].get("finished_at")
+        return value if isinstance(value, float) else 0.0
+
+    retained.sort(key=finished_at)
+    retained_bytes = sum(
+        os.path.getsize(str(job["archive_path"]))
+        for _, job in retained
+        if os.path.isfile(str(job["archive_path"]))
+    )
+    for job_id, job in retained:
```
sfl-pe-sync, line 475: WARNING: Retention enforcement can evict and unlink an older completed archive as soon as another job finishes, even though the older job is still within its TTL and its client has not downloaded it. That client can observe `done` and then immediately receive 404. Reserve storage before generation or fail the newly completing job rather than deleting an undownloaded completed result.

src/routes/documents.py
```diff
@@ -319,6 +644,178 @@ def upload_asset() -> ResponseReturnValue:
...
+                future = _export_executor.submit(generate_archive)
+            except RuntimeError:
+                with _export_jobs_lock:
+                    del _export_jobs[job_id]
+                _export_capacity.release()
+                return {"error": "Export service is unavailable"}, 503
+            future.add_done_callback(lambda _future: _export_capacity.release())
+            return {"job_id": job_id}, 202
+
+        if not _sync_export_capacity.acquire(blocking=False):
```
sfl-pe-sync, line 766: WARNING: Synchronous exports use `_sync_export_capacity` while asynchronous work is bounded independently by `_export_capacity` and the executor, so two sync requests can run alongside two async workers and double the intended expensive-render concurrency. Use one shared execution semaphore or route both modes through the same bounded executor.

frontend/src/handlers/exportQueue.ts
```diff
@@ -0,0 +1,144 @@
...
+    localId: string,
+    jobId: string,
+    fallbackFilename: string,
+    failedAttempts = 0,
+): Promise<void> {
+    try {
+        const response = await fetch(`${import.meta.env.VITE_API_URL}/api/documents/export/${jobId}`, { mode: "cors" });
+        if (!response.ok) throw new Error(`Failed to check export progress (${response.status})`);
+        const status = await response.json() as ExportStatus;
+        updateEntry(localId, {
```
sfl-pe-sync, line 64: WARNING: The queue entry becomes `done` and dismissible before the archive fetch and `blob()` conversion complete. If the user dismisses it during a slow transfer, any subsequent download failure cannot update the removed entry and retries continue invisibly. Keep the entry running/downloading until the blob has been received successfully, then mark it done.

sfl-pe-sync (changes requested):
The guided export workflow has several reliability and validation gaps: malformed JSON can cause a 500, completed archives can be evicted before retrieval, rendering concurrency exceeds its stated bound, and the UI reports completion before download finishes.

ValentinBoudevinSFL (changes requested):
1- Select Clear/Clear Visible buttons are not UI efficient

Image

Maybe it should be in a button, or increase the spacing but it looks a bit confusing

2- The colors don't match the other tab and VulnScout DA

3- Choose the output layout, does it need to be a third step menu "3.Layout"

Image

Can´t  just it be an hidden fields in "2.Export type" which is set to visble when the user select the type to "Reports"

4- (Optional) I belive we don´t need a scroll-down menu for the SBOM

Image

It is only a fixed list, maybe buttons are better

ValentinBoudevinSFL (approved):
LGTM but this ticket needs a part 2 with CLI integration

doc/source/api.md
```diff
@@ -690,6 +690,34 @@ Generate a report from a template or export an SBOM in the requested format.
...
+    { "name": "summary.adoc", "extension": "pdf" },
+    { "name": "CycloneDX 1.6", "extension": "json" }
+  ]
+}
+```
+
+`variant_ids` selects one or more variants belonging to the project. When omitted, all project variants are exported. Reports support either output mode. SBOM selections require `per_variant` mode and cannot be mixed with reports in the same request. Document and extension pairs must match non-asset entries returned by `GET /api/documents`. Custom assets are resources used by report templates and cannot be exported directly.
+
+**Response:** ZIP file download (`Content-Type: application/zip`).
+
```
sfl-pe-sync, line 720: WARNING: The documented contract only describes the synchronous ZIP response, but the new production workflow sends `"async": true` and depends on the 202 `job_id`, status polling endpoint, and download endpoint. API consumers following this documentation cannot implement or recover the advertised queued workflow. Document the async request flag, 202/status payloads, polling and download routes, terminal/error states, and retention semantics.

sfl-pe-sync (approved):
The guided export implementation is generally coherent, but the public API documentation omits the new asynchronous queue contract used by the frontend.

PR #517, opened 2026-08-12
Add initial setup assistant popup

reh-sfl:

When the user first launches Vulnscout we tell them how to work with adding project / variant to their instance.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Launch Vulnscout with a fresh database (e.g change vulnscout.db to vulnscout.db.old) then open the web dashboard, you should be presented with a popup instructing you how to start your first project / variant.

## Pull Request Checklist

### DIFF
--- a/frontend/src/pages/Explorer.tsx
+++ b/frontend/src/pages/Explorer.tsx
@@ -2,6 +2,7 @@ import { useState, useEffect, useCallback, useRef, useSyncExternalStore } from "
 import NavigationBar from "../components/NavigationBar";
 import OperationQueueModal from "../components/OperationQueueModal";
 import MessageBanner from "../components/MessageBanner";
+import Popup from "../components/Popup";
 import type { Package } from "../handlers/packages";
 import type { CVSS, Vulnerability } from "../handlers/vulnerabilities";
 import type { Assessment } from "../handlers/assessments";
@@ -73,6 +74,12 @@ function Explorer() {
     const [currentVariantIds, setCurrentVariantIds] = useState<string[] | undefined>(undefined);
     const [currentMultiOperation, setCurrentMultiOperation] = useState<string | undefined>(undefined);
     const [operationQueueOpen, setOperationQueueOpen] = useState(false);
+    const [setupRequirement, setSetupRequirement] = useState<
+        { kind: 'project' } | { kind: 'variant'; projectId: string } | null
+    >(null);
+    const [settingsDestination, setSettingsDestination] = useState<
+        { tab: 'projects'; projectId?: string } | null
+    >(null);
     const hadActiveScans = useRef(false);
     const grypeScanEntries = useSyncExternalStore(grypeSubscribe, grypeGetSnapshot);
     const nvdScanEntries = useSyncExternalStore(nvdSubscribe, nvdGetSnapshot);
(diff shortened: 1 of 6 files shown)


frontend/src/pages/Explorer.tsx
```diff
@@ -109,6 +116,24 @@ function Explorer() {
         return () => { cancelled = true; };
     }, []);
 
+    const loadSetupRequirement = useCallback(() => {
+        return Promise.all([Projects.list(), Variants.listAll()])
```
sfl-pe-sync, line 122: WARNING: `Projects.list()` converts every non-2xx response into `[]`, so a transient backend/auth failure is indistinguishable from a fresh database and this code displays “Add your first project” instead of surfacing the load failure. The trailing `catch` cannot handle that case because the promise already resolved successfully. Make the project listing reject on non-OK responses (as `Variants.listAll()` does) and show/retry an error state rather than treating request failure as empty setup data.

frontend/src/pages/Explorer.tsx
```diff
@@ -481,12 +536,13 @@ function Explorer() {
                 {tab === 'scans' && <ScanHistory variantId={currentVariantId} projectId={currentVariantId ? undefined : currentProjectId} onScanComplete={handleScanComplete} />}
                 {tab === 'review' && <Review variantId={currentVariantId} projectId={currentVariantId ? undefined : currentProjectId} onAssessmentChanged={handleAssessmentChanged} />}
                 {tab === 'exports' && <Exports variantId={currentVariantId} projectId={currentProjectId} variantIds={currentVariantIds} />}
-                {tab === 'settings' && <Settings onDataChanged={(message) => {
+                {tab === 'settings' && <Settings initialTab={settingsDestination?.tab} onDataChanged={(message) => {
                     if (message) setLoadingMessage(message);
                     Config.get().then(config => setDefaultConfig(config)).catch(() => {});
+                    loadSetupRequirement();
```
sfl-pe-sync, line 559: WARNING: Every settings mutation starts an untracked setup check, so consecutive creation of the first project and its first variant can produce overlapping requests; if the earlier post-project request resolves last, it overwrites the newer `null` result and the user is incorrectly prompted to add a variant that already exists. Cancel or generation-guard prior requests before applying `setSetupRequirement`, and add an out-of-order-response regression test.

sfl-pe-sync (changes requested):
The setup flow can misclassify API failures as an empty installation and can be left stale by out-of-order refreshes, so the new prompt is not yet reliable.

frontend/tests/unit_tests/test_explorer_scope_restore.tsx
```diff
@@ -320,6 +327,91 @@ describe('Explorer saved-scope validation', () => {
...
+            .mockResolvedValueOnce([{ id: 'variant-1', name: 'Variant 1', project_id: 'project-1' }])
+            .mockResolvedValueOnce([])
+            .mockResolvedValueOnce([{ id: 'variant-1', name: 'Variant 1', project_id: 'project-1' }]);
+
+        render(<Explorer />);
+        await waitFor(() => expect(mockProjectsList).toHaveBeenCalled());
+        fireEvent.click(screen.getByRole('button', { name: 'settings' }));
+        fireEvent.click(screen.getByRole('button', { name: 'settings changed' }));
+        fireEvent.click(screen.getByRole('button', { name: 'settings changed' }));
+        await waitFor(() => expect(mockProjectsList).toHaveBeenCalledTimes(3));
```
sfl-pe-sync, line 409: WARNING: This assertion runs while the active tab is still Settings, but the popup is rendered only when `tab === 'metrics'`, so it passes even if the stale response has incorrectly restored a setup requirement. In fact, the newest mocked variants response is `[]`, which would produce the variant prompt. Navigate back to Metrics before asserting and arrange the older request to report no variants while the newer request reports an existing variant, so the test fails without the generation guard.

sfl-pe-sync (changes requested):
The setup race guard is implemented, but its new regression test cannot observe the popup state and therefore does not protect the behavior it claims to cover.

frontend/src/pages/Explorer.tsx
```diff
@@ -414,6 +447,45 @@ function Explorer() {
                 />
```
sfl-pe-sync, line 447: WARNING: This first-run dialog opens asynchronously without moving focus into it, trapping focus, or giving the dialog an accessible name; keyboard and screen-reader users can remain focused on and navigate the obscured dashboard without being told that setup is required. Extend `Popup` to associate the heading via `aria-labelledby`, focus an element inside the dialog when it opens, contain Tab navigation, and restore the prior focus on close.

sfl-pe-sync (dismissed):
The setup flow is functionally coherent, but the newly automatic first-run modal is not keyboard- or screen-reader-accessible and should be corrected before merge.

ValentinBoudevin (changes requested):
LGTM

ValentinBoudevin (approved):

PR #521, opened 2026-08-12
Add stable review export updates

reh-sfl:

* Add Normal export and Append/update existing file modes to the Review export flow.
* Detect uploaded export formats and reconcile selected variants while preserving stable content for focused Git diffs.
* Remove variants that exist in the uploaded file but are no longer selected, with validation for malformed or unsupported files.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

* Run `python -m pytest tests/unit_tests/test_assessment_io_coverage.py tests/webapp_tests/test_review_endpoints.py -q`.
* Run `npm --prefix frontend test -- --runInBand tests/unit_tests/test_export_json.ts tests/unit_tests/test_review.tsx tests/unit_tests/test_review_transfer_modal.tsx --coverage=false`.
* Run `npm --prefix frontend run build`.
* In Review, export normally, then choose Append/update existing file, upload that export, change the variant selection, and confirm the downloaded file updates selected variants and removes unselected ones with a focused Git diff.

### Additional notes

The update path preserves stable OpenVEX identifiers, authors, ordering, and unchanged content where the format permits.

### Related Issue

No linked issue.

## Pull Request Checklist

### DIFF
--- a/frontend/src/helpers/exportJson.ts
+++ b/frontend/src/helpers/exportJson.ts
@@ -19,6 +19,28 @@ export function formatTimestampForFilename(date?: Date | string): string {
     return `${d.getFullYear()}${pad(d.getMonth() + 1)}${pad(d.getDate())}_${pad(d.getHours())}${pad(d.getMinutes())}${pad(d.getSeconds())}`;
 }
 
+export type ReviewExportFormat = 'custom' | 'openvex';
+
+/** Detect a supported Review export format from parsed JSON. */
+export function detectReviewExportFormat(data: unknown): ReviewExportFormat {
+    if (!data || typeof data !== 'object' || Array.isArray(data)) {
+        throw new Error('Export file must contain a JSON object.');
+    }
+    const document = data as Record<string, unknown>;
+    if (String(document['@context'] ?? '').includes('openvex') && Array.isArray(document.statements)) {
+        return 'openvex';
+    }
+    if (document.version === 1 && Array.isArray(document.assessments)) {
+        for (const section of ['ai_assessments', 'cvss', 'time_estimates']) {
+            if (document[section] !== undefined && !Array.isArray(document[section])) {
+                throw new Error(`Invalid VulnScout JSON: '${section}' must be an array.`);
+            }
+        }
+        return 'custom';
+    }
+    throw new Error('Unsupported export format. Expected VulnScout JSON or OpenVEX.');
+}
+
 /**
  * Trigger a JSON file download in the browser.
  */
--- a/frontend/src/pages/Review.tsx
+++ b/frontend/src/pages/Review.tsx
@@ -11,7 +11,7 @@ import FilterOption from "../components/FilterOption";
 import ToggleSwitch from "../components/ToggleSwitch";
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faCircleQuestion, faCircleInfo, faFileExport, faFileImport, faPenToSquare, faTrash, faBook, faCheck, faXmark } from '@fortawesome/free-solid-svg-icons';
-import { downloadJson, sanitizeFilename, formatTimestampForFilename } from '../helpers/exportJson';
+import { detectReviewExportFormat, downloadJson, sanitizeFilename, formatTimestampForFilename } from '../helpers/exportJson';
 import EditAssessment from '../components/EditAssessment';
 import type { EditAssessmentData } from '../components/EditAssessment';
 import type { Variant } from '../handlers/variant';
@@ -199,6 +199,9 @@ function Review({ variantId, projectId, onAssessmentChanged }: Readonly<Props>) 
     const [transferVariantIds, setTransferVariantIds] = useState<string[]>([]);
     const [transferFormat, setTransferFormat] = useState<'custom' | 'openvex'>('custom');
     const [importTimestampPolicy, setImportTimestampPolicy] = useState<'original' | 'current'>('original');
+    const [exportMode, setExportMode] = useState<'normal' | 'update'>('normal');
+    const [existingExportFile, setExistingExportFile] = useState<File | undefined>();
+    const [existingExportError, setExistingExportError] = useState<string | undefined>();
 
     const showMessage = useCallback((message: string, type: "error" | "success") => {
         setBannerMessage(message);
@@ -464,6 +467,9 @@ function Review({ variantId, projectId, onAssessmentChanged }: Readonly<Props>) 
     const openTransfer = useCallback((mode: 'import' | 'export') => {
         setTransferFormat('custom');
         setImportTimestampPolicy('original');
+        setExportMode('normal');
+        setExistingExportFile(undefined);
+        setExistingExportError(undefined);
         setTransferVariantIds(mode === 'export'
             ? variantId ? [variantId] : transferVariants.map(variant => variant.id)
             : []);
@@ -484,8 +490,43 @@ function Review({ variantId, projectId, onAssessmentChanged }: Readonly<Props>) 
         setTransferVariantIds(defaultVariantId ? [defaultVariantId] : []);
     }, [transferMode, transferVariants, variantId]);
 
+    const changeExportMode = useCallback((mode: 'normal' | 'update') => {
+        setExportMode(mode);
+        setExistingExportFile(undefined);
+        setExistingExportError(undefined);
+        if (mode === 'normal') changeTransferFormat('custom');
+    }, [changeTransferFormat]);
+
+    const existingExportSelection = useRef(0);
+    const handleExistingExportFile = useCallback(async (file?: File) => {
+        const selection = ++existingExportSelection.current;
+        setExistingExportFile(undefined);
+        setExistingExportError(undefined);
+        if (!file) return;
+        try {
+            const text = await new Promise<string>((resolve, reject) => {
+                const reader = new FileReader();
+                reader.onload = () => resolve(String(reader.result ?? ''));
+                reader.onerror = () => reject(new Error('Unable to read export file.'));
+                reader.readAsText(file);
+            });
+            if (selection !== existingExportSelection.current) return;
+            const parsed = JSON.parse(text);
+            const detectedFormat = detectReviewExportFormat(parsed);
+            changeTransferFormat(detectedFormat);
+            setExistingExportFile(file);
+        } catch (error) {
+            if (selection !== existingExportSelection.current) return;
+            setExistingExportError(error instanceof Error ? error.message : 'Unable to read export file.');
+        }
+    }, [changeTransferFormat]);
+
     const handleExportReview = useCallback(async () => {
         try {
+            if (exportMode === 'update' && !existingExportFile) {
+                showMessage('Choose a supported existing export file.', 'error');
+                return;
+            }
             const params = new URLSearchParams();
             transferVariantIds.forEach(variantId => params.append('variant_id', variantId));
             const endpoint = transferFormat === 'openvex' ? 'export' : 'export-custom-data';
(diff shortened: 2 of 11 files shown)


src/routes/assessments.py
```diff
@@ -807,6 +809,81 @@ def export_review_custom_data() -> ResponseReturnValue:
...
+            variant = DBVariant.get_by_id(variant_id)
+            if variant is None:
+                return {"error": f"Variant not found: {raw_variant_id}"}, 404
+            variant_ids.append(variant_id)
+
+        if export_format == 'openvex':
+            if len(variant_ids) != 1:
+                return {"error": "Exactly one variant_id is required for OpenVEX export"}, 400
+            handmade = DBAssessment.get_by_origin(variant_ids, origin="custom")
+            if not handmade:
```
sfl-pe-sync: ERROR: Rejecting an empty current export prevents the update operation from removing stale data. If the last handmade assessment (or all custom sections) was deleted, this branch returns 404 before reconciliation, so an existing file can never be updated to empty and continues to report deleted/unselected records. Build the empty document and pass it to `reconcile_review_export`; an empty `statements`/section set is a valid update result. Apply the same change to the custom-export emptiness check below.

src/helpers/assessment_io.py
```diff
@@ -164,6 +165,115 @@ def is_openvex_doc(doc: object) -> bool:
...
+    reconciled.extend(record for index, record in enumerate(current) if index not in consumed)
+    return reconciled
+
+
+def reconcile_review_export(existing: object, current: dict[str, Any]) -> dict[str, Any]:
+    """Update a Review export while minimizing ordering and content churn."""
+    export_format = detect_review_export_format(existing)
+    if detect_review_export_format(current) != export_format:
+        raise ValueError("Existing file format does not match the generated export")
+    assert isinstance(existing, dict)
```
sfl-pe-sync, line 368: ERROR: Merging every changed record as `{**old_record, **new_record}` retains standard fields that the current exporter deliberately omitted. For example, when an OpenVEX assessment changes from `not_affected` to `affected`, a previously emitted `justification` can survive because the new statement does not overwrite it, producing stale or invalid review semantics. Use the generated record as the source of truth for all schema-defined fields and preserve only explicitly recognized extension fields from the old record.

frontend/src/pages/Review.tsx
```diff
@@ -484,8 +490,39 @@ function Review({ variantId, projectId, onAssessmentChanged }: Readonly<Props>)
...
+                reader.onerror = () => reject(new Error('Unable to read export file.'));
+                reader.readAsText(file);
+            });
+            const parsed = JSON.parse(text);
+            const detectedFormat = detectReviewExportFormat(parsed);
+            changeTransferFormat(detectedFormat);
+            setExistingExportFile(file);
+        } catch (error) {
+            setExistingExportError(error instanceof Error ? error.message : 'Unable to read export file.');
+        }
```
sfl-pe-sync, line 521: WARNING: Multiple file selections can resolve out of order: an earlier `FileReader` may finish after a later one and overwrite `transferFormat` and `existingExportFile`, causing Update export to submit the wrong file. Track a selection generation/current file (or abort the previous read) and ignore stale completion handlers.

sfl-pe-sync (changes requested):
The stable-update feature is not ready to merge because deleted data cannot be reconciled to an empty export and changed records can retain obsolete schema fields; the frontend also has an out-of-order file-read race.

src/helpers/assessment_io.py
```diff
@@ -164,6 +165,119 @@ def is_openvex_doc(doc: object) -> bool:
...
+        for key, value in current.items():
+            if key != "@id":
+                result[key] = value
+        result["statements"] = _reconcile_records(
+            existing["statements"],
+            current["statements"],
+            _openvex_statement_identity,
+        )
+        return result
+
```
sfl-pe-sync, line 381: ERROR: Reusing the existing OpenVEX `@id` while copying the freshly generated `version` (normally `1`) makes every changed update retain or reset its revision number. OpenVEX consumers use the `@id`/`version` pair to order revisions, so they can treat the downloaded update as the same or an older document and ignore its changed statements. Preserve the existing version for a byte-identical document; otherwise validate it as an integer and set the result to `existing["version"] + 1` (with a regression test covering an existing version greater than 1).

sfl-pe-sync (changes requested):
The OpenVEX update path does not advance the document version while preserving its stable ID, so updated exports can be ignored as stale revisions.

src/helpers/assessment_io.py
reh-sfl: The goal is to have an updated file appended to the original one, the versioning is handled by VCS.

ValentinBoudevin (dismissed):
1- Export method should be the last option below Variants selection

Image

2- A click on "Append/update existing file" remove the click option between VulnScout/OpenVex

Image

3- Exported filename doesn't need to be the same as the input one.

reh-sfl:
@ValentinBoudevin for 2, it's normal because we automatically detect the file format from what the user uploaded. This also makes it better to keep the order as I implemented it as it will be confusing to users to see dimmed checkbox.
I fixed 3.

src/helpers/assessment_io.py
```diff
@@ -164,6 +165,119 @@ def is_openvex_doc(doc: object) -> bool:
...
+)
+
+
+def detect_review_export_format(doc: object) -> str:
+    """Return the supported Review export format represented by *doc*.
+
+    Raises ``ValueError`` with a user-facing explanation for malformed or
+    unsupported input.
+    """
+    if is_openvex_doc(doc):
```
sfl-pe-sync: ERROR: This is used as validation for the uploaded document, but it only recognizes a context substring and a `statements` array. For example, `{"@context":"openvex","statements":[]}` is accepted; reconciliation then deliberately omits the generated `@id`, so the endpoint returns an OpenVEX document with no required identifier. Invalid nested identity fields can likewise escape this check and raise an uncaught `TypeError` during reconciliation. Validate the required top-level OpenVEX fields and record field types (and the custom record identities) before reconciling, returning 400 for malformed uploads; alternatively retain generated required fields whenever the existing value is absent or invalid.

src/helpers/assessment_io.py
```diff
@@ -164,6 +165,119 @@ def is_openvex_doc(doc: object) -> bool:
...
+    existing: list[Any],
+    current: list[Any],
+    identity: Any,
+) -> list[Any]:
+    """Retain existing order for matching records and append new records."""
+    current_indices: dict[tuple[Any, ...], deque[int]] = defaultdict(deque)
+    for index, record in enumerate(current):
+        if not isinstance(record, dict):
+            raise ValueError("Generated export contains an invalid record")
+        current_indices[identity(record)].append(index)
```
sfl-pe-sync: ERROR: The reconciliation key is not unique for the assessment history that this application deliberately preserves: multiple custom assessments (and OpenVEX statements) can share variant, vulnerability, and package/products while differing by timestamp. Adding or deleting one such assessment causes the FIFO match to pair every existing record with the wrong current record, moving unchanged entries and transferring preserved `x-` extensions to a different assessment. First consume exact matches within each identity group, then pair only the remaining unambiguous changed records (using timestamp/stable IDs where available), so additions/deletions cannot rotate metadata across historical assessments.

frontend/src/pages/Review.tsx
```diff
@@ -484,8 +490,43 @@ function Review({ variantId, projectId, onAssessmentChanged }: Readonly<Props>)
         setTransferVariantIds(defaultVariantId ? [defaultVariantId] : []);
     }, [transferMode, transferVariants, variantId]);
 
+    const changeExportMode = useCallback((mode: 'normal' | 'update') => {
+        setExportMode(mode);
+        setExistingExportFile(undefined);
```
sfl-pe-sync, line 495: WARNING: Clearing the selected file does not invalidate an in-flight `FileReader`. If the user chooses a file and then switches back to Normal export (or closes and reopens the modal) before the read completes, that stale callback can repopulate `existingExportFile` and change the format; switching to Update can then submit the old file without a new selection. Increment the selection generation (or abort the active reader) whenever the mode/modal is reset as well as when another file is chosen.

sfl-pe-sync (changes requested):
The update flow still needs revision: malformed OpenVEX uploads can produce invalid exports, non-unique reconciliation identities can rotate extension metadata across assessment history, and file reads can outlive modal/mode resets.

src/helpers/assessment_io.py
```diff
@@ -164,6 +165,119 @@ def is_openvex_doc(doc: object) -> bool:
...
+        package_key = tuple(sorted(str(package) for package in packages)) if isinstance(packages, list) else ()
+        return variant_id, vuln_id, package_key
+    return variant_id, vuln_id, None
+
+
+def _openvex_statement_identity(record: dict[str, Any]) -> tuple[Any, ...]:
+    vulnerability = record.get("vulnerability", {})
+    products = record.get("products", [])
+    vuln_name = vulnerability.get("name") if isinstance(vulnerability, dict) else None
+    product_ids = tuple(
```
sfl-pe-sync, line 241: ERROR: OpenVEX statement identity treats `products` as an ordered tuple even though product-array order is not part of the statement's semantic identity. If an existing exporter writes the same products in a different order than `build_openvex_doc`, this record is treated as removed and re-added, changing statement order and discarding its preserved `x-` extension fields. Canonicalize product IDs (for example, sort them as `_record_identity` does for packages) before matching, and add a regression test with reordered products and an `x-` field.

sfl-pe-sync (changes requested):
The update flow can lose existing OpenVEX extension data when semantically identical product lists differ only in order, so the stable reconciliation behavior is not yet safe to merge.

src/routes/assessments.py
```diff
@@ -807,6 +809,75 @@ def export_review_custom_data() -> ResponseReturnValue:
...
+            variant_id, err = parse_uuid_or_400(raw_variant_id, "variant_id")
+            if err:
+                return err
+            if variant_id is None:
+                return {"error": "Internal error"}, 500
+            variant = DBVariant.get_by_id(variant_id)
+            if variant is None:
+                return {"error": f"Variant not found: {raw_variant_id}"}, 404
+            variant_ids.append(variant_id)
+
```
sfl-pe-sync, line 858: ERROR: The new update endpoint resolves each submitted variant UUID globally and never constrains the selection to the Review's project (or even requires all variants to belong to one project). A crafted multipart request can therefore combine or retrieve review data from variants outside the current project, bypassing the frontend's project-filtered selector. Include the project identity in the request and verify every variant belongs to it before querying assessments/building the export, using the repository's project-scoped lookup/authorization pattern.

sfl-pe-sync (changes requested):
The update flow introduces a blocking project-scoping gap: submitted variant IDs are trusted and resolved globally, so the backend can export review data outside the current project.

frontend/src/helpers/exportJson.ts
```diff
@@ -19,6 +19,28 @@ export function formatTimestampForFilename(date?: Date | string): string {
...
 }
 
+export type ReviewExportFormat = 'custom' | 'openvex';
+
+/** Detect a supported Review export format from parsed JSON. */
+export function detectReviewExportFormat(data: unknown): ReviewExportFormat {
+    if (!data || typeof data !== 'object' || Array.isArray(data)) {
+        throw new Error('Export file must contain a JSON object.');
+    }
+    const document = data as Record<string, unknown>;
```
sfl-pe-sync, line 29: WARNING: The browser detector accepts files that the new backend detector rejects, so malformed exports such as `{ "version": 1, "assessments": [{}] }` are shown as successfully detected and enable Update export, only to fail after submission. Mirror the backend's required top-level and record-identity validation here (including OpenVEX `@id`/author/timestamp/version and nested statement identities), or avoid presenting the file as valid until server validation succeeds.

sfl-pe-sync (changes requested):
The update flow has a frontend/backend validation mismatch that presents unsupported files as valid before submission.

ValentinBoudevinSFL (changes requested):
1- Image

Based on my previous feedback and your comment, I see what we should do:

Image

Right now a click on "Append/update existing file", even without a selected file, fix the "Format" value. 

Maybe "Format" section should only be visible in "Normal Export" mode. 

In addition adding a file to "Existing export file" the format name is already written down

Image

Note: I'm open to any other solution but I don't like to fix "Format" checkbox when no input file is provided

ValentinBoudevinSFL (changes requested):
Git description is still empty

ValentinBoudevinSFL (approved):

frontend/src/pages/Review.tsx
```diff
@@ -494,26 +535,38 @@ function Review({ variantId, projectId, onAssessmentChanged }: Readonly<Props>)
...
                 const err = await res.json().catch(() => ({}));
                 showMessage(err.error || 'Failed to export review data.', 'error');
                 return;
             }
+            const exported = await res.json();
             const ts = formatTimestampForFilename();
             if (transferFormat === 'openvex') {
                 const label = variantNames[transferVariantIds[0]] ?? 'variant';
-                downloadJson(await res.json(), `review_openvex_${sanitizeFilename(label)}_${ts}.json`);
+                downloadJson(exported, `review_openvex_${sanitizeFilename(label)}_${ts}.json`);
```
sfl-pe-sync, line 559: WARNING: The update response's `Content-Disposition` and the uploaded file's name are ignored, so an update of `tracked-review.json` is downloaded under a new timestamped `review_openvex_*`/`custom_data_*` name. That defeats the expected replace-in-place workflow for a Git-tracked stable export and makes the backend's preserved filename ineffective. In update mode, pass `existingExportFile.name` (sanitized) to `downloadJson`, while retaining generated names for normal exports.

src/helpers/assessment_io.py
```diff
@@ -160,8 +161,234 @@ def is_openvex_doc(doc: object) -> bool:
...
+            continue
+        consumed.add(current_index)
+        new_record = current_records[current_index]
+        extensions = {
+            key: value for key, value in old_record.items()
+            if isinstance(key, str) and key.startswith("x-")
+        }
+        reconciled.append(
+            old_record if old_record == new_record
+            else {**_record_without_extensions(new_record), **extensions}
```
sfl-pe-sync, line 356: WARNING: An otherwise unchanged record containing an `x-` extension can never take the `old_record == new_record` branch because generated records omit that extension. Rebuilding it as generated fields followed by extensions moves any extension keys to the end of the object, creating avoidable Git diff churn in the feature intended to preserve unchanged content. Track exact-content matches ignoring extensions and retain the original record verbatim for those matches; only rebuild records matched as changed.

sfl-pe-sync (dismissed):
The export update behavior has two non-blocking regressions in its stable-file workflow: downloaded updates discard the original filename, and unchanged records with extensions can be reordered.

src/routes/assessments.py
```diff
@@ -807,6 +809,75 @@ def export_review_custom_data() -> ResponseReturnValue:
...
+            )
+        else:
+            current = build_custom_data_export(variant_ids)
+
+        try:
+            updated = reconcile_review_export(existing, current)
+        except ValueError as error:
+            return {"error": str(error)}, 400
+
+        filename = re.sub(r'[^\w\-.]', '_', uploaded.filename) or 'review_export.json'
```
sfl-pe-sync, line 875: WARNING: Python's `\w` is Unicode-aware, so this sanitizer retains characters such as CJK letters even though WSGI response headers must be Latin-1 encodable. Updating an upload named, for example, `レビュー.json` can therefore fail while emitting `Content-Disposition` instead of returning the export. Use `werkzeug.utils.secure_filename` or an explicitly ASCII-only sanitizer, with a safe fallback filename.

sfl-pe-sync (approved):
The update flow has a filename compatibility defect: Unicode word characters can survive sanitization and make the download response header invalid under WSGI.

PR #522, opened 2026-08-12
Add scan history import support

reh-sfl:

* Add an Import control to the Scan history toolbar using the established transfer interaction.
* Accept exported diffs and full scans, deriving the relevant diff before persistence.
* Persist imported scan context atomically with destination-aware derivation, progress states, errors, tests, and documentation.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

* Run `python -m pytest tests/webapp_tests/test_scan_export.py -q`.
* Run `npm --prefix frontend test -- --runInBand tests/handlers/scans.test.ts`.
* Run `npm --prefix frontend run build`.
* In Scan history, select Import and upload first an exported diff and then a full scan; confirm each adds the derived context to the selected destination and reports invalid input without partial persistence.

### Additional notes

Full-scan imports calculate the diff against the destination context, and persistence rolls back atomically on failure.

### Related Issue

No linked issue.

## Pull Request Checklist

### DIFF
--- a/frontend/src/pages/ScanHistory.tsx
+++ b/frontend/src/pages/ScanHistory.tsx
@@ -35,9 +35,10 @@ import { extractSupplierName } from "../helpers/pkgId";
 import { formatSourceName } from "../helpers/sourceNames";
 import { downloadJson } from "../helpers/exportJson";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
-import { faPencil, faCheck, faXmark, faBug, faFilter, faShieldHalved, faLeaf, faFile, faCrosshairs, faTrash, faPlay, faBook, faDownload, faMagnifyingGlass, faBox, faClipboardCheck } from "@fortawesome/free-solid-svg-icons";
+import { faPencil, faCheck, faXmark, faBug, faFilter, faShieldHalved, faLeaf, faFile, faFileImport, faCrosshairs, faTrash, faPlay, faBook, faDownload, faMagnifyingGlass, faBox, faClipboardCheck } from "@fortawesome/free-solid-svg-icons";
 import type { IconDefinition } from "@fortawesome/free-solid-svg-icons";
 import ConfirmationModal from "../components/ConfirmationModal";
+import MessageBanner from "../components/MessageBanner";
 import RunScansWizard from "../components/RunScansWizard";
 import { refreshSourcesForScans } from "../helpers/refreshSources";
 import Variants from "../handlers/variant";
@@ -1112,13 +1113,16 @@ function ScanHistory({ variantId, projectId, onScanComplete }: Readonly<Props>) 
     const [showNvd, setShowNvd] = useState(true);
     const [showScc, setShowScc] = useState(true);
 
+    // Import state
+    const [importing, setImporting] = useState(false);
+    const [importNotice, setImportNotice] = useState<{ tone: 'success' | 'error'; text: string } | null>(null);
+    const importInputRef = useRef<HTMLInputElement>(null);
+
     // Export state
     const [exportMenuScanId, setExportMenuScanId] = useState<string | null>(null);
     const [exportingScanId, setExportingScanId] = useState<string | null>(null);
-    const [exportAllMenuOpen, setExportAllMenuOpen] = useState(false);
     const [exportingAll, setExportingAll] = useState(false);
     const exportMenuRef = useRef<HTMLDivElement>(null);
-    const exportAllMenuRef = useRef<HTMLDivElement>(null);
 
     // Scan wizard state
     const [scanWizardOpen, setScanWizardOpen] = useState(false);
@@ -1195,6 +1199,82 @@ function ScanHistory({ variantId, projectId, onScanComplete }: Readonly<Props>) 
         }
     }
 
+    /**
+     * Read the selected export files and import them in a single request.
+     *
+     * Each file holds either one export object or an array of them (as the
+     * Export All download does). They are flattened into one payload so the
+     * server imports the whole selection atomically: either every scan lands
+     * or none does, which is what makes a failed import safe to retry.
+     */
+    async function handleImportFile(event: React.ChangeEvent<HTMLInputElement>) {
+        const files = Array.from(event.target.files ?? []);
+        if (files.length === 0) return;
+        setImporting(true);
+        setImportNotice(null);
+        try {
+            const exports: unknown[] = [];
+            for (const file of files) {
+                let parsed: unknown;
+                try {
+                    parsed = JSON.parse(await file.text());
+                } catch (err) {
+                    const reason = err instanceof SyntaxError
+                        ? 'it is not valid JSON'
+                        : 'it could not be read';
+                    setImportNotice({
+                        tone: 'error',
+                        text: `Import failed: "${file.name}" was skipped because ${reason}.`,
+                    });
+                    return;
+                }
+                if (Array.isArray(parsed)) exports.push(...parsed);
+                else exports.push(parsed);
+            }
+            if (exports.length === 0) {
+                setImportNotice({
+                    tone: 'error',
+                    text: 'Import failed: the selected file contains no scan exports.',
+                });
+                return;
+            }
+
+            const response = await ScansHandler.importExport(exports);
+            if (!response.ok) {
+                setImportNotice({ tone: 'error', text: response.error });
+                return;
+            }
+            const { result } = response;
+            const scanLabel = result.imported_count === 1
+                ? (result.format === 'diff' ? '1 scan diff' : '1 full scan result')
+                : `${result.imported_count.toLocaleString()} scans`;
+            const parts = [
+                `${result.package_count.toLocaleString()} packages`,
+                `${result.finding_count.toLocaleString()} findings`,
+            ];
+            if (result.assessment_count > 0) {
+                parts.push(`${result.assessment_count.toLocaleString()} assessments`);
+            }
+            const skipped = result.skipped_count > 0
+                ? ` Skipped ${result.skipped_count.toLocaleString()} already imported scan${result.skipped_count === 1 ? '' : 's'}.`
+                : '';
+            setImportNotice({
+                tone: 'success',
+                text: `Imported ${scanLabel} with ${parts.join(', ')}.${skipped}`,
+            });
+            refreshScans();
+            onScanComplete?.();
+        } catch {
+            setImportNotice({
+                tone: 'error',
+                text: 'Import failed while reading the selected files.',
+            });
+        } finally {
+            event.target.value = '';
+            setImporting(false);
+        }
+    }
+
     // Derive the effective variant IDs to scan: explicit prop or unique IDs from loaded scans
     const effectiveVariantIds: string[] = variantId
         ? [variantId]
@@ -1262,18 +1342,6 @@ function ScanHistory({ variantId, projectId, onScanComplete }: Readonly<Props>) 
             return () => document.removeEventListener('mousedown', handleClickOutside);
         }
     }, [exportMenuScanId]);
-
-    useEffect(() => {
-        function handleClickOutside(e: MouseEvent) {
-            if (exportAllMenuRef.current && !exportAllMenuRef.current.contains(e.target as Node)) {
-                setExportAllMenuOpen(false);
-            }
-        }
-        if (exportAllMenuOpen) {
-            document.addEventListener('mousedown', handleClickOutside);
-            return () => document.removeEventListener('mousedown', handleClickOutside);
-        }
-    }, [exportAllMenuOpen]);
 
     function toggleVariant(vid: string) {
         setSelectedVariantIds(prev => {
@@ -1412,14 +1480,13 @@ function ScanHistory({ variantId, projectId, onScanComplete }: Readonly<Props>) 
     });
 
     // -- Export All (grouped by project/variant) --
-    async function handleExportAll(type: 'diff' | 'total') {
-        setExportAllMenuOpen(false);
+    async function handleExportAll() {
         setExportingAll(true);
         try {
-            const params = new URLSearchParams({ type });
+            const params = new URLSearchParams({ type: 'diff' });
             if (variantId) params.set('variant_id', variantId);
             else if (projectId) params.set('project_id', projectId);
-            await downloadFromEndpoint(`/api/scans/export?${params}`, `scans_${type}.json`);
+            await downloadFromEndpoint(`/api/scans/export?${params}`, 'scans_diff.json');
         } catch (err) {
             setError(err instanceof Error ? err.message : 'Failed to export scans.');
         } finally {
@@ -1440,6 +1507,17 @@ function ScanHistory({ variantId, projectId, onScanComplete }: Readonly<Props>) 
     // Column sizing — single lane
     const LANE_W = 36;            // px – timeline column
     const mainCX = LANE_W / 2;    // center-x of the lane
+
+    const importNoticeBanner = importNotice && (
+        <div className="sticky top-0 z-40">
+            <MessageBanner
+                type={importNotice.tone}
+                message={importNotice.text}
+                isVisible={true}
+                onClose={() => setImportNotice(null)}
+            />
+        </div>
+    );
 
     const menuBar = (
         <div className="rounded-md mb-4 p-2 bg-sky-800 text-white w-full flex flex-row items-center gap-2 flex-wrap">
@@ -1508,7 +1586,7 @@ function ScanHistory({ variantId, projectId, onScanComplete }: Readonly<Props>) 
                 sbom-cve-check
             </button>
 
-            {/* Right side: doc link + export + scan menu */}
+            {/* Right side: doc link + import/export + scan menu */}
             <div className="ml-auto flex items-center gap-3">
                 <a
                     href={docUrl}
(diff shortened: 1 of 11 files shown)


src/routes/scans.py
```diff
@@ -313,8 +319,168 @@ def _format_timestamp_for_filename(dt: datetime | str | None = None) -> str:
...
+        id=scan_id,
+        description=f"Imported scan from {payload['project_name']} / {payload['variant_name']}",
+        scan_type=scan_type,
+        scan_source=scan_source,
+        timestamp=timestamp,
+        variant_id=variant.id,
+    )
+    db.session.add(scan)
+    db.session.flush()
+    _persist_imported_scan_state(scan, packages, findings)
```
sfl-pe-sync: ERROR: `_persist_imported_scan_state` treats the top-level arrays as scan-local state for both formats, but `export-result` contains the global SBOM-plus-tool union and a non-first diff export is not a complete scan snapshot (for example, upgraded packages/findings are represented under `diff.*_upgraded` and omitted from the top-level arrays). Importing a tool full-result therefore assigns SBOM/other-tool findings to this tool source, while importing an upgrade diff loses the upgraded package/finding, corrupting later source-aware results and diffs. Reconstruct the destination scan state with format-specific logic from `diff`/`sources`, or extend the export contract with an explicit complete scan-local snapshot and import that instead.

src/routes/scans.py
```diff
@@ -313,8 +319,168 @@ def _format_timestamp_for_filename(dt: datetime | str | None = None) -> str:
...
+
+def _import_scan_export(payload: object) -> tuple[Scan, str]:
+    """Persist one scan diff or full-result export and return its format."""
+    if not isinstance(payload, dict):
+        raise ValueError("Scan import must be a JSON object")
+
+    variant = _import_destination(payload)
+    scan_id, scan_type, scan_source, timestamp = _import_scan_metadata(payload)
+
+    packages = payload.get("packages", [])
```
sfl-pe-sync: ERROR: Missing `packages` and `findings` are defaulted to empty arrays, so any JSON object with plausible metadata is accepted as an exported scan. A malformed payload with a recent `import_sbom` timestamp is persisted as an empty active SBOM and makes the variant's current packages/findings disappear. Require the fields appropriate to the declared scan type/format plus an export schema/version discriminator, and fully validate the payload before adding the `Scan`.

src/routes/scans.py
sfl-pe-sync: ERROR: Full-result exports include `assessments`, but the importer never reads or persists them, so restoring an exported scan silently loses all triage state. Import these records as variant-scoped assessments linked to the reconstructed findings while preserving an imported origin, or explicitly reject assessment-bearing payloads until that context is supported rather than reporting a successful full import.

sfl-pe-sync (changes requested):
The new endpoint is not safe to merge because it does not reconstruct the semantics of the two export formats, permits malformed empty scans to become active history, and reports success while dropping exported assessments.

src/routes/scans.py
```diff
@@ -313,8 +322,184 @@ def _format_timestamp_for_filename(dt: datetime | str | None = None) -> str:
...
+
+def _import_scan_export(payload: object) -> tuple[Scan, str]:
+    """Persist one complete scan-local diff export and return its format."""
+    if not isinstance(payload, dict):
+        raise ValueError("Scan import must be a JSON object")
+
+    if payload.get("export_version") != 1:
+        raise ValueError("'export_version' must be 1")
+    export_format = payload.get("export_format")
+    if export_format != "scan-diff":
```
sfl-pe-sync: ERROR: The import endpoint rejects every `full-result` export here and rejects every non-first `scan-diff` at line 435, so the documented and advertised workflow fails for both a full scan and almost every scan-history diff with HTTP 400. Implement reconstruction for both exported formats as the PR contract requires (including assessment handling), or consistently narrow the API, UI, documentation, and PR scope rather than presenting those files as importable.

sfl-pe-sync (changes requested):
The import path is not ready because it only accepts first-scan diff exports while the feature contract promises full-result and general diff imports.

ValentinBoudevin (changes requested):
Export requires a rework too:

Image

1- Frontend for the export should be coherent

2- Remove "All Scans Results" option


----

Import fails on my side:

```
Traceback (most recent call last):
  File "/scan/src/routes/scans.py", line 477, in import_scan
    scan, import_format = _import_scan_export(flask_request.get_json(silent=True))
                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/werkzeug/wrappers/request.py", line 608, in get_json
    data = self.get_data(cache=cache)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/werkzeug/wrappers/request.py", line 422, in get_data
    rv = self.stream.read()
         ^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/werkzeug/utils.py", line 100, in __get__
    value = self.fget(obj)  # type: ignore
            ^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/werkzeug/wrappers/request.py", line 351, in stream
    return get_input_stream(
           ^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/werkzeug/wsgi.py", line 187, in get_input_stream
    raise RequestEntityTooLarge()
werkzeug.exceptions.RequestEntityTooLarge: 413 Request Entity Too Large: The data value transmitted exceeds the capacity limit.
172.17.0.1 - - [13/Aug/2026 14:59:47] "POST /api/scans/import HTTP/1.1" 500 -
```

frontend/src/pages/ScanHistory.tsx
```diff
@@ -1521,42 +1525,29 @@ function ScanHistory({ variantId, projectId, onScanComplete }: Readonly<Props>)
...
-                                    onClick={() => handleExportAll('total')}
-                                    className="w-full text-left px-3 py-2 text-sm text-neutral-200 hover:bg-sky-900/40 rounded transition-colors"
-                                >
-                                    Export All Scan Results
-                                </button>
-                            </div>
-                        )}
-                    </div>
-                )}
+                <button
```
sfl-pe-sync, line 1602: ERROR: This replacement removes the existing toolbar Export dropdown and its `handleExportAll` path, so users can no longer export all scans in the current project/variant scope even though the backend endpoint remains and the updated documentation still advertises that workflow. Keep the batch Export control alongside Import (and retain its loading/error handling) rather than deleting it.

sfl-pe-sync (changes requested):
The import work introduces a blocking frontend regression by removing the documented scan-history batch export control.

frontend/src/pages/ScanHistory.tsx
reh-sfl: This is intentional

src/routes/scans.py
```diff
@@ -313,8 +324,266 @@ def _format_timestamp_for_filename(dt: datetime | str | None = None) -> str:
...
+        for finding in imported_findings:
+            findings_by_vulnerability.setdefault(finding.vulnerability_id, []).append(finding)
+
+    for assessment_entry in assessments:
+        if not isinstance(assessment_entry, dict):
+            raise ValueError("Assessment entries must be JSON objects")
+        vulnerability_id = _required_import_string(assessment_entry, "vulnerability_id").upper()
+        status = _required_import_string(assessment_entry, "status")
+        for finding in findings_by_vulnerability.get(vulnerability_id, []):
+            Assessment.create(
```
sfl-pe-sync: ERROR: Assessment linkage is reconstructed by vulnerability ID alone, but assessments are scoped to a specific finding/package (the latest-assessment query partitions by `Finding.package_id`). If the same CVE affects two imported packages with different triage states, each exported assessment is copied to every matching finding, producing duplicated and incorrect statuses. Include package/finding identity in the export schema and resolve each assessment only to that exact imported finding.

src/routes/scans.py
```diff
@@ -313,8 +324,266 @@ def _format_timestamp_for_filename(dt: datetime | str | None = None) -> str:
...
+                status=status,
```
sfl-pe-sync: ERROR: Imported assessments are created without their exported timestamp, so `Assessment.create` assigns the import time. Importing historical context into a variant with newer human triage can therefore make the old imported verdict the latest current status; it also places the assessment outside the imported scan's historical cutoff when a later scan already exists. Export and validate each assessment timestamp and pass it to `Assessment.create` (with an explicit conflict policy for destination assessments) so chronological ordering is preserved.

sfl-pe-sync (changes requested):
The scan-state reconstruction is substantially expanded, but assessment imports still corrupt package-specific linkage and timestamp-based latest-state semantics, so the change is not safe to merge.

src/routes/scans.py
```diff
@@ -313,8 +322,184 @@ def _format_timestamp_for_filename(dt: datetime | str | None = None) -> str:
...
+    if scan.scan_type == "sbom":
+        document = SBOMDocument(
+            path=f"imported-scan://{scan.id}",
+            source_name="Imported VulnScout scan",
+            format="vulnscout_json",
+            scan_id=scan.id,
+        )
+        db.session.add(document)
+        db.session.flush()
+
```
sfl-pe-sync: ERROR: Offset-aware timestamps are stored without normalization. SQLite drops timezone information, and the rest of the application reattaches UTC when reading timestamps, so importing `2026-01-01T00:00:00+14:00` is later interpreted as `2026-01-01T00:00:00Z` instead of the equivalent `2025-12-31T10:00:00Z`. This changes scan ordering, active-scan selection, and the derived diff. Convert every aware timestamp with `timestamp.astimezone(timezone.utc)` before assigning it to `Scan`; only attach UTC directly for naive timestamps, and add an offset regression test.

src/routes/scans.py
```diff
@@ -313,8 +322,184 @@ def _format_timestamp_for_filename(dt: datetime | str | None = None) -> str:
...
+    if "diff" in payload:
+        raise ValueError("Only first-scan diff exports contain a complete scan-local state")
+    assessment_fields = ("assessments", "newly_detected_assessments")
+    if any(payload.get(field) for field in assessment_fields):
+        raise ValueError("Assessment-bearing scan exports cannot be imported")
+
+    variant = _import_destination(payload)
+    scan_id, scan_type, scan_source, timestamp = _import_scan_metadata(payload)
+
+    if "findings" not in payload or not isinstance(payload["findings"], list):
```
sfl-pe-sync: ERROR: The accepted first-scan export includes a `vulnerabilities` collection, but this path ignores it and creates missing vulnerabilities with only their IDs. Importing into a fresh destination therefore drops exported descriptions, aliases, severity/CVSS, dates, and other vulnerability context while still reporting success. Validate and persist/merge each exported vulnerability record before creating observations, and test restoration into a database where those vulnerabilities do not already exist.

src/routes/scans.py
```diff
@@ -313,8 +322,184 @@ def _format_timestamp_for_filename(dt: datetime | str | None = None) -> str:
...
+        raise ValueError("Only 'scan-diff' exports can be imported")
+    if "diff" in payload:
+        raise ValueError("Only first-scan diff exports contain a complete scan-local state")
+    assessment_fields = ("assessments", "newly_detected_assessments")
+    if any(payload.get(field) for field in assessment_fields):
+        raise ValueError("Assessment-bearing scan exports cannot be imported")
+
+    variant = _import_destination(payload)
+    scan_id, scan_type, scan_source, timestamp = _import_scan_metadata(payload)
+
```
sfl-pe-sync: WARNING: Import performs database lookups per entry: `Package.find_or_create` once per package, `db.session.get(Vulnerability, ...)` once per finding, and `Finding.get_or_create` once per finding, with additional flushes for new vulnerabilities. A production-sized export therefore grows to several queries per finding while holding one write transaction and can make the import endpoint time out or contend on SQLite. Batch-resolve packages, vulnerability IDs, and findings up front (reusing `Package.bulk_find_or_create`) and bulk-add only missing rows.

sfl-pe-sync (changes requested):
The import path still corrupts timestamp ordering for non-UTC exports and silently discards exported vulnerability metadata; its per-entry query pattern also makes large imports unnecessarily expensive.

ValentinBoudevin (dismissed):
My comments are still valid, nothing changed

src/routes/scans.py
```diff
@@ -313,8 +336,744 @@ def _format_timestamp_for_filename(dt: datetime | str | None = None) -> str:
...
+def _import_assessment_entry(entry: object) -> _ImportAssessment:
+    if not isinstance(entry, dict):
+        raise ValueError("Assessment entries must be JSON objects")
+    package_key = _import_package_key(entry) if entry.get("package_name") else None
+    return _ImportAssessment(
+        vulnerability_id=_required_import_string(entry, "vulnerability_id").upper(),
+        package_key=package_key,
+        values={
+            "status": _required_import_string(entry, "status"),
+            "simplified_status": _optional_import_string(entry, "simplified_status"),
```
sfl-pe-sync, line 518: ERROR: Imported assessment statuses and `simplified_status` values are accepted verbatim, bypassing the normalized-status validation used by the assessment API. A modified export can therefore persist an arbitrary status or a contradictory display status, which then pollutes current-status summaries, filtering, metrics, and subsequent exports. Validate `status` against the canonical status mapping and derive (or strictly verify) `simplified_status` from it before creating the assessment.

src/routes/scans.py
```diff
@@ -313,8 +336,744 @@ def _format_timestamp_for_filename(dt: datetime | str | None = None) -> str:
...
+    # exports list them under ``newly_detected_assessments``.
+    assessments = [
+        _import_assessment_entry(entry)
+        for key in ("assessments", "newly_detected_assessments")
+        for entry in _import_array(payload, key, required=False)
+    ]
+    known_vulns = {vulnerability_id for _key, vulnerability_id in findings}
+    unmatched = sorted({a.vulnerability_id for a in assessments} - known_vulns)
+    if unmatched:
+        raise ValueError(
```
sfl-pe-sync, line 604: ERROR: This validation checks only that an assessment's vulnerability exists somewhere in the export, not that its declared `(package, vulnerability)` finding exists. `_persist_import_assessments` later resolves that pair with `.get(...)` and silently uses an empty target list on a mismatch, so the endpoint returns 201 while dropping the assessment. Validate every package-qualified assessment against the exact `findings` key set and reject unmatched entries before persistence.

sfl-pe-sync (changes requested):
The importer still permits invalid assessment state and can silently lose package-qualified assessments while reporting success, so the scan-history restoration contract is not yet safe to merge.

src/routes/scans.py
```diff
@@ -313,8 +336,744 @@ def _format_timestamp_for_filename(dt: datetime | str | None = None) -> str:
...
+            package_keys.append(key)
+
+    findings: list[FindingKey] = []
+    seen_findings: set[FindingKey] = set()
+    for entry in findings_raw:
+        key = _import_package_key(entry)
+        if key not in seen_packages:
+            seen_packages.add(key)
+            package_keys.append(key)
+        pair = (key, _required_import_string(entry, "vulnerability_id").upper())
```
sfl-pe-sync, line 589: ERROR: For an SBOM import, a finding whose package is absent from the declared `packages` array is silently appended to `package_keys`; `_persist_scan_import` then links every such key into the new `SBOMDocument`. A malformed export can therefore claim `packages: []` (or omit a specific package) yet make a package supplied only by `findings` part of the destination's active SBOM, so the persisted package set no longer matches the exported scan. Validate that every SBOM finding references a package already declared in `packages` and reject mismatches; keep any finding-only package resolution needed for tool scans separate from the SBOM membership list.

sfl-pe-sync (changes requested):
The import path still accepts contradictory SBOM package/finding state and persists undeclared packages into the active SBOM, so the change is not safe to merge.

ValentinBoudevinSFL (changes requested):
My feedbacks 1 and 2 are still valids

src/routes/scans.py
```diff
@@ -313,8 +336,744 @@ def _format_timestamp_for_filename(dt: datetime | str | None = None) -> str:
...
+            if as_exported != stored:
+                known.setdefault(as_exported, package)
+
+    created = False
+    for key in keys:
+        if key in known:
+            continue
+        name, version, supplier = key
+        package = Package(
+            name=name, version=version, cpe=[], purl=[],
```
sfl-pe-sync, line 663: ERROR: New packages are recreated with empty CPE, PURL, and licence fields, and neither export format carries those fields. On a real migration to a clean database (unlike the tests, which create the destination in the source database and therefore reuse its global Package rows), imported packages lose the identifiers required by the NVD/OSV scanners and subsequent SBOM exports. Extend the versioned export schema with the package metadata and persist or merge it during import; add a round-trip test using a separate clean database.

src/routes/scans.py
```diff
@@ -313,8 +336,744 @@ def _format_timestamp_for_filename(dt: datetime | str | None = None) -> str:
...
+    assessments already on the destination variant are left alone.
+    """
+    if not item.assessments:
+        return 0
+
+    # Assessments whose origin is unset are filtered out of every export and
+    # scan diff (SQL ``NOT IN`` drops NULLs), so an imported assessment must
+    # carry the origin a natively-produced one of the same kind would have —
+    # otherwise it silently disappears the next time the data is exported.
+    origin = item.scan_source if item.scan_type == "tool" else "sbom"
```
sfl-pe-sync, line 885: ERROR: Assessment provenance is derived from the enclosing scan instead of the exported assessment. Diff and full-result exports can contain assessments contributed by multiple active sources, so importing an SBOM export relabels tool assessments as `sbom`, while importing a tool full result relabels SBOM and other-tool assessments as that tool. This changes source-aware history and later exports. Include each assessment's validated `origin` in the export and restore it per assessment rather than assigning one origin to the whole payload.

sfl-pe-sync (changes requested):
The import path still loses package scanner metadata in clean migrations and rewrites assessment provenance, so exported scan context does not round-trip safely.

ValentinBoudevinSFL (changes requested):
Frontend LGTM

Image

I face this issue. I have an history with A,B,C and a scan to import with A,B,C,D.

I should just skip A,B,C which are already present and see the new D scan. 
But for the moment it just raise an error as A already exist

ValentinBoudevinSFL (approved):
LGTM

src/routes/scans.py
```diff
@@ -313,8 +336,744 @@ def _format_timestamp_for_filename(dt: datetime | str | None = None) -> str:
...
+    db.session.flush()
+
+    packages = _resolve_import_packages(item.package_keys)
+
+    if item.scan_type == "sbom":
+        # SBOM scans expose their package set through an SBOM document; tool
+        # scans reach theirs through their findings and need none.
+        document = SBOMDocument(
+            path=f"imported-scan://{scan.id}",
+            source_name=_IMPORT_SOURCE_LABEL,
```
sfl-pe-sync, line 810: ERROR: Every imported SBOM is rewritten as one synthetic `vulnscout_json` document, even though full exports carry per-package/finding `sources`. Consequently an imported SPDX/CycloneDX/OpenVEX finding is later reported by `/api/vulnerabilities` as `found_by: vulnscout_json`, multiple source documents are merged, and re-export no longer matches the source provenance. Extend the export schema with document identity/format and finding-to-document relationships, then recreate the corresponding `SBOMDocument`/`SBOMObservation` links instead of discarding them.

src/routes/scans.py
```diff
@@ -313,8 +336,744 @@ def _format_timestamp_for_filename(dt: datetime | str | None = None) -> str:
...
+        )
+        if batch_key in batch_keys:
+            raise FileExistsError(_import_entry_error(
+                index, len(exports),
+                f"{described} appears more than once for {location} "
+                f"in this import",
+            ))
+        batch_keys.add(batch_key)
+
+        if _find_duplicate_scan(item) is not None:
```
sfl-pe-sync, line 950: ERROR: Idempotency is based only on `(variant, type, source, timestamp)` and has no database constraint. Distinct scans are allowed to share a timestamp and are deterministically ordered by ID elsewhere, so one is silently skipped during batch migration; two concurrent imports can also both pass this check and commit duplicates. Persist the source scan UUID (with a per-destination uniqueness constraint/migration, while recognizing a native row whose ID matches it) and use that durable identity for duplicate handling.

frontend/src/pages/ScanHistory.tsx
```diff
@@ -1195,6 +1199,82 @@ function ScanHistory({ variantId, projectId, onScanComplete }: Readonly<Props>)
...
+        const files = Array.from(event.target.files ?? []);
+        if (files.length === 0) return;
+        setImporting(true);
+        setImportNotice(null);
+        try {
+            const exports: unknown[] = [];
+            for (const file of files) {
+                let parsed: unknown;
+                try {
+                    parsed = JSON.parse(await file.text());
```
sfl-pe-sync, line 1220: WARNING: The browser reads and parses every selected file before enforcing any size or export-count bound. A user can select files far beyond the server's 100 MiB/200-export limits, causing multiple full-size copies (`file.text`, parsed objects, and `JSON.stringify`) and freezing or exhausting the tab before the server can return 413/400. Check cumulative `File.size` and flattened entry count before continuing, using limits aligned with the API.

sfl-pe-sync (dismissed):
The import path still has blocking identity and fidelity defects: legitimate scans and package identities can be collapsed, concurrent imports are not idempotent, and SBOM provenance is rewritten into an incorrect source classification.

---
- `src/routes/scans.py:135` [error]: ERROR: Exporting the display form of `package_supplier` as the package identity collapses distinct persisted packages. The schema permits both `Organization: Acme (a@example.com)` and `Acme` for the same name/version, but both become `Acme`; import then deduplicates their package/finding keys and can attach findings or assessments to the wrong row (or drop one entirely). Preserve the exact supplier identity in the versioned export, optionally carrying a separate display name, and add a collision round-trip test.

sfl-pe-sync (approved):
No actionable defects found.

PR #526, opened 2026-08-18
fix(podman): use IPv4 for scan readiness checks

reh-sfl:

Probe the published Podman port through 127.0.0.1 rather than localhost. This avoids CI failures when localhost resolves to an IPv6 loopback address that the Podman port forward does not expose.

Also make the shared scan waiter return immediately after reporting a timeout. Previously it emitted both 'Timeout reached' and 'OK', producing an ambiguous command-substitution result and obscuring the actual readiness failure.

### Status

- [ ] READY
- [ ] HOLD
- [X] WIP (Work-In-Progress)


sfl-pe-sync (approved):
No actionable defects found.

ValentinBoudevin (approved):

PR #527, opened 2026-08-19
Add amend option for review exports

reh-sfl:
## Description

Add `--amend` to custom VulnScout JSON and OpenVEX review exports. It reconciles the generated export with the existing file in `.vulnscout/outputs` to preserve stable ordering and reduce Git diff churn.

## Checklist

- [x] READY
- [x] Tests added or updated
- [ ] Documentation updated

## Related issue

No linked issue.

### DIFF
--- a/frontend/src/pages/Review.tsx
+++ b/frontend/src/pages/Review.tsx
@@ -459,10 +459,10 @@ function Review({ variantId, projectId, onAssessmentChanged }: Readonly<Props>) 
         setShowOnlyOutdated(false);
     };
 
-    const transferVariants = useMemo(() => {
-        const scopedProjectId = projectId ?? allVariants.find(v => v.id === variantId)?.project_id;
-        return scopedProjectId ? allVariants.filter(v => v.project_id === scopedProjectId) : allVariants;
-    }, [allVariants, projectId, variantId]);
+    const transferProjectId = projectId ?? allVariants.find(v => v.id === variantId)?.project_id;
+    const transferVariants = useMemo(() => (
+        transferProjectId ? allVariants.filter(v => v.project_id === transferProjectId) : allVariants
+    ), [allVariants, transferProjectId]);
 
     const openTransfer = useCallback((mode: 'import' | 'export') => {
         setTransferFormat('custom');
@@ -539,6 +539,7 @@ function Review({ variantId, projectId, onAssessmentChanged }: Readonly<Props>) 
             if (exportMode === 'update' && existingExportFile) {
                 const formData = new FormData();
                 formData.append('file', existingExportFile);
+                if (transferProjectId) formData.append('project_id', transferProjectId);
                 transferVariantIds.forEach(variantId => formData.append('variant_id', variantId));
                 res = await fetch(new URL(
                     import.meta.env.VITE_API_URL + '/api/assessments/review/export-update',
@@ -566,7 +567,7 @@ function Review({ variantId, projectId, onAssessmentChanged }: Readonly<Props>) 
             console.error('Export error:', err);
             showMessage('Failed to export review data.', 'error');
         }
-    }, [existingExportFile, exportMode, showMessage, variantNames, transferVariantIds, transferFormat]);
+    }, [existingExportFile, exportMode, showMessage, transferProjectId, variantNames, transferVariantIds, transferFormat]);
 
     const handleImportReview = useCallback(() => {
         setTransferMode(null);
--- a/src/bin/cmd_assessments.py
+++ b/src/bin/cmd_assessments.py
@@ -14,6 +14,7 @@ from ..helpers.assessment_io import (
     import_statements,
     build_custom_data_export,
     import_custom_data,
+    reconcile_review_export,
 )
 from ..models.assessment import Assessment as DBAssessment
 from ..models.variant import Variant as DBVariant
@@ -61,8 +62,10 @@ def _print_custom_data_import_result(result: dict) -> None:
 @click.option("--project", "-p", required=True, help="Project name.")
 @click.option("--variant", "-v", default=None,
               help="Variant name. If empty, all project variants are exported.")
+@click.option("--amend", is_flag=True,
+              help="Update the existing output file while preserving stable review ordering.")
 @with_appcontext
-def export_custom_vulnscout_data_command(output_dir: str, project: str, variant: str | None) -> None:
+def export_custom_vulnscout_data_command(output_dir: str, project: str, variant: str | None, amend: bool) -> None:
     """Export custom VulnScout JSON data for one or all project variants."""
     project_obj = resolve_project(project)
     if variant:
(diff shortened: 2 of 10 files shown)


src/helpers/assessment_io.py
```diff
@@ -160,8 +161,234 @@ def is_openvex_doc(doc: object) -> bool:
...
+    if detect_review_export_format(current) != export_format:
+        raise ValueError("Existing file format does not match the generated export")
+    assert isinstance(existing, dict)
+
+    result = dict(existing)
+    if export_format == "openvex":
+        for key, value in current.items():
+            if key != "@id":
+                result[key] = value
+        result["statements"] = _reconcile_records(
```
sfl-pe-sync, line 379: ERROR: Reconciliation preserves the existing OpenVEX `@id` but overwrites its `version` with the freshly generated document's value (the generated exports start again at version 1). Any amended document therefore reuses an existing identity with a reset/non-incremented revision, producing an invalid OpenVEX revision whenever content or the timestamp changes. Preserve the existing identity and set `version` to `existing["version"] + 1` (with the required integer validation), and add a regression test starting from a version greater than 1.

src/routes/assessments.py
```diff
@@ -807,6 +809,75 @@ def export_review_custom_data() -> ResponseReturnValue:
...
+            return {"error": str(error)}, 400
+
+        raw_variant_ids = request.form.getlist('variant_id')
+        if not raw_variant_ids:
+            return {"error": "At least one variant_id is required"}, 400
+        variant_ids: list[UUID] = []
+        for raw_variant_id in raw_variant_ids:
+            variant_id, err = parse_uuid_or_400(raw_variant_id, "variant_id")
+            if err:
+                return err
```
sfl-pe-sync, line 859: ERROR: The new endpoint validates only that each UUID names some variant; it never binds those variants to the Review page's project or even requires all submitted variants to belong to one project. A crafted multipart request can therefore reconcile/export assessment data from unrelated projects in one response despite the UI's project-scoped selection. Send the current `project_id` and verify every variant belongs to it before building the export (and test mixed-project IDs).

frontend/src/pages/Review.tsx
```diff
@@ -484,8 +490,43 @@ function Review({ variantId, projectId, onAssessmentChanged }: Readonly<Props>)
         setTransferVariantIds(defaultVariantId ? [defaultVariantId] : []);
     }, [transferMode, transferVariants, variantId]);
 
+    const changeExportMode = useCallback((mode: 'normal' | 'update') => {
```
sfl-pe-sync, line 493: WARNING: Changing export mode clears the selected file state but does not invalidate an in-flight `FileReader`. If a file read finishes after the user switches back to normal mode or closes/reopens the modal, its callback can repopulate the stale file and change `transferFormat`, causing a later normal export to use the wrong format. Increment `existingExportSelection.current` whenever the mode/modal is reset, or abort the active reader, and cover that transition with an async-ordering test.

sfl-pe-sync (changes requested):
The amend flow has blocking correctness and scoping defects: OpenVEX revisions reset their version while retaining the document ID, and the new API does not enforce project ownership for submitted variants. The frontend also permits a stale file read to mutate later export state.

src/helpers/assessment_io.py
```diff
@@ -160,8 +161,238 @@ def is_openvex_doc(doc: object) -> bool:
...
+    candidates: deque[int],
+    current: list[Any],
+) -> int | None:
+    """Return the sole timestamp match among candidates, if one exists."""
+    timestamp_candidates = [
+        index for index in candidates
+        if current[index].get("timestamp") == old_record.get("timestamp")
+    ]
+    return timestamp_candidates[0] if len(timestamp_candidates) == 1 else None
+
```
sfl-pe-sync, line 281: ERROR: A normal assessment edit changes both its content and timestamp, so neither the exact-content pass nor this timestamp pass matches it to the existing record. Even when the reconciliation identity has exactly one old/current candidate, the old entry is dropped and the generated entry is appended, changing stable order and discarding any preserved `x-` extensions. Add a final identity-only match when exactly one candidate remains (while keeping the stricter passes for duplicate identities), and test an edited record whose timestamp changes.

src/bin/cmd_assessments.py
```diff
@@ -83,6 +86,8 @@ def export_custom_vulnscout_data_command(output_dir: str, project: str, variant:
         raise click.ClickException("No custom VulnScout data to export.")
     os.makedirs(output_dir, exist_ok=True)
     output_path = os.path.join(output_dir, f"custom_vulnscout_data_{filename_label}{_JSON_SUFFIX}")
```
sfl-pe-sync, line 88: ERROR: Both CLI amend paths still reject an empty generated export before reaching reconciliation (`No custom VulnScout data to export` here and `No custom assessments to export` in the OpenVEX command). After the last selected assessment/CVSS/estimate is deleted, `--amend` therefore cannot remove the stale records from the tracked file, while the web endpoint supports this case. Keep the empty-data rejection only for normal exports; amend mode should build/reconcile the empty current document so deletions persist, with CLI tests for both formats.

vulnscout
```diff
@@ -348,6 +348,8 @@ parse_and_run() {
                 ensure_running; EXPORT_ARGS+=(--export-custom-vulnscout-data); shift ;;
             export-custom-openvex-assessments|--export-custom-openvex-assessments)
                 ensure_running; EXPORT_ARGS+=(--export-custom-openvex-assessments); shift ;;
```
sfl-pe-sync, line 392: WARNING: The host CLI now accepts `--amend`, but its static `show_help` output and the existing export documentation in `doc/source/vulnscout-script.md` are unchanged. Users cannot discover the new option or its requirement that the expected output file already exist. Add `--amend` beneath both custom export commands in `--help` and document the filename/path behavior and failure mode.

sfl-pe-sync (changes requested):
The amend flow still loses stable ordering/extensions when an edited record receives a new timestamp, and the CLI cannot reconcile deletions when the selected export becomes empty. The host CLI help also omits the new option.

src/helpers/assessment_io.py
```diff
@@ -370,8 +370,12 @@ def reconcile_review_export(existing: object, current: dict[str, Any]) -> dict[s
     result = dict(existing)
     if export_format == "openvex":
         for key, value in current.items():
-            if key != "@id":
+            if key not in {"@id", "version"}:
                 result[key] = value
+            if type(existing["version"]) is int:
```
sfl-pe-sync, line 375: WARNING: The new validation accepts any exact `int`, including `0` and negative values, so amending an input with `version: -1` emits `version: 0`, which violates OpenVEX's positive-version requirement; a missing `version` also raises `KeyError` instead of the intended validation error. Read the field with an explicit presence check and require `type(version) is int and version >= 1` before incrementing, with regression tests for missing, zero, and negative versions.

sfl-pe-sync (dismissed):
The project-scoping and OpenVEX revision fixes are sound, but version validation still permits malformed inputs to produce an invalid OpenVEX revision (or an inconsistent exception).

ValentinBoudevinSFL (changes requested):
Helper and Readme should be updated too

ValentinBoudevinSFL (approved):

PR #528, opened 2026-08-19
Clarify project-only vulnerability refresh

reh-sfl:
## Description

Add and document the project-scoped `--refresh-data-vulnerability` CLI flow. Standalone refresh accepts no scope or `--project`; combining it with `--variant` returns an explicit usage error.

## Checklist

- [x] READY
- [x] Tests added or updated
- [x] Documentation updated

## Related issue

No linked issue.

### DIFF
--- a/src/bin/cmd_process.py
+++ b/src/bin/cmd_process.py
@@ -25,10 +25,12 @@ from ..extensions import batch_session, db as _db
 import click
 import json
 import os
-from typing import TYPE_CHECKING
+from typing import Callable, Iterable, TYPE_CHECKING
 from flask.cli import with_appcontext
 from sqlalchemy import and_, exists
 from ._common import DEFAULT_VARIANT_NAME, resolve_project_variant
+from ..controllers.projects import ProjectController
+from ..helpers.export_scope import compute_export_scope
 
 if TYPE_CHECKING:
     from ..controllers.packages import PackagesController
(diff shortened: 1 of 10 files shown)


src/bin/cmd_process.py
```diff
@@ -366,7 +462,24 @@ def _run_main() -> ControllersCache:
     # In batch / CI mode (INTERACTIVE_MODE != "true") we run it here so that
     # EPSS scores are available for --match-condition evaluation.
     interactive_mode = get_bool_env("INTERACTIVE_MODE", False)
-    if not interactive_mode:
+    observations_populated = False
```
sfl-pe-sync, line 468: ERROR: `latest_scan` was read before `read_inputs()` runs, so an input-plus-refresh invocation populates observations against the previous scan (or skips them on the first import). Because `observations_populated` then suppresses the final re-query, the newly imported scan never receives its observations. Re-fetch `ScanModel.get_latest()` after ingestion before this call, or keep the existing end-of-processing population path.

tests/CI_tests/vulnscout_wrapper_test.sh
```diff
@@ -0,0 +1,77 @@
...
+
+"$ROOT_DIR/vulnscout" --refresh-vulnerability-data
+grep -q -- '/scan/src/entrypoint.sh --refresh-vulnerability-data' "$VULNSCOUT_TEST_LOG"
+
+: > "$VULNSCOUT_TEST_LOG"
+"$ROOT_DIR/vulnscout" --project cli --refresh-vulnerability-data
+grep -q -- '/scan/src/entrypoint.sh --project cli --refresh-vulnerability-data' "$VULNSCOUT_TEST_LOG"
+
+: > "$VULNSCOUT_TEST_LOG"
+"$ROOT_DIR/vulnscout" --project cli --variant default --refresh-vulnerability-data
```
sfl-pe-sync, line 48: ERROR: This standalone `--variant` invocation is intentionally rejected by the host wrapper before `podman exec`; under `set -e` the new `test_ci` target therefore exits here every time. Assert a nonzero status and that no container execution was logged instead of expecting the arguments to be forwarded.

vulnscout
```diff
@@ -54,8 +56,8 @@ Note: The container runs as a daemon. All commands auto-start it if needed and
 leave it running afterward (including --serve). Use 'stop' or '--stop' to remove it.
 
 Global flags:
```
sfl-pe-sync, line 58: WARNING: The help text says `--variant` scopes a standalone refresh when combined with `--project`, but the implementation and the rest of the documentation explicitly reject every standalone refresh containing `--variant`. Describe it as input-command scope and state that standalone refresh does not accept it.

sfl-pe-sync (changes requested):
The project-scoped refresh is generally wired through the CLI, but input refreshes attach observations to stale scan state, and the newly added wrapper test contradicts the intended variant rejection and will fail the CI target. The host help also documents that restriction incorrectly.

PR #529, opened 2026-08-19
feat(operations): replace polling with a server-side queue and one SSE stream

reh-sfl:

* Add an in-process operation queue that owns scan and refresh sequencing, so ordering is no longer reconstructed in the browser on every page load.
* Add `GET /api/events/stream`, one Server-Sent Events connection that carries every operation update: a full `snapshot` first, then `operation` / `operation_removed` deltas, a `heartbeat` every 15 s, and `bye` on shutdown. `Last-Event-ID` replays from a bounded buffer, or triggers a fresh snapshot when the gap is too wide.
* Add `POST /api/operations`, which accepts a whole batch of scans and refreshes in one request and expands it into individually tracked operations. Scans run before refreshes; each family keeps its canonical order (`grype → nvd → osv → scc`, then `nvd → epss → ghsa → euvd`).
* Normalise the five previously disconnected progress mechanisms into one `OperationRegistry`: the four `ProgressTracker` singletons, the four per-variant `_*_scans_in_progress` closure dicts, `_upload_status`, `_export_jobs`, and the boot enrichment thread that previously reported nothing at all.
* Extract the four scan bodies out of `routes/scan_triggers.py` and the four refresh bodies out of `routes/bulk_refresh.py` into `controllers/scan_jobs.py` and `controllers/refresh_jobs.py`, reporting through a `JobContext` instead of mutating progress dictionaries directly.
* Add cooperative cancellation for scans, which previously had none — only the four bulk refreshes could be cancelled. A cancel request also terminates the Grype subprocess.
* Run three lanes that reproduce the concurrency the frontend used to enforce by hand: `pipeline` (all scans and refreshes, strictly serial), `export` (two at a time), `upload` (one thread each).
* Delete 21 endpoints: 9 status/progress, 8 triggers, and 4 cancels.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Open the app in two browser tabs, launch a two-variant `grype + nvd + EPSS` batch from the wizard, confirm both tabs show the identical queue, reload one mid-run and confirm it restores the full queue including still-`queued` items, then cancel a queued scan from one tab and watch it disappear in the other. DevTools should show exactly one open `/api/events/stream` request and no status or progress requests.

### Related Issue

No linked issue.

## Pull Request Checklist


src/routes/documents.py
```diff
@@ -683,34 +710,22 @@ def export_documents() -> ResponseReturnValue:
         filename = f"{project_name}_{suffix}_export.zip"
 
         if body.get("async") is True:
-            if not _export_capacity.acquire(blocking=False):
+            active_exports = sum(
```
sfl-pe-sync, line 713: ERROR: The previous `BoundedSemaphore` made the six-job export limit atomic, but this count-then-submit check is not serialized. Concurrent requests can all observe fewer than `EXPORT_MAX_ACTIVE_JOBS` and enqueue, allowing an unbounded number of CPU/disk-heavy exports (each permitted to produce a 512 MiB archive). Reserve capacity atomically—e.g. retain a semaphore or perform the active-count check plus registry creation under one lock—and release it when each operation settles.

src/controllers/operation_registry.py
```diff
@@ -0,0 +1,199 @@
...
+            if status == STATUS_RUNNING and operation["started_at"] is None:
+                operation["started_at"] = _now()
+            if status in TERMINAL_STATUSES:
+                if operation["finished_at"] is None:
+                    operation["finished_at"] = _now()
+                self._finished_at[op_id] = time.monotonic()
+            snapshot = dict(operation)
+            snapshot["logs"] = list(operation["logs"])
+        operation_events.publish("operation", snapshot)
+        return snapshot
```
sfl-pe-sync, line 135: ERROR: Registry mutation and event publication are separated by releasing `_lock`, so concurrent updates can be published in the opposite order from the states they represent. This is reachable when a worker finishes while another request cancels/relaunches the same deterministic operation ID; a newer queued state can be delivered before an older terminal snapshot, regressing the client and causing `waitForQueue` to track the wrong `queue_id`. Serialize mutation plus sequence assignment/publication, or add a per-operation generation/revision that clients use to reject stale deltas.

src/controllers/operation_registry.py
```diff
@@ -0,0 +1,199 @@
...
+            del self._operations[op_id]
+            self._finished_at.pop(op_id, None)
+        operation_events.publish("operation_removed", {"op_id": op_id})
+        return True
+
+    def prune(self) -> None:
+        """Drop terminal operations older than :data:`TERMINAL_TTL_SECONDS`."""
+        cutoff = time.monotonic() - TERMINAL_TTL_SECONDS
+        with self._lock:
+            expired = [op_id for op_id, at in self._finished_at.items() if at < cutoff]
```
sfl-pe-sync, line 152: WARNING: `prune()` is never invoked by the stream, snapshot endpoint, or queue. Random-ID upload operations (and failed/undownloaded exports) therefore remain forever with up to 500 log lines each, despite the advertised one-hour TTL, and every snapshot grows without bound. Run pruning periodically or before producing snapshots/list responses, with archive cleanup coordinated for export results.

sfl-pe-sync (changes requested):
The server-side queue migration has two blocking concurrency regressions: export admission is no longer atomic, and registry events can be emitted out of mutation order. Terminal operation pruning is also not wired up, causing unbounded in-process state growth.

src/routes/events.py
```diff
@@ -0,0 +1,155 @@
...
+def init_app(app: Flask) -> None:
+
+    @app.route(STREAM_PATH)
+    def operation_event_stream() -> ResponseReturnValue:
+        """Stream every operation update over a single connection.
+
+        Frames are ``snapshot`` (full state, sent first), ``operation`` and
+        ``operation_removed`` deltas, ``heartbeat`` every 15 seconds, and
+        ``bye`` on shutdown. Reconnecting clients that send ``Last-Event-ID``
+        are replayed from the buffer, or resnapshotted when the gap is too wide.
```
sfl-pe-sync, line 137: ERROR: Flask automatically serves `HEAD` for this GET route, so each `HEAD /api/events/stream` increments `_active_streams` but never iterates `_stream()`. Closing an unstarted generator does not execute its `finally`, leaving the slot permanently reserved; 32 unauthenticated HEAD requests can therefore make every real EventSource connection receive 503 until restart. Reject HEAD before acquiring capacity, or acquire/release the slot through lifecycle handling that also covers responses whose body is never iterated, and add a regression test for HEAD/non-consumed responses.

sfl-pe-sync (changes requested):
The queue/SSE migration has a blocking stream-capacity leak: HEAD requests can permanently exhaust all SSE slots and disable operation updates until restart.

frontend/src/pages/ScanHistory.tsx
```diff
@@ -1301,62 +1259,55 @@ function ScanHistory({ variantId, projectId, onScanComplete }: Readonly<Props>)
...
-        // An unavailable baseline must never block the scans: fall back to refreshing the whole selected scope.
+
+        const loadSelectedVulnerabilities = () => variantIds.length > 1
+            ? Vulnerabilities.list(undefined, projectId, undefined, undefined, variantIds, 'union')
+            : Vulnerabilities.list(variantIds[0], projectId);
+        // Baseline taken before the scans so only newly discovered vulnerabilities get refreshed.
+        // An unavailable baseline must never block the scans: fall back to refreshing the whole scope.
         const existingVulnerabilityIds = new Set(refreshTypes.length === 0 ? [] : (
             await loadSelectedVulnerabilities().catch(() => [])
         ).map(vulnerability => vulnerability.id));
```
sfl-pe-sync, line 1349: ERROR: The wizard enqueues only `scanJobs` here and waits for that batch to finish before enqueueing its requested refreshes in a second request. During the scans, another tab can enqueue (for example) `refresh:epss`; it will be accepted behind these scans, then the wizard's later EPSS enqueue receives 409 and the user-selected refresh is skipped. This also means the queue never contains the advertised whole `grype + nvd + EPSS` batch. Reserve the refresh operations in the original atomic batch and resolve their post-scan target IDs when they execute (or otherwise add a backend deferred-target job) so no intervening request can steal the refresh slot.

frontend/src/components/OperationQueueModal.tsx
```diff
@@ -71,27 +135,30 @@ function OperationQueueModal({ isOpen, onClose }: Readonly<Props>) {
...
+                                return (
+                                    <OperationQueuePanel
+                                        key={operation.op_id}
+                                        operation={operation}
+                                        icon={icon}
+                                        colors={colors as never}
+                                        positionLabel={positionLabelFor(operation, operations)}
+                                        onDismiss={() => void Operations.dismiss(operation.op_id)}
+                                        onCancel={isActive(operation) && !cancelRequested.has(operation.op_id)
+                                            ? () => requestCancel(operation.op_id)
```
sfl-pe-sync, line 157: ERROR: This presents cancellation for every active operation, but running upload and export runners never call `check_cancelled()` (nor install a cancel hook). The API therefore returns success while an SBOM import still commits all data or an export still builds its archive; `_run_body` then labels the completed side effect `cancelled`, and the upload caller does not reload the imported data. Either implement cooperative cancellation/cleanup in those runners or expose a cancellable capability and reject/hide cancellation for unsupported operation kinds.

sfl-pe-sync (changes requested):
The server-side queue rewrite has blocking sequencing and cancellation contract defects: wizard refreshes can be displaced by another request, and accepted upload/export cancellations do not stop their side effects.

ValentinBoudevinSFL (changes requested):
Commits need a description

src/controllers/operation_registry.py
```diff
@@ -0,0 +1,201 @@
...
+            "label": label,
+            "lane": lane,
+            "scope": scope,
+            "status": STATUS_QUEUED,
+            "progress": {"current": 0, "total": 0, "message": "Queued"},
+            "logs": [],
+            "error": None,
+            "queue_id": queue_id,
+            "position": position,
+            "options": options or {},
```
sfl-pe-sync, line 94: ERROR: `options` contains the raw execution inputs (`ids`, `exclude_ids`, and variant lists), and every `update()` republishes a shallow copy of the whole operation. A large refresh therefore retransmits the complete vulnerability-ID set on every progress frame to every tab; deferred refreshes also mutate the same dict shared with `JobContext`, so the resolved IDs silently enter later frames. This can turn a 10k-ID refresh into hundreds of megabytes of SSE traffic, overflow clients, and make each recovery snapshot equally large. Keep runner inputs private to the queued job/`JobContext`, and expose only bounded, UI-required metadata in registry snapshots and events.

src/controllers/refresh_jobs.py
```diff
@@ -0,0 +1,444 @@
...
+def _nvd_sleep_interval() -> float:
+    """Seconds between NVD REST calls: 6 s without an API key, 0.6 s with one."""
+    return 0.6 if os.getenv("NVD_API_KEY") else 6.0
+
+
+def _safe_commit(label: str) -> None:
+    """Commit, rolling back on failure, then expunge so the identity map stays bounded."""
+    try:
+        db.session.commit()
+    except Exception as exc:
```
sfl-pe-sync, line 60: ERROR: Commit failures are logged and swallowed here, so a locked/full/unavailable database rolls back the current batch but the refresh continues and ultimately reaches `STATUS_DONE` with a “complete” message. The client then reloads and reports success even though updates were lost. Roll back and re-raise the commit exception (with an actionable sanitized message) so the queue marks the operation as failed; previously committed chunks can remain as an explicitly reported partial result.

src/routes/operations.py
```diff
@@ -0,0 +1,412 @@
...
+
+        OpenAPI:
+        body JsonObject optional Batch of jobs to queue.
+        response 202 JsonObject Queue accepted with the created operations.
+        response 400 Error Invalid job specification.
+        response 404 Error Referenced variant not found.
+        response 409 Error One of the requested operations is already active.
+        """
+        body = request.get_json(force=True, silent=True) or {}
+        jobs = body.get("jobs")
```
sfl-pe-sync, line 340: WARNING: A valid non-object JSON body such as `[1]` or `"jobs"` is truthy, so `body.get("jobs")` raises `AttributeError` and returns a 500 instead of the documented 400. Validate that the decoded body is a dictionary before accessing it and add a regression case for scalar/non-empty-array JSON.

src/routes/documents.py
```diff
@@ -719,49 +734,37 @@ def update_progress(current: int, total: int, element_name: str) -> None:
...
-                        _prune_export_jobs()
+                    with _export_archive_lock:
+                        _prune_export_archives()
                         _reserve_retained_archive(os.path.getsize(archive_path))
-                        _export_jobs[job_id].update({
-                            "status": "done",
-                            "progress": "Export ready",
+                        ctx.message("Export ready")
+                        ctx.set_result({
                             "archive_path": archive_path,
```
sfl-pe-sync, line 742: WARNING: Storing `archive_path` in the public operation result exposes the server's temporary filesystem path through both the SSE stream and `GET /api/operations`; the old status response explicitly omitted this internal field. Keep artifact paths in a private server-side map keyed by `op_id` and publish only `download_ready` and the safe filename.

sfl-pe-sync (changes requested):
The unified queue has blocking reliability and scalability defects: refreshes can report success after failed commits, and unbounded execution inputs are repeatedly broadcast in every SSE update. The operation API also mishandles non-object JSON, and export events disclose internal temporary paths.

reh-sfl:
This PR has been split into a 6-PR stack so it can be reviewed in reasonable pieces. Nothing here changed — the stack is built from this branch's own commits, and the tip of the stack (#553) is commit-identical to `full-sse-operations`.

| # | PR | Base | Contents | Diff |
|---|----|------|----------|------|
| 1 | #548 | `staging` | operation queue core; scan/refresh job bodies extracted | +2987 / −951 |
| 2 | #549 | #548 | `POST /api/operations` and `GET /api/events/stream` | +1677 / −55 |
| 3 | #550 | #549 | delete the 21 legacy polling endpoints | +6 / −5366 |
| 4 | #551 | #550 | route exports and uploads through the queue | +536 / −398 |
| 5 | #552 | #551 | frontend SSE operation store; queue rendered from it | +1595 / −209 |
| 6 | #553 | #552 | migrate remaining frontend workflows; delete the polling client | +942 / −4485 |

All six are drafts. Review and merge in order — each targets the one before it, so #548 is the only one that lands on `staging` directly and the rest collapse into it as they merge.

This PR stays open as the integration reference until #553 merges, at which point it can be closed as superseded.

PR #531, opened 2026-08-21
Group Assessment

sfl-zwang:

This PR turns today's assessment grouping in VulnModal into a persistent state in the database. It adds a new table that links a group to an assessment, allowing you to retrieve assessments by group ID. Each assessment in the history now includes a copy button to copy either the assessment ID or the group ID, depending on whether the displayed assessment is a single assessment or one targeting multiple variants and packages.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

All operations related to assessments should work the same as before.

VulnModal and the Review page should load the assessments as usual.

A new copy button is added next to each assessment:
image

image

A copied text is displayed when clicking the copy button:

image

### Additional notes

### Related Issue

## Pull Request Checklist


src/migrations/versions/x0a1b2c3d4e5_add_assessment_group_members.py
```diff
@@ -0,0 +1,87 @@
...
+               f.vulnerability_id AS vuln_id,
+               a.timestamp, a.status, a.simplified_status, a.status_notes,
+               a.justification, a.impact_statement, a.workaround, a.origin
+        FROM assessments a
+        JOIN findings f ON f.id = a.finding_id
+    """)).mappings().all()
+
+    buckets: dict[tuple, list] = {}
+    for row in rows:
+        key = (
```
sfl-pe-sync, line 93: ERROR: The backfill key omits project identity, so assessments for the same CVE with identical imported timestamps/content in variants belonging to different projects are assigned one group. A project-filtered read then exposes only part of that group, but delete/reconcile/approve/reject load every member by group_id and can mutate the other project's assessments. Join through variants and include project_id in the bucket key (and add a cross-project migration regression test) so a persisted group can never cross a project boundary.

frontend/src/components/VulnModal.tsx
```diff
@@ -221,6 +218,52 @@ type VariantScopedSnapshot = {
...
+    // therefore the status table above the history — reflect the write.
+    const refreshAllVulnAssessments = useCallback(async (): Promise<Assessment[] | null> => {
+        const projectQuery = projectId ? `?project_id=${encodeURIComponent(projectId)}` : '';
+        try {
+            const r = await fetch(import.meta.env.VITE_API_URL + `/api/vulnerabilities/${encodeURIComponent(vuln.id)}/assessments${projectQuery}`, { mode: 'cors' });
+            const data = await r.json();
+            if (Array.isArray(data)) {
+                const fullAssessments = data.flatMap(asAssessment).filter((a): a is Assessment => !Array.isArray(a));
+                const fullById = new Map(fullAssessments.map(a => [a.id, a]));
+                vuln.assessments = vuln.assessments.map(a => fullById.get(a.id) ?? a);
```
sfl-pe-sync: ERROR: This refresh only replaces assessments whose IDs were already present in `vuln.assessments`; it never adds rows created by reconcile or removes rows deleted by reconcile. `saveEditedAssessment` immediately builds and patches the vulnerability status summary from that stale array, so changing a group's package/variant targets can leave the displayed and cached vulnerability status based on deleted records or missing new records. Reconcile `vuln.assessments` from the returned full list using the modal's scope, and compute the summary from that reconciled list.

sfl-pe-sync (changes requested):
Persistent grouping is not safe to merge yet: the migration can create cross-project groups, and group reconciliation can leave vulnerability status derived from stale assessment membership.

ValentinBoudevinSFL (changes requested):
Looks very nice, in addition to VE comments I would mention a require change in review tab:

VulnModal allow to copy an assessment ID: 

Image

But the ID should be also present in a column in the review tab:

Image

Main purpose of the AI review feature is to review custom assessments. Should be straightforward to copy an ID. 
Add a new column in this tab

reh-sfl:
@sfl-zwang @ValentinBoudevinSFL I would suggest the following:
1. This currently copies either a group_id or assessment_id depending on the type (grouped vs single). I would suggest adding a toggle in the section header to show a grouped or ventilated view in case the user wants to target a specific node.
2. I'm not a big fan of adding a new column that displays an ID, we need to think about a way to display this in a elegant manner.

src/migrations/versions/x0a1b2c3d4e5_add_assessment_group_members.py
WilsonZiweiWang: Fixed in 74c3464. `backfill_groups()` now `LEFT JOIN`s `variants` and prepends `v.project_id` to the bucket key, so a persisted group can never cross a project boundary. Assessments with `variant_id IS NULL` have no project and fall into their own `NULL` bucket, so they group only with each other and never join a project's group.

Added `tests/unit_tests/test_assessment_group_members_migration.py` with four cases: cross-project rows never share a group, same-project rows still group, variant-less rows stay isolated, and unique rows stay ungrouped. Verified the cross-project and variant-less cases fail when the `project_id` key component is removed.

frontend/src/components/VulnModal.tsx
WilsonZiweiWang: Fixed in 74c3464. `refreshAllVulnAssessments` no longer merges by id — it rebuilds `vuln.assessments` from the returned full list, scoped the way the modal is scoped (filtered to `variantId` when set, otherwise the whole project-scoped list), so reconcile-created rows are added and reconcile-deleted rows are dropped. It returns that reconciled list, and `saveEditedAssessment` now feeds the returned list to `buildStatusSummary`/`patchVuln` instead of `[...vuln.assessments]`, falling back to `vuln.assessments` only if the refresh request itself fails.

sfl-zwang:
@ValentinBoudevinSFL 

A copy button is added to the action cell to copy either an assessment ID or a group ID.

image

src/models/assessment_group_member.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def __repr__(self) -> str:
+        return (
+            f"<AssessmentGroupMember assessment_id={self.assessment_id}"
+            f" group_id={self.group_id}>"
+        )
+
+    @staticmethod
+    def create_group(
+        assessment_ids: "list[uuid.UUID]",
```
sfl-pe-sync, line 130: ERROR: `create_group` accepts arbitrary assessment IDs (and an arbitrary existing `group_id`) without enforcing the group's vulnerability/project/content invariants. The new batch path groups only by `vuln_id`, so one batch containing the same CVE for variants in two projects—or two statuses for that CVE—creates one persisted group; the incremental `group_id` path likewise checks only the vulnerability, not the project or shared content. Group reads then show the first member's content, while delete/reconcile/approve/reject mutate every member, including the other project's or semantically different rows. Enforce one project, vulnerability, origin/content tuple per group centrally (preferably with group metadata/constraints), key batch grouping by those invariants, validate joins, and add cross-project and mixed-content write tests.

src/models/assessment.py
```diff
@@ -345,13 +345,21 @@ def set_workaround(self, workaround: str, timestamp: Optional[str] = None) -> No
 
     def to_dict(self) -> dict:
         ts = ensure_utc_iso(self.timestamp)
+        # Function-local import: the two model modules would otherwise import
+        # each other at module load time.
+        from flask import has_app_context
+        from .assessment_group_member import AssessmentGroupMember
+        group_id = None
+        if getattr(self, "id", None) is not None and has_app_context():
+            group_id = AssessmentGroupMember.get_group_id(self.id)
```
sfl-pe-sync: WARNING: Every `Assessment.to_dict()` now executes a separate membership query. List/export endpoints serialize collections this way, and `build_groups()` already bulk-loads memberships but then calls `to_dict()` for every assessment, turning an N-row review/history response into N additional database queries despite its three-query goal. Eager/bulk-load membership once and pass or expose the already-loaded `group_id` during serialization instead of querying from `to_dict()`.

sfl-pe-sync (changes requested):
The persistent grouping paths still permit cross-project and semantically heterogeneous groups, making group-wide mutations unsafe; serialization also introduces an avoidable N+1 query regression.

src/migrations/versions/x0a1b2c3d4e5_add_assessment_group_members.py
```diff
@@ -0,0 +1,97 @@
...
+        JOIN findings f ON f.id = a.finding_id
+        LEFT JOIN variants v ON v.id = a.variant_id
+    """)).mappings().all()
+
+    buckets: dict[tuple, list] = {}
+    for row in rows:
+        key = (
+            row["project_id"],
+            row["vuln_id"], row["timestamp"], row["status"],
+            row["simplified_status"], row["status_notes"], row["justification"],
```
sfl-pe-sync, line 96: ERROR: The backfill groups rows without including `responses`, even though the group serializer exposes only `head.responses` and reconcile applies one response set to every member. Two imported assessments with the same timestamp/status/notes but different VEX responses will therefore be fused; one response is hidden and later group operations mutate/delete both as if they were one action. Include a stable representation of `responses` in the bucket key and add a migration regression test with otherwise-identical rows carrying different responses.

src/routes/assessments.py
```diff
@@ -1453,39 +1543,109 @@ def delete_assessment(assessment_id: str) -> ResponseReturnValue:
...
-        existing = DBAssessment.get_by_id(assessment_id)
-        if existing is None:
-            return {"error": "Assessment not found"}, 404
-        if existing.origin != "ai":
-            return {"error": "Not a pending AI assessment"}, 400
-        group, err = _resolve_ai_group(existing)
-        if err is not None:
+    @app.route("/api/assessment-groups/<group_id>/reject", methods=["POST"])
+    def reject_ai_group(group_id: str) -> ResponseReturnValue:
+        """Reject every AI-origin assessment in a group, deleting the whole group.
```
sfl-pe-sync, line 1641: WARNING: This generic group DELETE path does not reject pending AI members, unlike the single-assessment delete guard, `validate_deletions`, and the documented contract. A caller can therefore bypass the required AI reject endpoint and delete an AI group through the custom-assessment operation. Return 400 when any member has `origin == "ai"` and cover the case with an endpoint test.

sfl-pe-sync (changes requested):
The persistent grouping design has blocking compatibility and migration-semantic defects: legacy AI approval APIs are removed, and backfill can merge assessments with different VEX responses. Group DELETE also bypasses the pending-AI guard.

---
- `src/routes/assessments.py:1527` [error]: ERROR: Replacing the existing `/api/assessments/<assessment_id>/approve` and `/reject` routes with group-only routes breaks the published API for CLI/integration clients; only this PR's frontend and tests were migrated, so existing callers now receive 404. Preserve the old endpoints as compatibility wrappers that resolve/promote the addressed assessment's group (or introduce an explicit versioned deprecation) while keeping the new group endpoints.

src/migrations/versions/x0a1b2c3d4e5_add_assessment_group_members.py
ValentinBoudevinSFL: Looks a valid issue to me

ValentinBoudevinSFL (changes requested):
VE error look valid to me. That apart, it makes more sense to me in review tab to have Actions in the following order "Edit/ Copy/ Delete".  
I don't think having Delete in the middle is very common in UI. 

Also the copy button should stay white/gray, same as in VulnModal

sfl-zwang:
> VE error look valid to me. That apart, it makes more sense to me in review tab to have Actions in the following order "Edit/ Copy/ Delete". I don't think having Delete in the middle is very common in UI.
> 
> Also the copy button should stay white/gray, same as in VulnModal

The action cell now:
image

src/helpers/assessment_io.py
```diff
@@ -959,13 +969,17 @@ def _import_assessments(
...
                     result[imported_key] += 1
+                    entry_created_ids.append(db_a.id)
                 except Exception as e:
                     result["errors"].append({
                         "vuln_id": vuln_name,
                         "package": pkg_string_id,
                         "error": str(e),
                     })
 
+            if len(entry_created_ids) > 1:
```
sfl-pe-sync, line 980: ERROR: Grouping only `entry_created_ids` from one imported record does not preserve groups produced by VulnScout's own exports. Both custom and OpenVEX exporters serialize each database assessment separately (one package/variant row, with no group token), so every imported record normally creates one ID and this branch—and the analogous OpenVEX branch—never groups anything; a multi-package or multi-variant group round-trips as unrelated singletons. Export an import-local group identity (or a group-shaped record) and remap all entries sharing it to one new group during import, with export→import tests for multi-package and multi-variant groups.

sfl-pe-sync (changes requested):
The persistent grouping model is improved, but grouped assessments can still be corrupted by single-row updates, and export/import round-trips discard group identity; both block merge.

---
- `src/routes/assessments.py:1560` [error]: ERROR: This compatibility update still mutates only the addressed assessment even when it belongs to a persisted group. The Review page invokes it once per member, so a failed later PUT/DELETE/POST leaves a group whose members have different content (or only part of the requested targets), while group reads expose only the first member and later actions mutate the whole corrupted group. Resolve membership here and either reject grouped rows in favor of `/reconcile` or update the whole group transactionally; switch `Review.handleSaveEdit` to one reconcile request as VulnModal already does.

src/migrations/versions/x0a1b2c3d4e5_add_assessment_group_members.py
sfl-pe-sync: Confirmed fixed in the latest patch: the migration key now canonicalizes `responses`, including tests for differing values, order-only differences, and NULL/empty equivalence.

src/helpers/assessment_io.py
ValentinBoudevinSFL: Not revelant

ValentinBoudevinSFL (approved):
LGTM

PR #532, opened 2026-08-21
vulnscout: update the helper for consistency

ValentinBoudevinSFL:

VulnScout CLI and entrypoints helper needed some clarification. 

Now, all commands are described to be link with --project, --project and --variants, or none.

Also updated the commands `--export-custom-openvex-assessments  ` Export custom OpenVEX assessments and `--import-custom-openvex-assessments` to use `default` project if not specified

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

```
./vulnscout --help
```

### DIFF
--- a/src/bin/cmd_scans.py
+++ b/src/bin/cmd_scans.py
@@ -100,7 +100,7 @@ def delete_empty_scans_command():
     click.echo(f"Deleted {summary['scans_deleted']} empty scans")
 
 
-@click.command("delete-orphaned-vulnerabilities")
+@click.command("delete-orphaned-vulns")
 @with_appcontext
 def delete_orphaned_vulnerabilities_command():
     """Permanently remove CVEs absent from every project and variant."""
--- a/src/entrypoint.sh
+++ b/src/entrypoint.sh
@@ -95,60 +95,88 @@ export SBOM_CVE_CHECK_AUTO_UPDATE="${SBOM_CVE_CHECK_AUTO_UPDATE:-1}"
 
 show_help() {
     cat <<EOF
-VulnScout Entrypoint
-Usage: docker exec <container> /scan/src/entrypoint.sh [COMMAND] [OPTIONS]
+VulnScout - Vulnerability Scanner
+Usage: /scan/src/entrypoint.sh [--project <name>] [--variant <name>] <command> [options]
 
-Setting:
-  --project <name>          Project name for the next input command (default: 'default')
-    --variant <name>          Variant name for the next input command or standalone refresh (requires --project)
+Project and variant options:
+        --project <name>                Select a project
+        --variant <name>                Select a variant within --project
 
-Input commands:
-  --add-spdx <path>         Add an SPDX 2/3 SBOM file or archive
-  --add-cve-check <path>    Add a Yocto CVE check JSON file
-  --add-yocto-vex <path>    Add a Yocto VEX JSON file
-  --add-openvex <path>      Add an OpenVEX JSON file
-  --add-cdx <path>          Add a CycloneDX file
-  --add-grype <path>        Add a Grype results file
-  --perform-grype-scan      Perform a Grype scan on the added inputs
-  --perform-nvd-scan        Run an NVD CPE-based vulnerability scan
-  --perform-osv-scan        Run an OSV PURL-based vulnerability scan
-  --perform-sbom-cve-check-scan  Run a sbom-cve-check vulnerability scan
-    --refresh-vulnerability-data  Refresh EPSS, NVD, EUVD, and GHSA data; with no inputs, refreshes all vulnerabilities
-  --add-asset <path>        Stage an image asset for use in report templates
+Commands without --project or --variant:
 
-Scan & output commands:
-  --serve                   Run scan then start interactive web UI (port 7275)
-  --report <template>       Generate a report from a template in /cache/vulnscout/templates/
-  --report <path>           Stage a local report template file and generate a report from it
-  --export-spdx             Export project as SPDX 3.0 SBOM to /scan/outputs/
-  --export-cdx              Export project as CycloneDX 1.6 SBOM to /scan/outputs/
-  --export-openvex          Export project as OpenVEX document to /scan/outputs/
-    --export-custom-vulnscout-data  Export custom VulnScout JSON data to /scan/outputs/
-        --import-custom-vulnscout-data <path>  Import custom VulnScout JSON data
-            --use-original-timestamps    Preserve file timestamps instead of using system time
-            --use-current-timestamps     Use current system time instead of file timestamps
-    --export-custom-openvex-assessments  Export custom OpenVEX assessments for --variant
-    --import-custom-openvex-assessments <path>  Import custom OpenVEX assessments into --variant
-    --export-context          Export AI assessment context (project/variant description, threat model, etc.) to /scan/outputs/context-export.json
-    --import-context <path>   Import AI assessment context from a JSON file (overwrites matching project/variant)
-  --match-condition <expr>  Exit code 2 if condition met (e.g. "cvss >= 9.0")
-  --delete-scan <id>        Delete a past scan by its ID
-    --delete-outdated         Permanently delete outdated packages and assessments
-    --delete-empty-scans      Permanently delete scans with no recorded changes
-    --delete-orphaned-vulnerabilities  Delete CVEs absent from every project and variant
+        Container lifecycle:
+        daemon                          Keep the container running
+        --version                       Show current version
+        --help, -h                      Show this help message
 
-Data retrieval commands:
-  --list-projects           List all projects and their variants
-  --list-scans              List all past scans
-  --json                    Output objects in JSON format
+        Development and web interface:
+        --serve                         Run scan then start web UI (port 7275)
 
-Configuration commands:
-  --config <key> <value>    Set a persistent config value
-  --config-list             Show current configuration
-  --config-clear <key>      Remove a config key
+        Reports and assets:
+        --report <template|path>        Generate a report from a template
+        --add-asset <path>              Stage an image asset for report templates
 
-Container lifecycle:
-  --help, -h                Show this help message
+        Data maintenance:
+        --delete-scan <id>              Delete a past scan by ID
+        --delete-outdated               Delete outdated packages and assessments
+        --delete-empty-scans            Delete scans with no recorded changes
+        --delete-orphaned-vulns         Delete CVEs absent from every project and variant
+
+        Refresh vulnerability data:
+        --refresh-vulnerability-data    Refresh all projects (optionally limited by --project and --variant)
+
+        Data listing:
+        --list-projects                 List all projects and their variants (optionally in JSON format with --json)
+        --list-scans                    List all past scans (optionally in JSON format with --json)
+
+        Configuration:
+        --config <key> <value>          Set a config value
+        --config-list                   Show current configuration
+        --config-clear <key>            Remove a config key
+
+        Import or export AI assessment context for a project:
+        --export-context                Export AI assessment context
+        --import-context <path>         Import AI assessment context
+
+Commands with optional --project and --variant (uses "default" project and variant if not specified):
+
+        Import a SBOM or vulnerability data file into the selected project/variant:
+        --add-spdx <path>               Add an SPDX 2/3 SBOM file or archive
+        --add-cve-check <path>          Add a Yocto CVE check JSON file
+        --add-yocto-vex <path>          Add a Yocto VEX JSON file
+        --add-openvex <path>            Add an OpenVEX JSON file
+        --add-cdx <path>                Add a CycloneDX file
+        --add-grype <path>              Add a Grype results file
+
+        Trigger scanning to detect new vulnerabilities in the selected project/variant:
+        --perform-grype-scan            Run Grype after merging any inputs
+        --perform-nvd-scan              Run an NVD CPE-based scan
+        --perform-osv-scan              Run an OSV PURL-based scan
+        --perform-sbom-cve-check-scan   Run an sbom-cve-check scan
+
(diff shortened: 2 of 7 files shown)


src/bin/cmd_process.py
```diff
@@ -433,12 +445,30 @@ def populate_observations(scan, vulnCtrl, log_prefix: str = "merger_ci") -> None
...
+            scoped_scans = ScanModel.get_by_variant_id(variant.id)
+        else:
+            scope = compute_export_scope(project_id=project.id)
+            scoped_scans = ScanModel.get_by_project(project.id)
+
+    controllers = ControllersCache(scope=scope)
     vulnCtrl: VulnerabilitiesController = controllers.vulnerabilities
     assessCtrl: AssessmentsController = controllers.assessments
-    latest_scan = ScanModel.get_latest()
+    latest_scan = scoped_scans[-1] if scoped_scans else ScanModel.get_latest()
```
sfl-pe-sync: ERROR: The selected scoped scan is not retained through observation creation. In the normal (non-refresh) path, lines 538-540 replace it with `ScanModel.get_latest()`, so `process --project A --variant X` can parse A's inputs but attach the encountered vulnerabilities to a newer scan from project B; shared package/finding rows then create cross-project observations, while A's scan is left unpopulated. An empty scoped scan list also falls back to the global latest scan here. Keep the scoped scan for the entire run and skip/fail when the requested scope has no scan instead of falling back globally.

vulnscout
```diff
@@ -520,16 +543,15 @@ parse_and_run() {
         [ "${REFRESH_VULNERABILITY_DATA:-false}" = true ] && scan_args+=(--refresh-vulnerability-data)
         exec_container --project "$PROJECT" --variant "$VARIANT" "${PENDING_ARGS[@]}" "${scan_args[@]}" "${REPORT_ARGS[@]}" "${EXPORT_ARGS[@]}" || _exit=$?
     elif [ "${REFRESH_VULNERABILITY_DATA:-false}" = true ]; then
-        local -a refresh_args=(--refresh-vulnerability-data)
-        [ "$PROJECT_SPECIFIED" = true ] && refresh_args=(--project "$PROJECT" "${refresh_args[@]}")
+        local -a refresh_args=(--project "$PROJECT" --refresh-vulnerability-data)
```
sfl-pe-sync: ERROR: This unconditionally adds the default project to standalone refreshes, removing the documented global behavior of `./vulnscout --refresh-vulnerability-data`. Existing automation that omits `--project` will now refresh only the `default` project's active vulnerabilities, leaving every other project stale with no indication. Preserve the prior `PROJECT_SPECIFIED` guard (or add an explicit all-projects mode before changing the default) and update the CLI contract intentionally.

src/bin/cmd_scans.py
```diff
@@ -100,7 +100,7 @@ def delete_empty_scans_command():
     click.echo(f"Deleted {summary['scans_deleted']} empty scans")
 
 
-@click.command("delete-orphaned-vulnerabilities")
+@click.command("delete-orphaned-vulns")
```
sfl-pe-sync, line 103: ERROR: Renaming the registered Click command removes the existing `delete-orphaned-vulnerabilities` CLI contract, so direct Flask callers, existing host scripts, and the repository's documented command now fail with “No such command.” This PR is presented as help clarification and provides no migration path. Keep the old command as an alias while adding the shorter spelling, and update the user documentation.

sfl-pe-sync (changes requested):
The help cleanup introduces blocking CLI compatibility regressions and a scoped-processing path that can write observations to another project's latest scan.

src/entrypoint.sh
```diff
@@ -475,7 +504,7 @@ cmd_scan() {
         # passed through to stdout unchanged.
```
sfl-pe-sync, line 503: ERROR: A standalone `--refresh-vulnerability-data --project P` with no `--variant` is now silently converted to `--variant default`. `_run_main()` therefore computes a variant scope and refreshes only `P/default`, skipping every other variant in the requested project even though both help texts describe `--variant` as optional for refresh. Append `--variant` only when one was explicitly supplied; preserve project-wide scope when `VARIANT_NAME` is empty.

sfl-pe-sync (changes requested):
The scope propagation still breaks project-wide standalone refreshes by narrowing them to the default variant, so the change is not ready to merge.

vulnscout
```diff
@@ -520,16 +544,15 @@ parse_and_run() {
...
         [ "$VARIANT_SPECIFIED" = true ] && refresh_args=(--variant "$VARIANT" "${refresh_args[@]}")
         exec_container "${refresh_args[@]}" || _exit=$?
     elif [[ "$DATA_REQUESTED" != "" ]]; then
         exec_container "--$DATA_REQUESTED" "${DATA_ARGS[@]}"
     elif [ -n "$MATCH_CONDITION" ] || [ ${#REPORT_ARGS[@]} -gt 0 ] || [ ${#EXPORT_ARGS[@]} -gt 0 ]; then
         # No new inputs but a match-condition, report, or export was given: run against existing DB data
         ensure_running
-        local -a mc_args=(--project "$PROJECT" --variant "$VARIANT")
+        local -a mc_args=(--project "$PROJECT" --variant "${VARIANT:-default}")
```
sfl-pe-sync: ERROR: This blanket `${VARIANT:-default}` changes every no-input export to `P/default`, not just the OpenVEX assessment commands targeted by this PR. In particular, `vulnscout --project P --export-custom-vulnscout-data` is documented and implemented to export every variant when `--variant` is omitted, but this line now makes `cmd_export_custom_vulnscout_data` receive `VARIANT_NAME=default` and silently produces an incomplete, default-only backup; project-wide SBOM and context exports are narrowed the same way. Build the shared arguments with `--project` only and append `--variant` only when `VARIANT_SPECIFIED` is true; keep the new OpenVEX default inside its dedicated entrypoint handlers.

sfl-pe-sync (changes requested):
The helper refactor still has a blocking scope regression: omitting `--variant` now silently truncates project-wide exports to the default variant.

tests/cli_tests/test_cmd_process.py
```diff
@@ -60,15 +60,38 @@ def test_process_command_invokes_run_main(self, app):
...
+        run_main.assert_called_once_with(
+            refresh_vulnerability_data=True,
+            project_name=None,
+            variant_name=None,
+        )
+
+    def test_process_command_propagates_project_variant_scope(self, app):
+        with patch("src.bin.cmd_process._run_main") as run_main:
+            result = app.test_cli_runner().invoke(args=[
+                "process",
```
sfl-pe-sync: WARNING: This test patches `_run_main`, so it proves only Click argument forwarding and cannot detect whether the new scope is honored once controllers, scans, observations, and refreshes are used. Add an integration test with at least two projects/variants and differently ordered scans that invokes the real processing path and asserts that only the requested scope is read and written, including the no-scan case.

sfl-pe-sync (dismissed):
The new process-scoping behavior lacks a realistic isolation regression test; the added mock-based test cannot exercise the database and scan-selection path.

src/bin/cmd_process.py
```diff
@@ -433,12 +445,30 @@ def populate_observations(scan, vulnCtrl, log_prefix: str = "merger_ci") -> None
...
+    variant_name: str | None = None,
+) -> ControllersCache:
     """Core processing logic (usable both from the CLI command and directly)."""
-    controllers = ControllersCache()
+    scope = None
+    scoped_scans = None
+    if project_name:
+        project = ProjectController.get_by_name(project_name)
+        if project is None:
+            raise click.ClickException(f"project not found: {project_name}")
```
sfl-pe-sync: ERROR: This computes an export scope before the selected scan has been parsed. For a newly merged SBOM, that scan is already the active SBOM but its `SBOMPackage` links are only created later by `read_inputs()`, so `compute_export_scope()` returns an empty `package_ids` set; constructing `VulnerabilitiesController` with it then preloads none of the project's existing vulnerabilities. A package-only SBOM update can therefore omit still-applicable existing findings from post-treatment and subsequent output even when the package remains present. Do not use the pre-ingestion export scope for mutable scan processing; select the scoped scan explicitly, then build/rebuild the read scope after its package links have been persisted (or use dedicated project/variant filtering that does not depend on active package links).

src/entrypoint.sh
```diff
@@ -586,25 +616,19 @@ cmd_import_custom_vulnscout_data() {
...
     cd "$BASE_DIR"
     local output_dir="${OUTPUTS_DIR:-/scan/outputs}"
     flask --app src.bin.webapp db upgrade
     flask --app src.bin.webapp export-custom-openvex-assessments \
-        --project "$PROJECT_NAME" --variant "$VARIANT_NAME" --output-dir "$output_dir"
+        --project "$PROJECT_NAME" --variant "$variant_name" --output-dir "$output_dir"
     setup_user
 }
 
 cmd_import_custom_openvex_assessments() {
```
sfl-pe-sync, line 627: WARNING: The new default-variant import behavior leaves the required backend suite failing at `test_import_custom_openvex_assessments_requires_variant`, which still asserts the removed error contract. Update that end-to-end test to invoke the command without `--variant` and verify the assessments are actually imported into `default` (and add the equivalent export/default-project coverage) so the changed contract is both intentional and mergeable.

sfl-pe-sync (changes requested):
The scoped processing path builds an empty, stale package scope before newly merged SBOM contents are persisted, which can drop applicable existing vulnerability data; the intended OpenVEX default behavior also leaves the backend CI gate failing.

src/entrypoint.sh
```diff
@@ -475,7 +504,8 @@ cmd_scan() {
         # passed through to stdout unchanged.
```
sfl-pe-sync, line 503: ERROR: Every normal container scan now invokes `flask process --project ...`, but the registered `process` command does not define a `--project` option. The final-image Docker test reaches this path and fails with `Error: No such option '--project'`, so SBOM inputs are merged but never processed and the UI is still marked ready with an empty result set. Remove unsupported arguments here, or add and fully wire the corresponding Click options before forwarding them.

sfl-pe-sync (changes requested):
The entrypoint’s new scope forwarding breaks the final container’s normal scan path because `flask process` does not accept the forwarded project option.

vulnscout
```diff
@@ -44,72 +44,95 @@ show_help() {
...
+    Data listing:
+    --list-projects                 List all projects and their variants
+    --list-scans                    List all past scans
+
+    Configuration:
+    --config <key> <value>          Set a config value
+    --config-list                   Show current configuration
+    --config-clear <key>            Remove a config key
+
+Commands with optional --project and --variant (uses default project and variant if not specified):
```
sfl-pe-sync: WARNING: This heading says an omitted `--variant` selects the default variant, but the no-input forwarding below now intentionally omits `--variant` unless it was explicitly supplied, and the backend SBOM export contract exports all variants of the selected project in that case. A user following this help can therefore generate a broader export than expected. Split the commands by their real defaults (for example, document SBOM exports as all project variants when omitted) or make the runtime behavior match the advertised default-variant scope.

vulnscout
```diff
@@ -44,72 +44,95 @@ show_help() {
...
+    --delete-outdated               Delete outdated packages and assessments
+    --delete-empty-scans            Delete scans with no recorded changes
+    --delete-orphaned-vulns         Delete CVEs absent from every project and variant
+
+    Refresh vulnerability data:
+    --refresh-vulnerability-data    Refresh all projects (optionally limited by --project and --variant)
+
+    Data listing:
+    --list-projects                 List all projects and their variants
+    --list-scans                    List all past scans
```
sfl-pe-sync: WARNING: The rewritten host help removes `--json`, although `parse_and_run` still accepts it and forwards it for the listing commands, while the container entrypoint help continues to advertise it. Restore `--json` beside `--list-projects`/`--list-scans` so the two helpers remain consistent and the supported machine-readable interface remains discoverable.

sfl-pe-sync (changes requested):
The helper rewrite still misdocuments export scoping and omits a supported listing flag, so it does not yet provide a consistent CLI contract.

vulnscout
ValentinBoudevinSFL: This command doesn't need to be in the helper

vulnscout
ValentinBoudevinSFL: I know we have some inconsistencies. Will be a part of a future PR in the next release to define the scope

vulnscout
```diff
@@ -44,72 +44,95 @@ show_help() {
...
-  --list-scans                    List all past scans
-  --json                          Output objects in JSON format
-
-Configuration:
-  --config <key> <value>          Set a config value
-  --config-list                   Show current configuration
-  --config-clear <key>            Remove a config key
+    --import-custom-vulnscout-data <path> Import custom VulnScout JSON data
+
+    Import or export AI assessment context for a project:
```
sfl-pe-sync: WARNING: This groups `--import-context` under project-scoped commands, but that command does not consume `--project`: `cmd_import_context` forwards only the file, and the backend imports according to the project/variant names embedded in that file. A user invoking `--project P --import-context file.json` can therefore update a different embedded project (or have entries ignored) despite the new help promising project P. Document import as unscoped and driven by embedded identities; keep only `--export-context` in the project/variant-scoped section.

sfl-pe-sync (changes requested):
The revised help still misstates the scoping contract for context imports; previously reported blocking defects also remain unresolved.

vulnscout
ValentinBoudevinSFL: Fixed

sfl-pe-sync (approved):
No actionable defects found.

reh-sfl (approved):

PR #537, opened 2026-08-24
Support project-wide match conditions

reh-sfl:

Change the match-conditions arguments so that:
* No project or variant: evaluate default/default.
* --project A: evaluate all variants in project A.
* --project A --variant V: evaluate only A/V.
* --variant V: evaluate V in the default project.
* Unknown scope: exit 1.
* Shared vulnerability IDs: report once.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Test the above configurations.

## Pull Request Checklist

### DIFF
--- a/src/bin/cmd_process.py
+++ b/src/bin/cmd_process.py
@@ -28,7 +28,7 @@ import os
 from typing import Callable, Iterable, TYPE_CHECKING
 from flask.cli import with_appcontext
 from sqlalchemy import and_, exists
-from ._common import DEFAULT_VARIANT_NAME, resolve_project_variant
+from ._common import DEFAULT_VARIANT_NAME, resolve_project, resolve_project_variant
 from ..controllers.projects import ProjectController
 from ..helpers.export_scope import compute_export_scope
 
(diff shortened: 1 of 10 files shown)


src/bin/cmd_process.py
```diff
@@ -493,7 +517,14 @@ def _run_main(refresh_vulnerability_data: bool = False) -> ControllersCache:
     failed_vulns = []
     if match_condition:
         verbose("merger_ci: Start evaluating conditions")
-        failed_vulns = evaluate_condition(controllers.vulnerabilities, controllers.assessments, match_condition)
+        condition_controllers = ControllersCache(
```
sfl-pe-sync: ERROR: Building one union `ExportScope` for the whole project loses the association between each vulnerability/package and its variant-specific assessments. `AssessmentsController` filters only by `variant_id`, so `evaluate_condition` can select an assessment from another (or historical) variant for a vulnerability that is active in a different variant; for example, a newer `not_affected` assessment in variant B can suppress an `affected` match in variant A. Evaluate each variant with its own scope and union/deduplicate the matched vulnerability IDs (or make condition evaluation preserve package/variant identity) so project-wide matching means “matches in any variant.”

sfl-pe-sync (changes requested):
Project-wide scope resolution is wired through the wrappers, but the union scope conflates variant-specific vulnerability and assessment state, so conditions can produce incorrect pass/fail results across variants.

sfl-pe-sync (approved):
No actionable defects found.

ValentinBoudevinSFL (approved):
LGTM

PR #538, opened 2026-08-28
Assessment with multi targets

sfl-zwang:

* An assessment is now one row that says *where* it applies, instead of one row per package/variant. A new `assessment_targets` table holds the `(assessment, variant, finding)` triples, and the old `assessments.variant_id` / `assessments.finding_id` columns are gone.
* The old grouping mechanism is removed: `assessment_group_member` and its model are deleted, since an assessment is now its own group and its members are its targets.
* Migration `x0a1b2c3d4e5` was rewritten. On upgrade, it creates the new table, gives every existing assessment one target from its old columns, fuses rows that one user action had duplicated (same project, same CVE, same content and timestamp) into a single row with several targets, then drops the old columns. Downgrade fans every target back out into its own row.
* Writing an assessment over several packages and/or variants now creates one row with many targets instead of many near-identical rows. This applies to the single-write endpoint, the batch endpoint, and the group reconcile endpoint.
* Reads follow the same path: listings, filtering by variant or project, scan queries, staleness ("outdated") checks, and cleanup of removed packages all reach variants and findings through the target rows.
* Import and export were inverted to match. Custom-data export is now version 2 with an explicit `targets` list per assessment, OpenVEX writes one statement per target, and an OpenVEX export only lists products for the variants being exported.
* Reports expand an assessment into one entry per target, and drop targets belonging to variants outside the export scope.
* Frontend: the vulnerability modal sends one request for a multi-variant write rather than one per variant, reads `variant_ids` back from the API, and tells groups apart by how many targets they cover.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

1. Since a migration is modified, it is recommended to start a new instance with a fresh database: start the app on it and confirm the upgrade runs.
2. In the UI, open a vulnerability and write one assessment covering N packages and M variants. It should still appear as one entry in the history with all four targets listed, not N * M entries. Editing it and removing one package should keep the entry and drop only that target; removing every target should delete it.
3. Filter the review table by variant and by project, and confirm an assessment shows up under every variant it targets.
4. Export custom data and OpenVEX from a project, then re-import the custom data into a clean database. The assessments should come back with the same targets, and re-importing the same file twice should not create duplicates. An OpenVEX export should list only the variants that were exported.
5. Generate a report (PDF/HTML) for a project and confirm each assessment shows one entry per target, and that no variant outside the export appears.

In short, all other assessment-related features should work as before.

### Additional notes

### Related Issue

## Pull Request Checklist


src/migrations/versions/x0a1b2c3d4e5_add_assessment_targets.py
```diff
@@ -0,0 +1,243 @@
...
+Revision ID: x0a1b2c3d4e5
+Revises: w9f0a1b2c3d4
+Create Date: 2026-08-19 00:00:00.000000
+"""
+import json
+from alembic import op
+import sqlalchemy as sa
+
+
+revision = 'x0a1b2c3d4e5'
```
sfl-pe-sync, line 12: ERROR: This reuses a revision ID that already exists on the staging base branch. Any database that has applied the old `x0a1b2c3d4e5` migration will be considered current, so Alembic will never create `assessment_targets` or drop the scalar columns; the upgraded application will then fail as soon as it queries the missing table. Keep the published revision intact and add a new successor migration that transforms `assessment_group_members`/the old assessment columns into the new schema.

src/migrations/versions/x0a1b2c3d4e5_add_assessment_targets.py
```diff
@@ -0,0 +1,243 @@
...
+
+    ``OR IGNORE`` keeps the backfill idempotent so it can be re-run against a
+    table that already holds some of these triples -- a retried or partially
+    applied upgrade, or a caller that seeded rows through the ORM before
+    invoking the backfill directly.  On a clean upgrade the target table is
+    created by this same revision, so every inserted triple is distinct and the
+    clause is a no-op.
+    """
+    orphans = connection.execute(sa.text(
+        "SELECT COUNT(*) FROM assessments"
```
sfl-pe-sync, line 87: ERROR: The previous schema explicitly permits `variant_id`/`finding_id` to be NULL, and legacy/unscoped assessment creation paths have produced such rows, but this migration aborts the entire deployment and instructs operators to resolve or delete them. That makes a valid pre-upgrade database non-upgradable and risks assessment loss. Add a deterministic data-preserving backfill for nullable legacy rows (for example, derive applicable variants from observations/scans and preserve any genuinely unscoped remainder in a representable compatibility form) before enforcing the new target invariant.

src/controllers/assessments.py
```diff
@@ -165,7 +183,10 @@ def _matches_current_variant(self, assessment: Assessment) -> bool:
         """
         if self.current_variant_id is None:
             return True
-        return assessment.variant_id is None or assessment.variant_id == self.current_variant_id
+        return (
+            assessment.single_variant_id is None
```
sfl-pe-sync: ERROR: `single_variant_id is None` now also means a real assessment targets multiple variants, not only that it is unscoped. Consequently a multi-target assessment for variants A/B is treated as matching an ingestion for unrelated variant C; `gets_by_vuln_pkg`/`warm_packages` can then deduplicate against it and skip creating C's assessment. Match `current_variant_id` against `assessment.target_rows` (and only treat an actually empty target set as legacy/global), and reuse that predicate in `warm_packages`.

sfl-pe-sync (changes requested):
The multi-target model is not deployable yet: existing staging databases will skip the rewritten migration, valid nullable legacy rows cannot be upgraded, and ingestion can misclassify multi-variant assessments as applying to unrelated variants.

src/migrations/versions/x0a1b2c3d4e5_add_assessment_targets.py
ValentinBoudevinSFL: Irrelevant to me because we compare to a main DB, not a staging one. 
We are overriding a wrong change made in staging

src/migrations/versions/x0a1b2c3d4e5_add_assessment_targets.py
ValentinBoudevinSFL: Irrelevant to me because we compare to a main DB, not a staging one. 
We are overriding a wrong change made in staging

src/controllers/assessments.py
ValentinBoudevinSFL: Looks valid to me

ValentinBoudevinSFL (dismissed):
LGTM, but look at VE comments

src/routes/assessments.py
```diff
@@ -202,33 +199,64 @@ def _get_db_assessment_dicts(
...
+                    ranked.c.justification,
+                    ranked.c.impact_statement,
+                    ranked.c.responses,
+                    ranked.c.workaround,
+                    ranked.c.vulnerability_id,
+                    ranked.c.name,
+                    ranked.c.version,
+                    ranked.c.supplier,
+                )
+                .where(ranked.c.target_rank == 1)
```
sfl-pe-sync: ERROR: This ranking keeps only one representative target per assessment, and the serializer below emits neither the remaining packages nor `variant_ids`. Therefore `GET /api/assessments` misrepresents every multi-target assessment as applying to one arbitrary package/variant; unfiltered clients and status summaries can hide its coverage under the other variants. Preserve the one-assessment response shape but aggregate all matching targets into `packages` and `variant_ids` (or serialize eagerly loaded Assessment objects) instead of discarding them.

src/helpers/assessment_io.py
```diff
@@ -657,6 +756,23 @@ def _export_assessments(
...
+                # Every target is exported even when the export is scoped to a
+                # subset of variants. Unlike a published VEX document, this
+                # format is a self-describing backup: each target carries its
+                # own variant_id and is re-imported verbatim. Truncating the
+                # list to the scope would make a round-trip lossy, silently
+                # dropping targets from the assessments it restores.
+                # ``variant`` (the human-readable name) travels with the id
+                # because a variant_id generated by another VulnScout instance
+                # never matches a local one -- see ``_resolve_v2_target``.
+                "targets": [
```
sfl-pe-sync, line 828: ERROR: A variant-scoped custom-data export admits an assessment when any target matches, then exports every target, including variants outside `variant_ids`. Selecting variant A can therefore disclose and later restore targets from variant B, contradicting the endpoint's scoped-export contract and the update-export promise to remove unselected variants. Filter the exported target list to `variant_ids` (and derive `packages` from that same subset); if whole-assessment backups must remain lossless, they cannot simultaneously be presented as variant-scoped exports.

src/helpers/assessment_io.py
```diff
@@ -203,7 +242,7 @@ def detect_review_export_format(doc: object) -> str:
         return "openvex"
     if not isinstance(doc, dict):
         raise ValueError("Export file must contain a JSON object")
-    if doc.get("version") == 1 and isinstance(doc.get("assessments"), list):
+    if doc.get("version") in (1, 2) and isinstance(doc.get("assessments"), list):
```
sfl-pe-sync, line 286: ERROR: Enabling version 2 here still routes its records through the validation at line 252, which requires a string `variant_id`. The exporter intentionally writes `variant_id: null` for a cross-variant assessment and puts its scope in `targets`, so `export-update` rejects VulnScout's own valid v2 export as containing an invalid record. Validate v2 assessment records from their non-empty `targets` entries and allow the legacy top-level `variant_id` to be null.

src/helpers/assessment_io.py
```diff
@@ -894,6 +1072,97 @@ def _import_assessments(
...
+                    if error is not None:
+                        result["errors"].append({
+                            "vuln_id": vuln_name,
+                            "package": pkg_string,
+                            "error": error,
+                        })
+                        continue
+                    assert v_id is not None and f_id is not None
+
+                    existing = duplicate_multitarget_assessment_exists(
```
sfl-pe-sync, line 1170: ERROR: De-duplicating each target before checking the complete incoming set corrupts multi-target imports. If target A already exists as a single-target assessment and the file contains one assessment for A+B, A is removed here and a new B-only assessment is created, so the imported judgement no longer has the exported target set. Resolve the full set first and perform only the exact-set duplicate check; an overlapping subset/superset must not remove individual targets. The OpenVEX import loop has the same per-target pattern and should use the same fix.

src/helpers/assessment_io.py
```diff
@@ -874,6 +991,67 @@ def _resolve_variant(raw_item: dict) -> "_uuid.UUID | None":
...
+        pkg = db.session.execute(
+            db.select(Package).where(
+                Package.name == name,
+                Package.version == version,
+                Package.supplier == supplier,
+            )
+        ).scalar_one_or_none()
+        if pkg is None:
+            return None, None, f"Package '{pkg_string}' not found"
+        finding = db.session.execute(
```
sfl-pe-sync, line 1103: ERROR: Version-2 import accepts any globally existing Finding and pairs it with the resolved variant without verifying that the package/vulnerability was ever observed in that variant. A backup from a different instance, or a crafted file, can therefore create a target for variant B using a finding that exists only in variant A, violating the same invariant enforced by interactive writes. Resolve through an Observation/Scan constrained to `resolved_variant_id` (for example by reusing `find_valid_finding`) and reject unobserved pairs.

src/routes/assessments.py
```diff
@@ -73,7 +74,7 @@ def _is_scanner_author(author: str | None) -> bool:
 def _has_pending_ai(vuln_id: str, variant_id: UUID | None) -> bool:
     """True if a pending AI assessment already exists for this (vuln, variant)."""
     for a in DBAssessment.get_by_vulnerability(vuln_id):
-        if a.origin == "ai" and a.variant_id == variant_id:
+        if a.origin == "ai" and a.single_variant_id == variant_id:
```
sfl-pe-sync: ERROR: `single_variant_id` is `None` for a valid AI assessment targeting multiple variants, so an existing pending AI assessment for A+B does not block another pending suggestion for A or B. Check whether `variant_id` occurs in `a.target_rows` instead; only an actually empty target set should use any legacy unscoped behavior.

sfl-pe-sync (changes requested):
The multi-target model is not yet safe to merge: core listing serialization loses targets, scoped custom exports include out-of-scope targets, valid v2 exports cannot be updated, and import/deduplication can fabricate or alter target sets.

src/controllers/assessments.py
sfl-pe-sync: Confirmed fixed in the current patchset: `_matches_current_variant` now checks membership in `assessment.target_rows`, and both `gets_by_vuln_pkg` and `warm_packages` use it.

frontend/src/components/VulnModal.tsx
```diff
@@ -1132,7 +1132,7 @@ type VariantScopedSnapshot = {
 
     const latestAssessmentFor = (variantIdValue: string, pkg: string | null): Assessment | null =>
         allVulnAssessments
-            .filter(a => a.origin !== "ai" && a.variant_id === variantIdValue && (pkg === null || a.packages.includes(pkg)))
+            .filter(a => a.origin !== "ai" && appliesToVariant(a, variantIdValue) && (pkg === null || a.packages.includes(pkg)))
```
sfl-pe-sync: ERROR: This combines variant membership and package membership independently, so a sparse assessment targeting `(variant A, pkg1)` and `(variant B, pkg2)` is treated as also covering `(A, pkg2)` and `(B, pkg1)`. The per-variant status matrix can therefore display a verdict on package/variant pairs that were never assessed. Preserve the target pairs in the API/`Assessment` type and match the exact `(variant_id, package)` pair here (and in the `assessmentPkgs` calculation below) instead of crossing `variant_ids` with the union of `packages`.

src/routes/settings.py
```diff
@@ -582,7 +591,7 @@ def _compute_copy_assessment_operations(
             processed_target_finding_ids: set = set()
 
             for assessment in source_assessments:
-                source_finding = assessment.finding
+                source_finding = _finding_for_variant(assessment, source_uuid)
```
sfl-pe-sync: ERROR: `_finding_for_variant` returns only the first target for a variant, but this PR now creates one assessment with several package findings in the same variant. Because `source_assessments` contains that assessment only once, copy/preview processes one package and silently drops every other target. Iterate all target findings for `source_uuid` (or fan the assessment out per target) throughout the copy-operation and selection-validation paths.

src/models/finding.py
```diff
@@ -34,8 +35,13 @@ class Finding(Base):
     vulnerability: Mapped["Vulnerability"] = relationship(back_populates="findings")
     observations: Mapped[list["Observation"]] = relationship(
         back_populates="finding", cascade="all, delete-orphan")
-    assessments: Mapped[list["Assessment"]] = relationship(
-        back_populates="finding", cascade="all, delete-orphan")
+    assessment_targets: Mapped[list["AssessmentTarget"]] = relationship(
```
sfl-pe-sync, line 41: ERROR: Removing the delete cascade without adding cleanup to `Finding.delete()` leaves the public model deletion path—and parent cascades such as `Package.delete()`—trying to null `assessment_targets.finding_id`, which is part of a non-nullable primary key. With referenced findings this raises during flush (or can leave dangling rows where FK enforcement is disabled). Reap each target and delete assessments that lose their last target in a Finding mapper event or equivalent cascade-safe hook so every deletion path is covered, not only the two route helpers named in the comment.

src/controllers/assessment_groups.py
```diff
@@ -24,70 +21,52 @@
...
-        head = members[0]
+    for assessment in assessments:
+        stale = stale_by_assessment.get(str(assessment.id), set())
         targets = [
             {
-                "variant_id": str(m.variant_id) if m.variant_id else None,
-                "package": pkg,
-                "outdated": pkg in stale_by_assessment.get(str(m.id), set()),
-                "assessment_id": str(m.id),
+                "variant_id": str(row.variant_id),
```
sfl-pe-sync, line 64: WARNING: `stale` is the union of package strings stale in any targeted variant, so when the same package target is stale in variant A but current in variant B, both group targets are marked outdated. `annotate_assessments_outdated` now exposes `stale_targets` specifically to retain the variant dimension; key this flag by `(row.variant_id, package name)` rather than package string alone.

sfl-pe-sync (changes requested):
Blocking multi-target regressions remain in exact package/variant status matching, variant-copy completeness, and model-level finding deletion; group staleness also loses variant precision.

frontend/src/handlers/assessments.ts
```diff
@@ -200,7 +267,8 @@ const removeDuplicateAssessments = (assessments: Assessment[]): Assessment[] =>
             assessment.workaround || ''
         ].join('|');
 
-        const duplicateKey = `${assessment.vuln_id}::${packagesKey}::${assessment.status}::${descriptionsKey}::${assessment.variant_id ?? ''}`;
+        const variantsKey = assessmentVariantIds(assessment).slice().sort().join(',');
+        const duplicateKey = `${assessment.vuln_id}::${packagesKey}::${assessment.status}::${descriptionsKey}::${variantsKey}`;
```
sfl-pe-sync: ERROR: This key still flattens packages and variants independently, so two distinct sparse assessments with the same sets but different pairings—e.g. `(A, openssl)/(B, zlib)` versus `(A, zlib)/(B, openssl)`—collide and `Explorer` drops one via `removeDuplicateAssessments`. Build the key from sorted `targets` pairs when present (falling back to the legacy flat fields only when pairs are absent), and add a regression test for equal package/variant sets with different pairings.

sfl-pe-sync (changes requested):
The multi-target flow still loses semantically distinct sparse assessments in frontend deduplication, so the patchset is not ready to merge.

src/routes/_assessment_group.py
```diff
@@ -182,39 +183,69 @@ def parse_reconcile_payload(
...
+    covered: set[str] = set()
+    for variant_id in variant_ids:
+        findings, _absent = validate_assessment_findings(packages, vuln_id, variant_id)
+        for package in packages:
+            finding = findings.get(package.id)
+            if finding is not None:
+                resolved[(package.string_id, variant_id)] = finding
+                covered.add(package.string_id)
+
+    selected_packages = {package.string_id for package in packages}
```
sfl-pe-sync, line 241: ERROR: `resolve_target_set` only records whether each package resolved in at least one variant; it never verifies that every requested variant UUID exists. A request with `variant_ids=[valid_variant, random_uuid]` and packages observed in the valid variant therefore returns success and stores only the valid target, and an existing variant from another project is likewise silently ignored when none of these packages resolve there. The create and reconcile APIs thus accept invalid scope as a partial success. Resolve all requested `Variant` rows first, reject missing IDs, and validate their project before applying the intentional sparse package/variant intersection.

sfl-pe-sync (changes requested):
The multi-target paths still accept invalid variant scope and silently persist only the subset that resolves, so the API can report success for a partially applied assessment.

ValentinBoudevinSFL (dismissed):

sfl-zwang:
This PR will be superseded by:
[Assessment with multi targets (1/4)](https://github.com/savoirfairelinux/vulnscout/pull/543)
[Assessment with multi targets (2/4)](https://github.com/savoirfairelinux/vulnscout/pull/544)
[Assessment with multi targets (3/4)](https://github.com/savoirfairelinux/vulnscout/pull/545)
[Assessment with multi targets (4/4)](https://github.com/savoirfairelinux/vulnscout/pull/546)

sfl-pe-sync (changes requested):
The multi-target read path still duplicates one assessment per targeted finding in vulnerability detail and group-history responses, breaking the core one-assessment/one-history-entry contract.

---
- `src/routes/assessments.py:960` [error]: ERROR: This per-finding traversal now appends the same `Assessment` once for every finding it targets. For a multi-package assessment, `list_assess_by_vuln` returns duplicate records, and the identical loop in `list_assessment_groups` passes duplicates to `build_groups`, so the UI history renders one identical group per target instead of one group containing all targets. Query `DBAssessment.get_by_vulnerability(vuln_id)` once, or de-duplicate `rows` by assessment ID before serializing/building groups, while retaining the project-target predicate.

src/migrations/versions/x0a1b2c3d4e5_add_assessment_targets.py
```diff
@@ -0,0 +1,243 @@
...
+        LEFT JOIN variants v ON v.id = a.variant_id
+    """)).mappings().all()
+
+    buckets: dict[tuple, list] = {}
+    for row in rows:
+        key = (
+            row["project_id"],
+            row["vuln_id"], row["timestamp"], row["status"],
+            row["simplified_status"], row["status_notes"], row["justification"],
+            row["impact_statement"], row["workaround"], row["origin"],
```
sfl-pe-sync, line 139: ERROR: The fusion key omits `source`, but fusion now deletes every non-surviving assessment rather than merely grouping rows for display. Two otherwise-identical assessments from different scanners/import sources at the same timestamp are therefore collapsed and one provenance value is permanently lost; subsequent OpenVEX output also reports only the survivor's scanner. Include `source` in the key (and add a migration regression covering distinct sources) so only genuinely equivalent rows are fused.

src/routes/_scan_queries.py
```diff
@@ -262,12 +268,14 @@ def _assessment_rows_for_scans(scan_ids: List[uuid_module.UUID]) -> Sequence[Ass
         )
         .select_from(Observation)
         .join(Finding, Finding.id == Observation.finding_id)
-        .join(Assessment, Assessment.finding_id == Finding.id)
+        .join(AssessmentTarget, AssessmentTarget.finding_id == Finding.id)
```
sfl-pe-sync, line 271: ERROR: This join returns one row per target, but `_assessments_detail_for_scan` still de-duplicates the result by assessment ID and retains only the first row's package. An automated assessment imported from one multi-product statement therefore appears with an arbitrary single package in scan details and scan exports, and re-importing that export loses the other targets. Preserve/serialize every target row (or aggregate a target list per assessment and teach the scan export/import format to round-trip it) instead of collapsing these rows by assessment ID.

frontend/src/handlers/vulnerabilities.ts
```diff
@@ -134,13 +134,18 @@ const buildStatusSummary = (assessments: Assessment[], currentPackages?: string[
...
+    // Group assessments by variant. A cross-variant assessment covers every
+    // variant it targets, so it is counted once per target rather than being
+    // dropped into the fallback bucket. Assessments with no target at all are
     // grouped together under a single fallback key so they contribute one slot.
     const byVariant = new Map<string, Assessment[]>();
     assessments.forEach((assessment) => {
-        const key = assessment.variant_id ?? '__no_variant__';
-        if (!byVariant.has(key)) byVariant.set(key, []);
-        byVariant.get(key)!.push(assessment);
+        const variantIds = assessmentVariantIds(assessment);
```
sfl-pe-sync, line 152: ERROR: A multi-target assessment is added wholesale to every targeted variant after the earlier active-package filter has tested only the flattened package union. For example, an assessment covering `(variant A, oldpkg)` and `(variant B, currentpkg)` passes because `currentpkg` is active, then its verdict is counted for A even though A's only target is absent from the active SBOM. Filter each variant membership by that variant's exact target packages (using a per-variant active-package map/server-provided active target scope) before selecting the latest status.

sfl-pe-sync (changes requested):
The multi-target refactor still has blocking data-integrity issues: migration fusion can erase assessment provenance, scan detail/export paths collapse multi-package targets, and frontend summaries can apply a verdict to a variant whose targeted package is no longer active.

src/migrations/versions/x0a1b2c3d4e5_add_assessment_targets.py
```diff
@@ -0,0 +1,244 @@
...
+Revision ID: x0a1b2c3d4e5
+Revises: w9f0a1b2c3d4
+Create Date: 2026-08-19 00:00:00.000000
+"""
+import json
+from alembic import op
+import sqlalchemy as sa
+
+
+revision = 'x0a1b2c3d4e5'
```
sfl-pe-sync, line 12: ERROR: This replaces the migration already present on the `staging` base while retaining revision `x0a1b2c3d4e5`. Any database that has run the published assessment-group migration is already stamped with this revision, so Alembic will skip this upgrade and the application will query a missing `assessment_targets` table. Restore the published migration and add this transformation as a new successor revision that migrates `assessment_group_members` and the scalar columns.

src/migrations/versions/x0a1b2c3d4e5_add_assessment_targets.py
```diff
@@ -0,0 +1,244 @@
...
+
+    ``OR IGNORE`` keeps the backfill idempotent so it can be re-run against a
+    table that already holds some of these triples -- a retried or partially
+    applied upgrade, or a caller that seeded rows through the ORM before
+    invoking the backfill directly.  On a clean upgrade the target table is
+    created by this same revision, so every inserted triple is distinct and the
+    clause is a no-op.
+    """
+    orphans = connection.execute(sa.text(
+        "SELECT COUNT(*) FROM assessments"
```
sfl-pe-sync, line 87: ERROR: The prior schema allows `finding_id` and `variant_id` to be NULL, but this upgrade aborts when such a valid legacy row exists; the added regression test even codifies the deployment failure. Provide a deterministic, data-preserving representation/backfill for legacy unscoped assessments before dropping the columns instead of requiring operators to delete assessment history.

src/routes/_assessment_group.py
sfl-pe-sync, line 259: ERROR: `resolve_target_set` validates packages only against variants where a finding resolves, so an unknown UUID or a variant from another project is silently ignored when another requested variant supplies the package. The create/reconcile APIs then return success with a smaller scope than requested. Resolve every requested `Variant` first, reject missing IDs, and enforce that all belong to the target project before building the intentionally sparse target intersection.

src/helpers/assessment_io.py
```diff
@@ -894,6 +1132,97 @@ def _import_assessments(
...
+                    if error is not None:
+                        result["errors"].append({
+                            "vuln_id": vuln_name,
+                            "package": pkg_string,
+                            "error": error,
+                        })
+                        continue
+                    assert v_id is not None and f_id is not None
+
+                    existing = duplicate_multitarget_assessment_exists(
```
sfl-pe-sync: ERROR: Per-target duplicate removal happens before validating the incoming assessment as one complete target set. If target A already exists and an imported v2 assessment covers A+B, A is discarded and a new B-only assessment is created, so import no longer preserves the exported judgment or its grouping. Resolve all targets first and deduplicate only on the complete target set; apply the same rule to the OpenVEX path.

src/routes/vulnerabilities.py
```diff
@@ -763,6 +763,45 @@ def _h(v: int | None) -> Iso8601Duration | None:
...
+                    .select_from(SBOMDocument)
+                    .join(Scan, Scan.id == SBOMDocument.scan_id)
+                    .join(SBOMPackage, SBOMPackage.sbom_document_id == SBOMDocument.id)
+                    .join(Package, Package.id == SBOMPackage.package_id)
+                    .join(_PkgVariant2, (
+                        (_PkgVariant2.name == Package.name)
+                        & (_PkgVariant2.version == Package.version)
+                    ))
+                    .join(Finding, Finding.package_id == _PkgVariant2.id)
+                    .where(SBOMDocument.scan_id.in_(current_scan_ids))
```
sfl-pe-sync, line 788: ERROR: The new active-package map joins `_PkgVariant2` to each SBOM package by name and version but omits supplier, although package identity and assessment targets include supplier. If two suppliers provide the same name/version, a current SBOM package from supplier A marks supplier B's finding current and its assessment affects the variant status. Include normalized supplier equality in this join so `packages_current_by_variant` contains only exact package identities.

sfl-pe-sync (changes requested):
The multi-target model is not ready to merge: upgrade compatibility can leave production without the new table or block on valid legacy data, scoped writes can silently omit requested variants, imports can split target sets, and supplier-distinct packages can produce incorrect current statuses.

src/migrations/versions/x0a1b2c3d4e5_add_assessment_targets.py
ValentinBoudevinSFL: Does not need to be fix

ValentinBoudevinSFL: Does not need to be fix

src/migrations/versions/x0a1b2c3d4e5_add_assessment_targets.py
ValentinBoudevinSFL: Does not need to be fix

src/helpers/assessment_io.py
ValentinBoudevinSFL: Fixed

src/migrations/versions/x0a1b2c3d4e5_add_assessment_targets.py
```diff
@@ -0,0 +1,244 @@
...
+    )).scalar()
+    if orphans:
+        raise RuntimeError(
+            f"{orphans} assessment(s) have no valid target: finding_id or"
+            " variant_id is NULL.  Resolve or delete them before upgrading."
+        )
+    connection.execute(sa.text(
+        "INSERT OR IGNORE INTO assessment_targets (assessment_id, variant_id, finding_id)"
+        " SELECT id, variant_id, finding_id FROM assessments"
+    ))
```
sfl-pe-sync, line 98: ERROR: The migration uses SQLite-only `INSERT OR IGNORE` here (and `UPDATE OR IGNORE` in `fuse_duplicates`), but VulnScout explicitly supports PostgreSQL via `psycopg2-binary` and configurable `SQLALCHEMY_DATABASE_URI`. PostgreSQL will fail this upgrade with a syntax error before the schema is usable. Use SQLAlchemy/dialect-specific conflict handling (or portable select/insert/update logic) for both statements and add a PostgreSQL migration test.

src/routes/scans.py
```diff
@@ -878,34 +867,34 @@ def _persist_import_assessments(
...
-        if entry.package_key is not None:
-            finding = findings_by_pair.get((entry.package_key, entry.vulnerability_id))
-            targets = [finding] if finding is not None else []
+        if entry.package_keys is not None:
+            targets = [
+                finding
+                for package_key in entry.package_keys
+                if (finding := findings_by_pair.get(
+                    (package_key, entry.vulnerability_id))) is not None
+            ]
```
sfl-pe-sync, line 884: ERROR: The replacement duplicate check compares only the complete target set plus `status`/`origin` (and no timestamp is supplied here), while the removed `_assessment_identity` also distinguished `simplified_status`, `status_notes`, `justification`, and `impact_statement`. Importing a scan that has the same targets/status/origin but different assessment details will now be treated as a duplicate and the new details silently discarded. Include all persisted assessment content in the duplicate identity (and apply the same complete identity in the custom/OpenVEX import paths).

sfl-pe-sync (changes requested):
The multi-target refactor still has two blocking data-path defects: its rewritten migration is not portable to the supported PostgreSQL runtime, and scan-import deduplication can silently discard distinct assessment content.

src/migrations/versions/x0a1b2c3d4e5_add_assessment_targets.py
ValentinBoudevinSFL: Fixed

src/routes/scans.py
ValentinBoudevinSFL: Fixed

src/models/assessment_target.py
```diff
@@ -0,0 +1,160 @@
...
+        f"({variant_id}, {finding_id})"
+        for variant_id, finding_id in pairs
+        if variant_id not in projects or finding_id not in vulns
+    ]
+    if missing:
+        raise GroupInvariantError("Unknown target: " + ", ".join(sorted(missing)))
+
+    distinct_projects = {projects[variant_id] for variant_id, _ in pairs}
+    distinct_vulns = {(vulns[finding_id] or "").upper() for _, finding_id in pairs}
+
```
sfl-pe-sync, line 108: ERROR: `validate_targets` verifies only that all variants share a project and all findings share a CVE; it never verifies that each `(variant_id, finding_id)` pair is backed by an observation for that variant. The new v2 custom-data path resolves a globally existing `Finding` and then calls `Assessment.create`, so a file can attach a package/finding observed only in another project to the selected project's variant, making the assessment appear in the wrong project. Validate every pair through `Observation -> Scan.variant_id` (or reuse the historical-aware finding validation used by the write API) before accepting the target set, and add a cross-project import regression test.

src/migrations/versions/x0a1b2c3d4e5_add_assessment_targets.py
sfl-pe-sync: Confirmed: the migration now uses SQLAlchemy table operations instead of SQLite-only conflict syntax.

sfl-pe-sync (changes requested):
The multi-target refactor still permits invalid variant/finding pairings through imported data, which can surface another project's package assessment in the selected project. The two previously discussed portability and duplicate-identity fixes are present.

src/routes/scans.py
sfl-pe-sync: Confirmed: scan import duplicate detection now includes the persisted content fields and source in addition to the complete target set.

src/models/assessment_target.py
ValentinBoudevinSFL: Fixed

src/routes/assessments.py
```diff
@@ -1438,33 +1481,22 @@ def add_assessments_batch() -> ResponseReturnValue:
         results: list[AssessmentDict] = []
```
sfl-pe-sync, line 1499: ERROR: The batch endpoint still creates one assessment per input item using a single `variant_id`; it cannot create one assessment spanning several variants, and the added regression test explicitly expects two groups for two same-CVE variant items. This leaves `/api/assessments/batch` inconsistent with the new one-user-action/one-assessment model and with the PR's stated multi-variant batch behavior. Accept `variant_ids` on each batch item and resolve the full package/variant target set (or fuse compatible items from the same action) before calling `create_assessment_record`, then assert that a multi-variant batch write returns one row with all targets.

sfl-pe-sync (changes requested):
The multi-target refactor still leaves the batch API producing separate assessment rows across variants, so the core one-assessment-per-action behavior is incomplete and the PR is not ready to merge.

src/models/assessment_target.py
sfl-pe-sync: Confirmed: the v2 custom-data import path now resolves each finding through `Finding.get_observed_by_variant`, so the reported cross-project attachment path is fixed.

src/routes/assessments.py
ValentinBoudevinSFL: Fixed

src/migrations/versions/x0a1b2c3d4e5_add_assessment_targets.py
```diff
@@ -0,0 +1,267 @@
...
+Revision ID: x0a1b2c3d4e5
+Revises: w9f0a1b2c3d4
+Create Date: 2026-08-19 00:00:00.000000
+"""
+import json
+from alembic import op
+import sqlalchemy as sa
+
+
+revision = 'x0a1b2c3d4e5'
```
sfl-pe-sync, line 12: ERROR: This replaces the already-published `x0a1b2c3d4e5` revision that exists on `staging` (`assessment_group_members`) with different schema operations. Databases that already recorded this revision will skip the new upgrade entirely, leaving `assessment_group_members` and the old assessment columns in place while the application expects `assessment_targets` and no scalar columns, causing production upgrades to fail at runtime. Restore the published migration unchanged and add a new successor revision that migrates from `assessment_group_members`/the scalar columns to `assessment_targets`.

sfl-pe-sync (changes requested):
The multi-target implementation still has a blocking production upgrade defect because it rewrites an Alembic revision already present on staging; existing databases will not receive the new schema.

src/routes/assessments.py
sfl-pe-sync: Confirmed fixed in the latest patchset: each batch item now accepts `variant_ids`, resolves the complete target set, and creates one assessment row.

src/migrations/versions/x0a1b2c3d4e5_add_assessment_targets.py
ValentinBoudevinSFL: Not relevant as the main branch of the project doesn´t contain x0a1b2c3d4e5

frontend/src/pages/Review.tsx
```diff
@@ -114,10 +114,12 @@ function toReviewRow(
 // How long the copy button shows its "copied" confirmation before reverting.
 const COPIED_FEEDBACK_MS = 2000;
 
-/** The clipboard payload for a row: its group id when grouped, otherwise the
- *  id of its single assessment. Mirrors VulnModal's copy buttons. */
+/** The clipboard payload for a row: prefixed by whether the row spans more
+ *  than one (variant, package) target, since that's the user-facing
+ *  distinction between "a group" and "a single assessment". Mirrors
+ *  VulnModal's copy buttons. */
```
sfl-pe-sync, line 120: ERROR: Review rows can now represent one assessment ID with multiple targets, but `handleSaveEdit` still iterates `editingRow.targets` and issues per-target PUT/DELETE requests using each target's `assessment_id`. All targets of a multi-target row carry the same ID, so removing one package calls `DELETE /api/assessments/<same-id>` and deletes the entire assessment (including targets the user kept); unchanged targets may also trigger repeated updates/deletes of that same row. Route Review edits through `Assessments.reconcileGroup(...)` once, as `VulnModal` already does, passing the complete desired package/variant set and existing assessment IDs.

sfl-pe-sync (changes requested):
The multi-target persistence model is not wired into the Review-page edit flow: editing or removing one target can delete the complete assessment and its retained targets. This blocks merge independently of the previously reported migration issue.

frontend/src/pages/Review.tsx
ValentinBoudevinSFL: Fixed

src/routes/_assessment_group.py
```diff
@@ -242,27 +278,30 @@ def resolve_targets(
...
-    # A row that already exists for a still-selected combo stays a target even
-    # if the finding lookup above missed it, otherwise the reconcile would read
-    # it as deselected and delete a row the user only meant to edit.
+    expected_project_id: UUID | None = None
+    if existing_by_key:
+        existing_variant = Variant.get_by_id(next(iter(existing_by_key))[1])
+        if existing_variant is not None:
+            expected_project_id = existing_variant.project_id
+    try:
+        resolved, unobserved_initial = resolve_target_set(
```
sfl-pe-sync, line 357: ERROR: Reconcile reconstructs coverage by crossing the flattened `packages` and `variant_ids` sets, so it cannot preserve a sparse assessment such as `(A, openssl) + (B, zlib)`. A content-only edit can silently add `(A, zlib)` and `(B, openssl)` whenever those findings were also observed, changing the scope of the verdict. Accept and resolve an explicit target-pair list through the reconcile API and both edit callers; retain the flat-set cross-product only as a legacy fallback.

frontend/src/components/MultiEditBar.tsx
```diff
@@ -125,51 +125,46 @@ function MultiEditBar ({vulnerabilities, selectedVulns, resetVulns, appendAssess
...
-                    triples.push({ vuln_id, variant_id: baseVariantId, packages: pkgs });
-                }
+                targets.push({ vuln_id, variant_ids: selectedVariantIds, packages: pkgs });
             }
         } else {
-            // All-variants context — one item per (vuln, variant, pkg)
+            // All-variants context — one item per vulnerability, covering all
+            // variants where its selected packages were historically observed.
             await Promise.all(selectedVulns.map(async (vuln_id) => {
                 const variants = await Variants.listByVuln(vuln_id).catch(() => []);
```
sfl-pe-sync, line 148: ERROR: In a project-scoped all-variants view, this unscoped lookup returns variants for the CVE from every project. The new one-item payload then combines those IDs, and `resolve_target_set` rejects the mixed-project target set, causing the entire bulk assessment action to fail whenever the CVE also exists in another project. Pass `projectId` into `MultiEditBar` and filter the returned variants to that project (or add a project-scoped backend query) before constructing `variant_ids`.

sfl-pe-sync (changes requested):
The multi-target edit path still loses exact sparse target semantics, and project-scoped bulk assessment can fail when a vulnerability exists in another project. These core regressions block approval.

frontend/src/pages/Review.tsx
sfl-pe-sync: Confirmed: the per-target PUT/DELETE loop was replaced with one reconcile request, so the originally reported whole-assessment deletion bug is addressed.

src/routes/_assessment_group.py
ValentinBoudevinSFL: Fixed

frontend/src/components/MultiEditBar.tsx
ValentinBoudevinSFL: Fixed

src/routes/_assessment_group.py
```diff
@@ -125,20 +114,43 @@ def parse_reconcile_payload(
     if not isinstance(vuln_id, str) or not vuln_id:
         return None, {"error": "vuln_id is required"}
 
-    packages = data.get("packages")
-    if (not isinstance(packages, list) or not packages
-            or not all(isinstance(p, str) and p for p in packages)):
-        return None, {"error": "packages must be a non-empty list of package ids"}
+    target_pairs: "list[tuple[str, UUID]] | None" = None
```
sfl-pe-sync, line 117: ERROR: Rejecting an empty explicit `targets` list makes the deletion branch in `apply_reconcile` unreachable. When an editor removes every target, `reconcileTargetPairs` produces `[]`, this parser returns 400, and the assessment remains instead of being deleted as the PR contract specifies. Treat an explicitly supplied empty list as a valid authoritative target set (while still rejecting a missing/invalid legacy package/variant payload), and ensure both edit clients pass empty selections through rather than falling back to the existing sets.

src/controllers/assessments.py
```diff
@@ -157,15 +188,24 @@ def _matches_current_variant(self, assessment: Assessment) -> bool:
...
+        targeting variants A and B match an ingestion for unrelated variant C.
         """
         if self.current_variant_id is None:
             return True
-        return assessment.variant_id is None or assessment.variant_id == self.current_variant_id
+        target_variant_ids = {t.variant_id for t in assessment.target_rows}
+        if not target_variant_ids:
+            # No targets at all: legacy / API-created record, applies anywhere.
+            return True
+        return self.current_variant_id in target_variant_ids
```
sfl-pe-sync, line 209: ERROR: This variant-only predicate is insufficient for `gets_by_vuln_pkg` and `warm_packages` now that coverage can be sparse. For an assessment targeting `(variant A, openssl)` and `(variant B, zlib)`, a lookup for `(variant A, zlib)` passes because the assessment contains variant A and its flattened `packages` contains zlib; ingestion deduplication can then suppress the real A/zlib assessment. Match the exact target pair (current variant plus the requested package/finding) in package-specific lookups instead of testing variant membership independently.

sfl-pe-sync (changes requested):
The two reported scope fixes are present, but empty-target reconciliation still cannot delete an assessment and ingestion deduplication still conflates sparse target pairs, so the change is not ready to merge.

src/routes/_assessment_group.py
sfl-pe-sync: Confirmed: the current patch sends explicit target pairs and the backend treats them as authoritative, preserving sparse coverage.

frontend/src/components/MultiEditBar.tsx
sfl-pe-sync: Confirmed: `projectId` is now passed into `MultiEditBar` and both all-variants lookups filter by it.

src/routes/_assessment_group.py
ValentinBoudevinSFL: Fixed

src/controllers/assessments.py
ValentinBoudevinSFL: Fixed

sfl-pe-sync (approved):
No actionable defects found.

src/routes/_assessment_group.py
sfl-pe-sync: Confirmed: explicit empty `targets` now passes validation, reaches reconciliation, and deletes the targetless assessment.

src/controllers/assessments.py
sfl-pe-sync: Confirmed: package-specific lookups now match the exact `(variant, finding)` target pair, including warmed-cache paths.

ValentinBoudevinSFL (approved):

PR #542, opened 2026-09-03
Add granular CLI report export scopes

reh-sfl:

* Add per-project and per-variant report export selection to the CLI wrapper and entrypoint.
* Reuse report scope semantics already exposed by the web export flow.
* Reject invalid scopes without broadening the export and preserve unscoped behavior.
* Document the new CLI options and add wrapper and end-to-end coverage.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

* Run `python -m pytest tests/end_to_end_tests/test_merger_ci.py -k 'report_command' -q`.
* Run `bash tests/CI_tests/vulnscout_wrapper_test.sh`.
* Export once without scope, once with `--project <project>`, and once with `--project <project> --variant <variant>`; inspect that each report contains the expected records.

### Additional notes

Project-only scope includes all variants in the selected project. Variant scope requires its project, and invalid selections fail rather than silently exporting broader data.

### Related Issue

No linked issue.

## Pull Request Checklist


src/bin/cmd_export.py
```diff
@@ -127,14 +127,42 @@ def export_command(
...
+            if variant_obj is None:
+                raise click.ClickException(
+                    f"variant '{variant}' not found in project '{project}'"
+                )
+            scope = compute_export_scope(variant_id=variant_obj.id)
+        else:
+            scope = compute_export_scope(project_id=project_obj.id)
+
+    controllers = ControllersCache(scope=scope)
     templ = Templates(controllers)
```
sfl-pe-sync, line 150: ERROR: Scoped reports construct a scoped `PackagesController` but never call `_preload_cache()`. For a non-`None` scope, `PackagesController.to_dict()` and `__iter__()` deliberately refuse to fall back to the global DB and therefore return no packages until that cache is populated, so package-oriented sections of every project/variant report are silently empty even when the selected active SBOM contains packages. Preload the package cache here (as `export_command` and the web document export path already do) before constructing/rendering `Templates`, and add a scoped report assertion for an expected package, not only a CVE.

sfl-pe-sync (changes requested):
The new scope resolution rejects invalid selections correctly, but scoped CLI reports can silently omit all package data because their scoped package controller is never preloaded.

ValentinBoudevinSFL:
I believe this PR should be changed. 

The current scope is project/variant to generate a report. 

But we can't generate a report for all variants in the project. 

Adding the possibility to export all projects in a single report causes a lot of questions: How to reflect it on the web interface? How to add specific project variables in the template for reports?.... 

I'm not saying it is something we will never do but that would be the subject of another PR.

Right now, the scope of this PR is to add --project scope

ValentinBoudevinSFL:
Please squash the two commits

src/bin/cmd_export.py
```diff
@@ -127,14 +127,26 @@ def export_command(
               help="Directory where generated reports are written.")
 @click.option("--format", "output_format", default=None,
               help="Output format override: pdf or html (default: use template extension).")
+@click.option("--project", "-p", required=True,
+              help="Project name. The report includes all variants in this project.")
```
sfl-pe-sync, line 131: ERROR: The report command only accepts a project and always computes a project-wide scope, so `./vulnscout --project P --variant V --report ...` cannot produce the advertised variant report and can expose records from every other variant in P. Add a `--variant` option, resolve it with `DBVariant.get_by_name_and_project`, fail if it is unknown, compute the scope from its variant ID, and forward the selected variant from `cmd_report`; replace the test that rejects `--variant` with isolation coverage for two variants in one project.

sfl-pe-sync (changes requested):
The project-only path is implemented, but the advertised per-variant export scope is absent, causing variant-targeted reports to broaden to the whole project.

src/bin/cmd_export.py
ValentinBoudevinSFL: Looks valid feedback to me

src/bin/cmd_export.py
```diff
@@ -127,14 +127,37 @@ def export_command(
...
+        if variant_obj is None:
+            raise click.ClickException(
+                f"variant '{variant}' not found in project '{project}'"
+            )
+        scope = compute_export_scope(variant_id=variant_obj.id)
+    else:
+        scope = compute_export_scope(project_id=project_obj.id)
+
+    controllers = ControllersCache(scope=scope)
     templ = Templates(controllers)
```
sfl-pe-sync, line 162: ERROR: Passing `scope` into `ControllersCache` only scopes packages, vulnerabilities, and assessments; `Templates.render()` still populates `projects`, `variants`, `scans`, and `sbom_documents` through each controller's global `get_all()`. A custom project- or variant-scoped report can therefore expose sibling projects/variants, their scan metadata, and SBOM document paths, and the newly documented claim that `variants` is in-scope is false. Extend the scoped cache/template context so these collections are filtered from `scope.variant_ids` (and related project/scan/document IDs), and add a custom-template isolation test covering every exported context collection.

sfl-pe-sync (changes requested):
Scoped report rendering still exposes global project, variant, scan, and SBOM-document context, so the new isolation contract is not yet safe to merge.

src/bin/cmd_export.py
ValentinBoudevinSFL: Fixed

src/helpers/export_scope.py
```diff
@@ -40,16 +42,52 @@ class ExportScope:
...
+def _scope_for_variants(
+    selected_ids: set[uuid.UUID],
+    active_sbom_scan_ids: list[uuid.UUID],
+    project_ids: set[uuid.UUID] | None = None,
+) -> ExportScope:
+    """Build package and relational context IDs for *selected_ids*."""
+    variants = [variant for variant_id in selected_ids
+                if (variant := Variant.get_by_id(variant_id)) is not None]
+    all_scan_ids = {
+        scan.id
```
sfl-pe-sync, line 73: WARNING: Computing any scoped export now walks every historical scan for each selected variant and then issues one `SBOMDocument.get_by_scan()` query per scan. Because `compute_export_scope()` is also used by normal SBOM/VEX exports, web document generation, and scoped processing—not only report templates—this adds O(variants + historical scans) queries and materializes unrelated history on every scoped operation. Fetch the relational IDs with set-based queries (or compute this report-only context lazily in `Templates`) so non-template exports do not pay this cost.

sfl-pe-sync (approved):
The scope isolation is functionally addressed, but constructing the expanded scope introduces an avoidable N+1 query regression across all scoped export and processing paths.

ValentinBoudevinSFL (approved):

PR #543, opened 2026-09-11
Assessment with multi targets (1/4)

sfl-zwang:

Relationship to #538 and the split strategy

PR #538 ("Assessment with multi targets") replaces the current one-assessment-row-per-package/variant model with a single assessment row plus an assessment_targets table holding (assessment, variant, finding) triples, and retires the assessment_group_members mechanism. It's too large to review or merge safely as one unit.

This series rebuilds that same change as four independently deployable PRs — expand, migrate readers, migrate readers again, contract — that merge one after another into a shared integration branch, `assessment-multi-target-base`, before that branch lands on staging as a single unit. #538 itself stays open as the reference for what the finished series must reproduce, and closes once the base branch merges.

The reason for four PRs instead of one: assessments.variant_id/finding_id can't be dropped until every reader has moved off them, but the reader migration can't happen until the new table exists — and none of that can be sliced further while the old columns are gone, since a row can only name one target until they are. So the columns stay in place through PR-A, PR-B, and PR-C, dual-written on every write, while readers move onto target rows group by group. Because a reader reading through target rows behaves identically to one reading the old columns, each of those PRs changes no visible behavior and is safe to ship alone. PR-D removes the invariant, drops the columns, and turns on the real feature (one row covering many packages/variants).

This is PR-A, the first of the four.

What PR-A does

Adds the assessment_targets table alongside the existing assessments.variant_id/finding_id columns, and makes every assessment write populate both. Nothing reads target rows yet, so this PR changes no behavior — it's pure plumbing for the PRs that follow.

Specifically:
- Amends migration x0a1b2c3d4e5 to create assessment_targets (not superseding it — that revision only ever existed on staging, never released) and backfill exactly one target per existing assessment from the old columns. Adds a unique constraint on assessment_targets.assessment_id — enforced by the database, not convention — since it's what makes the next two PRs safe: without it, a stray double-write would turn every listing into duplicates while each reader still looked fine on its own.
- New AssessmentTarget model with the invariants that keep a target set sane (validate_targets, GroupInvariantError) and cleanup logic (reap_targets) that removes target rows — and any assessment left with none — when a finding or variant is deleted.
- Assessment.create() and add_target() now write a target row for every assessment, mirroring the same variant/finding pair into the old scalar columns so the ~31 places across seven files that still read those columns as SQL expressions keep working untouched.
- Audited every write path that bypasses those ORM helpers (the SCC bulk writer, cmd_process.py ingestion) so nothing can create a target-less assessment.

Verification: on a fresh database, every assessment has exactly one target row after flask db upgrade; the UI is unchanged when writing an assessment. Full backend/frontend suites pass at the existing coverage gates.

Note for reviewers pulling this locally: revision x0a1b2c3ss this whole series — delete your development database before testing, or Alembic will record the revision as already applied and silently skip the rewritten version in a later PR.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

### Additional notes

### Related Issue

## Pull Request Checklist


src/migrations/versions/x0a1b2c3d4e5_add_assessment_targets.py
```diff
@@ -0,0 +1,497 @@
...
+``assessment_group_members`` is the half that already shipped on ``staging``
+under this same revision id.  It stays here, unchanged in effect, because the
+model, the controller and the routes still read that table throughout PR-A,
+PR-B and PR-C.  **PR-D removes this half**, along with the group-member model
+it serves.
+
+``assessment_targets`` is the new half: the storage the reader migrations in
+PR-B and PR-C move onto.
+
+Both halves live in one revision on purpose.  This revision is amended in place
```
sfl-pe-sync, line 18: ERROR: This amends revision `x0a1b2c3d4e5` after that revision has already been applied on staging. Alembic will not execute this new `upgrade()` body on a database already stamped at that revision, so `assessment_targets` remains absent while this PR's model and read/write paths immediately query it; the first assessment operation will fail. Preserve the published revision and add a new successor revision that creates and backfills `assessment_targets`.

src/migrations/versions/x0a1b2c3d4e5_add_assessment_targets.py
```diff
@@ -0,0 +1,497 @@
...
+    The backfill is re-runnable: it inserts only for assessments that hold no
+    target yet.  Before inserting anything it proves that every target already
+    present agrees with the scalar columns of its assessment.  A disagreement
+    cannot be reconciled here -- ``uq_assessment_targets_assessment_id`` allows
+    only one target per assessment, so the correct row cannot be added
+    alongside the wrong one -- and swallowing it would leave the target table
+    and the scalar columns permanently contradicting each other while the
+    migration reported success.  It raises instead.
+    """
+    orphans = connection.execute(sa.text(
```
sfl-pe-sync, line 447: ERROR: The upgrade aborts on any assessment with a NULL variant/finding, but the pre-existing custom-data format accepted variant-less assessments and this same PR explicitly continues creating them. A database containing valid legacy imported data therefore cannot upgrade. Provide an explicit data-preserving migration for these rows (or a staged compatibility representation) instead of requiring operators to manually resolve or delete accepted records.

src/models/assessment.py
```diff
@@ -160,9 +171,12 @@ def vuln_id(self) -> str:
         val = self._vuln_id
         if val:
             return val
+        # ``validate_targets`` guarantees every target shares one
+        # vulnerability, so any one of them resolves it.
         try:
-            if self.finding:
-                return self.finding.vulnerability_id or ""
+            for target in self.target_rows:
```
sfl-pe-sync, line 177: ERROR: `assessment_io._import_assessments()` still creates target-less rows for entries without a variant, but this fallback now derives `vuln_id` and `packages` exclusively from `target_rows`; `get_by_vulnerability()` and `get_by_package()` were similarly changed to inner-join targets. Such successfully imported assessments therefore serialize with an empty vulnerability/package and disappear from vulnerability/package reads and exports. Keep the scalar `finding` fallback/read path until variant-less imports are retired, or reject that input consistently in this phase.

sfl-pe-sync (changes requested):
The additive target model is not safely deployable: already-stamped databases never receive the table, accepted legacy NULL-target data blocks fresh upgrades, and newly accepted variant-less imports become unreadable through the changed model queries.

ValentinBoudevinSFL (approved):
LGTM

PR #544, opened 2026-09-11
Assessment with multi targets (2/4)

sfl-zwang:
> Review this PR after https://github.com/savoirfairelinux/vulnscout/pull/543

Relationship to #538 and the split strategy

This is PR-B, the second of four PRs rebuilding #538 as independently deployable slices that merge one after another into the shared integration branch assessment-multi-target-base, which lands on staging as a single unit once all four have landed.

PR-A (merged/pending ahead of this one) added the assessment_targets table alongside the existing assessments.variant_id/finding_id columns and made every write populate both, so the old columns and the new table currently agree by construction. That's what makes this PR safe: PR-B moves a group of readers from querying the old columns to joining through assessment_targets instead, and under the one-target-per-assessment invariant PR-A enforces, a join through target rows returns exactly what a filter on the old columns returned. So this PR changes no visible behavior — it's a straight rewiring of internals, not a feature change.

What PR-B does

Moves the scan-related and cleanup read paths off Assessment.variant_id/finding_id and onto joins through assessment_targets:

- src/routes/_scan_queries.py, _scan_diff.py, _scan_helpers.py — the queries that reach a variant or finding from an assessment during scan diffing and result building now join AssessmentTarget instead of filtering on the scalar columns.
- src/routes/scans.py, src/helpers/outdated_cleanup.py — same move for scan-route lookups and outdated-assessment cleanup. Cleanup now also drops only the stale target of a partly-outdated assessment rather than the whole row, so an assessment with one variant gone but another still valid survives correctly instead of being deleted outright.
- Fixes cleanup to ignore target rows whose parent assessment has already been deleted (a residue case that can arise since SQLite's foreign_keys pragma is off in this app) — without it, cleanup counts could over-report or double-process rows.

Verification: filtering the review table by variant and by project, and running a scan diff between two scans, produce the same results as staging. Full backend/frontend suites pass at the existing coverage gates.

Same note as PR-A: revision x0a1b2c3d4e5 is still amended in place across this series — delete your development database before testing.
### Status

- [ ] READY
- [ ] HOLD
- [X] WIP (Work-In-Progress)

### How to verify this change

### Additional notes

### Related Issue

## Pull Request Checklist


src/helpers/outdated_cleanup.py
```diff
@@ -437,6 +457,11 @@ def delete_outdated_data(candidate_ids: dict[str, object] | None = None) -> dict
             for assessment in outdated_assessments
             if assessment["finding_id"] is not None
         }
+        # Core bulk DELETE bypasses the mapper events that reap target rows,
+        # so the targets go first and explicitly -- otherwise they survive
+        # pointing at a deleted assessment and keep their finding looking
+        # referenced to ``_delete_orphaned_findings`` below.
+        _delete_in_chunks(AssessmentTarget, AssessmentTarget.assessment_id, outdated_assessment_ids)
```
sfl-pe-sync, line 464: ERROR: This deletes every target row for an assessment as soon as any one target is classified outdated, and the following statement deletes the assessment itself. Once PR-D removes the one-target constraint, an assessment spanning a stale target and a still-current target will therefore lose the valid target and its shared assessment, contradicting this PR's stated partial-cleanup behavior and causing assessment data loss. Keep the stale `(assessment_id, variant_id, finding_id)` keys from `_outdated_assessments()`, delete only those target rows, and delete an `Assessment` only when no target rows remain; add a regression case with one stale and one current target after the uniqueness constraint is removed.

sfl-pe-sync (changes requested):
The target-join rewiring is generally consistent, but outdated cleanup still deletes whole assessments rather than only stale targets, which will cause data loss when multi-target assessments are enabled.

PR #547, opened 2026-09-14
Assessment split

ValentinBoudevin:

* An assessment is now one row that says *where* it applies, instead of one row per package/variant. A new `assessment_targets` table holds the `(assessment, variant, finding)` triples, and the old `assessments.variant_id` / `assessments.finding_id` columns are gone.
* The old grouping mechanism is removed: `assessment_group_member` and its model are deleted, since an assessment is now its own group and its members are its targets.
* Migration `x0a1b2c3d4e5` was rewritten. On upgrade, it creates the new table, gives every existing assessment one target from its old columns, fuses rows that one user action had duplicated (same project, same CVE, same content and timestamp) into a single row with several targets, then drops the old columns. Downgrade fans every target back out into its own row.
* Writing an assessment over several packages and/or variants now creates one row with many targets instead of many near-identical rows. This applies to the single-write endpoint, the batch endpoint, and the group reconcile endpoint.
* Reads follow the same path: listings, filtering by variant or project, scan queries, staleness ("outdated") checks, and cleanup of removed packages all reach variants and findings through the target rows.
* Import and export were inverted to match. Custom-data export is now version 2 with an explicit `targets` list per assessment, OpenVEX writes one statement per target, and an OpenVEX export only lists products for the variants being exported.
* Reports expand an assessment into one entry per target, and drop targets belonging to variants outside the export scope.
* Frontend: the vulnerability modal sends one request for a multi-variant write rather than one per variant, reads `variant_ids` back from the API, and tells groups apart by how many targets they cover.

### Status

- [x] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

1. Since a migration is modified, it is recommended to start a new instance with a fresh database: start the app on it and confirm the upgrade runs.
2. In the UI, open a vulnerability and write one assessment covering N packages and M variants. It should still appear as one entry in the history with all four targets listed, not N * M entries. Editing it and removing one package should keep the entry and drop only that target; removing every target should delete it.
3. Filter the review table by variant and by project, and confirm an assessment shows up under every variant it targets.
4. Export custom data and OpenVEX from a project, then re-import the custom data into a clean database. The assessments should come back with the same targets, and re-importing the same file twice should not create duplicates. An OpenVEX export should list only the variants that were exported.
5. Generate a report (PDF/HTML) for a project and confirm each assessment shows one entry per target, and that no variant outside the export appears.

In short, all other assessment-related features should work as before.

### Additional notes

### Related Issue

## Pull Request Checklist


src/migrations/versions/x0a1b2c3d4e5_add_assessment_targets.py
```diff
@@ -0,0 +1,243 @@
...
+Revision ID: x0a1b2c3d4e5
+Revises: w9f0a1b2c3d4
+Create Date: 2026-08-19 00:00:00.000000
+"""
+import json
+from alembic import op
+import sqlalchemy as sa
+
+
+revision = 'x0a1b2c3d4e5'
```
sfl-pe-sync, line 12: ERROR: This reuses the already-published `x0a1b2c3d4e5` revision that exists on `staging` as `add_assessment_group_members`. Any deployed database already stamped at that revision will consider this migration applied, so Alembic will not create `assessment_targets` or remove the scalar columns; the new ORM and queries will then fail against the old schema. Keep the published migration intact and add a new revision whose `down_revision` is `x0a1b2c3d4e5`, migrating existing group/scalar data to targets and dropping `assessment_group_members`.

sfl-pe-sync (changes requested):
The assessment-target refactor cannot be deployed over an existing staging installation because it rewrites an already-applied Alembic revision; a new follow-up migration is required.

PR #555, opened 2026-09-15
 Support exact variant/package target selection for assessments

ValentinBoudevin:

Add a new selection menu/array instead of the previous "Apply to variants" and "Apply to packages". 

Now it is possible valid sparse combinations such as (v1, p1) and (v2, p2) without implicitly selecting cross-pairs.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

-Go in Vulnerability tab and edit a vulnerability

image

-Go in Review tab and edit a vulnerability

image


frontend/src/components/StatusEditor.tsx
```diff
@@ -193,10 +212,11 @@ function StatusEditor ({onAddAssessment, progressBar, clearFields: shouldClearFi
             justification !== "none" ||
             statusNotes !== "" ||
             workaround !== "" ||
-            impact !== ""
+            impact !== "" ||
+            selectedTargets.length > 0
         );
         onFieldsChange?.(hasChanges);
```
sfl-pe-sync, line 232: WARNING: In exact mode, a single available variant/package pre-populates `initialTargets`, so this makes `onFieldsChange(true)` fire immediately even though the user has not edited anything (and it becomes true again after a successful reset). The modal can therefore show a discard-unsaved-changes prompt for a pristine or just-saved assessment. Compare canonical selected target keys with `initialTargets`—as `EditAssessment` does with its defaults—instead of treating every non-empty default selection as a change, and cover the single-target initialization/reset case.

sfl-pe-sync (changes requested):
The exact sparse target persistence path is sound, but the create editor now reports its automatically selected default target as an unsaved user change.

frontend/src/components/StatusEditor.tsx
ValentinBoudevin: Fixed

frontend/src/components/StatusEditor.tsx
```diff
@@ -67,6 +75,20 @@ function StatusEditor ({onAddAssessment, progressBar, clearFields: shouldClearFi
         variants?.length === 1 ? [variants[0].id] : []
     );
```
sfl-pe-sync, line 77: ERROR: `exactTargetMode` becomes false whenever `variantPackageMap` is still loading or its request fails, so an exact-selection flow silently renders the legacy independent variant/package controls. The Add button remains enabled during `findingsLoading`; on a slow or failed lookup a user can submit without `targets`, recreating the full package × variant cross-product this PR is meant to prevent. Keep exact mode keyed to `exactTargetSelection`, block submission until compatibility data is available (and surface lookup failure), and apply the same rule to `EditAssessment`, which has the same fallback condition.

sfl-pe-sync (changes requested):
The exact-target backend path is validated, and the prior pristine-state issue is fixed, but exact creation/editing can still fall back to legacy cross-product semantics while compatibility data is unavailable, so the change is not safe to merge yet.

frontend/src/components/StatusEditor.tsx
sfl-pe-sync: Confirmed: the canonical target-key comparison and reset coverage address the reported pristine-state issue.

frontend/src/components/StatusEditor.tsx
ValentinBoudevin: Fixed

sfl-pe-sync (dismissed):
The reported VulnModal compatibility fallback is fixed, but Review still converts per-variant package lookup failures into valid empty compatibility data, leaving exact targets incorrectly unavailable without an error.

frontend/src/components/StatusEditor.tsx
sfl-pe-sync: WARNING: The VulnModal path is fixed, but Review still catches each `Packages.list(variant.id)` failure as `[variant.id, []]` and passes that map without `findingsError`/`findingsLoading`. A failed lookup is therefore presented as a successfully loaded variant with every package unavailable, so users cannot add or remove valid exact targets for that variant. Track loading/error state in Review and pass it to `EditAssessment` instead of converting request failures to empty compatibility.

frontend/src/components/VulnModal.tsx
```diff
@@ -2090,9 +2097,14 @@ type VariantScopedSnapshot = {
                                                         )]}
                                                         availablePackages={projectPackages}
                                                         defaultSelectedPackages={groupPackages}
-                                                        variantPackageMap={Object.keys(variantPackageMap).length > 0 ? variantPackageMap : undefined}
+                                                        defaultSelectedTargets={targets.map(target => ({
```
sfl-pe-sync, line 2216: ERROR: Variantless legacy assessments are now passed into exact-target mode because `defaultSelectedTargets` is truthy even when every target has `variant_id: null`. `EditAssessment` drops those null targets when initializing `selectedTargetKeys`, so a status-only save emits empty `packages`/`targets`; `saveEditedAssessment` then takes the legacy branch with an empty desired key set and deletes the assessment instead of updating it. Keep variantless groups in the legacy editor (or preserve null-target selections end-to-end), and add a status-only edit regression test for a `{variant_id: null}` group.

sfl-pe-sync (changes requested):
The exact-pair flow works for variant-scoped assessments, but editing a legacy variantless assessment can delete it on an otherwise content-only save, so the change is not safe to merge yet.

frontend/src/components/VulnModal.tsx
ValentinBoudevin: Fixed

frontend/src/components/TargetPairSelector.tsx
```diff
@@ -0,0 +1,168 @@
...
+                            const rowSelected = allSelected(variantKeys);
+                            return (<tr key={variant.id} className="border-t border-gray-700">
+                                <th className="sticky left-0 bg-gray-800 px-3 py-2 text-left font-medium text-gray-200">
+                                    <div>{variant.name}</div>
+                                    <button
+                                        type="button"
+                                        disabled={variantKeys.length === 0}
+                                        onClick={() => setKeys(variantKeys, !rowSelected)}
+                                        aria-label={`${rowSelected ? 'Deselect' : 'Select'} all for variant ${variant.name}`}
+                                        className="mt-1 text-xs font-normal text-cyan-300 hover:text-cyan-200 disabled:cursor-not-allowed disabled:text-gray-600"
```
sfl-pe-sync, line 123: WARNING: This renders the full variants × packages matrix and repeatedly performs linear `includes` scans while building compatibility keys, headers, rows, and cells, making render work O(V × P × K) and creating V × P DOM cells. A normal SBOM with thousands of packages and several variants can therefore freeze the assessment modal, whereas the previous controls rendered only V + P inputs. Precompute each variant's packages as a `Set` and paginate/virtualize or otherwise bound the matrix before rendering it.

sfl-pe-sync (dismissed):
The exact-pair semantics are implemented correctly, but the unbounded compatibility matrix introduces a significant performance risk for production-sized SBOMs.

frontend/src/components/VulnModal.tsx
sfl-pe-sync: Confirmed: exact mode now requires at least one non-null variant target, and the new regression test verifies a status-only variantless edit uses PUT without deleting the assessment.

reh-sfl:
1. When I edit an assessment and deselect all targets, it deletes that assessment without warning me, the normal behavior should be similar to a standard deletion (aka a popup confirmation)

2. When editing a custom assessment from the Review page with a larger set of targets, an error bullet shows quiclkly and disappears (can't see what it is), make sure state for messageBanner is display = false unless an explicit exception is raised.

frontend/src/components/EditAssessment.tsx
```diff
@@ -72,6 +78,24 @@ function EditAssessment({
     const [selectedPackages, setSelectedPackages] = useState<string[]>(
         defaultSelectedPackages ?? (availablePackages?.length === 1 ? [availablePackages[0]] : [])
     );
+    const exactTargetMode = Boolean(
```
sfl-pe-sync, line 81: ERROR: `exactTargetMode` still depends on `availableVariants.length`. In Review, the editor opens with `editVariants = []` while `Variants.listByVuln` is pending, and it remains empty if that request fails. During that state this guard disables exact mode despite variant-scoped `defaultSelectedTargets`; Save bypasses the compatibility check and emits no `targets`, so `Review.persistEditedAssessment` falls back to `reconcileTargetPairs` and can expand a sparse assessment such as `(v1,p1),(v2,p2)` into the full cross-product. Derive exact mode from the existing non-null targets independently of asynchronously loaded variants, and block saving until variant compatibility loading succeeds; add a regression test that clicks Save before `listByVuln` resolves and when it rejects.

sfl-pe-sync (changes requested):
The exact-target editor still has a blocking async fallback: Review can save a variant-scoped sparse assessment through legacy cross-product reconciliation while its variant list is loading or failed.

frontend/src/components/EditAssessment.tsx
ValentinBoudevinSFL: Fixed

sfl-pe-sync: Confirmed: exact mode now derives from the persisted non-null targets, and saves are blocked during both variant loading and lookup failure. The added regressions cover both paths.

sfl-pe-sync (approved):
No actionable defects found.

reh-sfl (approved):

PR #558, opened 2026-09-16
Ensure Per-Target Pending Assessments and Add Native Package Scan Filtering

ValentinBoudevin:

* Create an initial Pending Assessment for every unassessed [(variant, finding)] target discovered by a scanner, including findings already known from another variant.
* Add an Ignore -native packages option to the scan wizard.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

- Go in the Scan tab and perform a scan. Now if performed on two variants or more, vulnerabilities are not placed in "unkown" status anymore

-In the Scan Wizard a new checkbox to ignore -native packages was added

image


sfl-pe-sync (approved):
No actionable defects found.

reh-sfl:
Implementation LGTM, but I would avoid skipping -native packages, because they can introduce vulnerabilities in the output binaries. For example a vulnerability in a compiler vulnerability (though not loaded into the board) can expose unsafe flags or a memory leak, and often silicon-specific flaws that could've been corrected in the IR compilation for that architecture / chip.

reh-sfl (approved):

PR #559, opened 2026-09-16
fix(packages): repair variant comparisons

ValentinBoudevin:

Restrict package metadata enrichment to the selected comparison pair. This prevents a third variant from leaking into package membership.

Initialize a valid compared variant when reopening a select-mode scope. The browser previously displayed the first option while React still held an empty value, leaving Apply disabled and preventing data from loading.

Cover the three-variant package case and restored selector state.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

I have three variants v1,v2,v3 with p2 specific to v2 and v3.

But comparing v1/v2 does not return p2 as it should. The package p2 should be visible as it is contain in v2 but not v1.

Wasn´t the case before this fix


frontend/src/components/ProjectVariantSelector.tsx
```diff
@@ -102,7 +102,9 @@ function ProjectVariantSelector({ defaultProject, defaultScope, onApply }: Reado
     }, [variants, compareBaseVariantId]);
 
     useEffect(() => {
-        if (selectedCompareVariantId === compareBaseVariantId) {
+        const compareIsValid = selectedCompareVariantId !== compareBaseVariantId
```
sfl-pe-sync, line 106: ERROR: On the initial render `variants` is still empty while `selectedCompareVariantId` may contain a valid persisted `defaultScope.compare_variant_id`; this check therefore marks that saved value invalid, clears it, and after the fetch selects the first non-base variant instead. Reopening a comparison whose saved compare variant is not the first option will silently compare against the wrong variant. Defer membership validation until the variants request has completed (or at least until the list is non-empty), while still initializing an actually empty saved value once variants are available, and add a regression test with a non-first saved compare variant.

sfl-pe-sync (changes requested):
The backend pairwise enrichment fix is sound, but the selector validity effect can overwrite a valid persisted comparison before variants finish loading and cause the wrong variant to be compared.

frontend/src/components/ProjectVariantSelector.tsx
ValentinBoudevin: Fixed

sfl-pe-sync: Confirmed fixed: validation now waits for variants to load, preserves a valid non-first persisted comparison, and covers both empty and persisted selections with regression tests.

sfl-pe-sync (approved):
No actionable defects found.

reh-sfl (approved):

PR #560, opened 2026-09-17
fix(scans): honor selected variant scope

ValentinBoudevin:

Pass multi-variant selections from Explorer into ScanHistory. Filter the scan wizard choices to the selected subset instead of exposing every variant in the project.

Add subset-aware scan history queries so the Scans tab excludes entries from unselected variants. Reload when the selection changes and ignore stale requests from the previous scope.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Set 3 variants v1,v2,v3. Then go in the Scan tab and change the selector to v1 and v2. 

With this fix, v3 doesn´t leak and appear anymore in the history


frontend/src/pages/ScanHistory.tsx
```diff
@@ -1084,12 +1085,12 @@ function ScanHistory({ variantId, projectId, onScanComplete }: Readonly<Props>)
     const sccRunning = sccEntries.some(e => e.status === "running" || e.status === "queued");
```
sfl-pe-sync, line 1092: ERROR: This refresh path is not covered by the cancellation guard in the later loading effect. If a scan-completion/description/delete refresh starts for `[v1,v2,v3]`, the user switches to `[v1,v2]`, and that older refresh resolves after the new scoped load, its unconditional `setScans` restores v3 history. Use a shared request generation or AbortController for every history load (including `refreshScans`) and ignore results whose scope is no longer current.

frontend/src/pages/ScanHistory.tsx
```diff
@@ -1262,7 +1270,7 @@ function ScanHistory({ variantId, projectId, onScanComplete }: Readonly<Props>)
             setAllVariants([]);
```
sfl-pe-sync, line 1329: ERROR: The variant-menu request has no cleanup or stale-response guard even though it now reruns when `variantIds` changes. A slower request for `[v1,v2,v3]` can resolve after the request for `[v1,v2]` and overwrite `allVariants`/`selectedVariantIds`, exposing v3 in the wizard and allowing an out-of-scope scan. Add an effect-local cancellation flag or AbortController, as done for scan history, and test out-of-order `Variants.list` resolutions.

sfl-pe-sync (changes requested):
The subset filtering is implemented end to end, but two asynchronous request paths can still restore data or wizard choices from the previous variant scope, so the core leakage fix is not reliable under out-of-order responses.

frontend/src/pages/ScanHistory.tsx
ValentinBoudevinSFL: Fixed

frontend/src/pages/ScanHistory.tsx
ValentinBoudevinSFL: Fixed

frontend/src/pages/ScanHistory.tsx
```diff
@@ -1247,23 +1279,36 @@ function ScanHistory({ variantId, projectId, onScanComplete }: Readonly<Props>)
 
     // Fetch variants scoped to the current view for the scan menu
     useEffect(() => {
+        let cancelled = false;
```
sfl-pe-sync, line 1302: ERROR: Changing `variantIds` leaves `allVariants` and `selectedVariantIds` from the previous subset usable until the new `Variants.list` request resolves. During a slow request from `[v1,v2,v3]` to `[v1,v2]`, the wizard can still be opened and launched with `v3`, so an out-of-scope scan remains possible even though stale responses are now ignored. Clear/constrain both states immediately when the scope changes (and disable launch while variants are loading), or derive the wizard and launch IDs by intersecting them with the current scope.

sfl-pe-sync (changes requested):
The stale-response fixes are sound, but the wizard retains the previous subset while the replacement variant request is pending, leaving a blocking path to launch scans for a deselected variant.

frontend/src/pages/ScanHistory.tsx
sfl-pe-sync: Confirmed: the shared request generation and abort controller now prevent completion refreshes from restoring stale history.

frontend/src/pages/ScanHistory.tsx
sfl-pe-sync: Confirmed: cleanup and cancellation now prevent an older variant response from overwriting the current response.

frontend/src/pages/ScanHistory.tsx
ValentinBoudevinSFL: Fixed

sfl-pe-sync (dismissed):
No actionable defects found.

frontend/src/pages/ScanHistory.tsx
sfl-pe-sync: Confirmed: the current scope now immediately makes the wizard unavailable, clears its prior choices, and intersects both displayed and launched variants with the explicit selection.

src/routes/scans.py
```diff
@@ -1072,11 +1072,40 @@ def import_scan() -> ResponseReturnValue:
...
+            project_id = flask_request.args.get('project_id')
+            if project_id:
+                try:
+                    project_uuid = uuid_module.UUID(project_id)
+                except ValueError:
+                    return jsonify({"error": "Invalid project_id"}), 400
+                if any(variant.project_id != project_uuid for variant in variants):
+                    return jsonify({"error": "Variant does not belong to project"}), 400
+            scans = ScanController.get_by_variants(variant_ids)
+            cache_ids = ','.join(sorted(str(variant_id) for variant_id in variant_ids))
```
sfl-pe-sync: ERROR: Every distinct caller-controlled variant subset is inserted into the module-level scan-list cache, which is only cleared on scan/assessment mutations. Because variant IDs are discoverable and there are exponentially many valid combinations, repeated requests can retain many complete serialized histories and exhaust a long-lived worker's memory. Do not cache arbitrary subset scopes, or replace this with a strictly bounded/evicting cache keyed by the normalized IDs.

sfl-pe-sync (changes requested):
The selected-variant behavior is correctly enforced, but the new subset route introduces an unbounded caller-controlled cache that can exhaust production worker memory and must be fixed before merge.

src/routes/scans.py
ValentinBoudevin: Fixed

sfl-pe-sync (approved):
No actionable defects found.

src/routes/scans.py
sfl-pe-sync: Confirmed: subset responses now bypass the process-wide cache, so caller-controlled scopes no longer accumulate retained serialized histories.

frontend/src/pages/ScanHistory.tsx
```diff
@@ -1213,7 +1244,28 @@ function ScanHistory({ variantId, projectId, onScanComplete }: Readonly<Props>)
     // Derive the effective variant IDs to scan: explicit prop or unique IDs from loaded scans
     const effectiveVariantIds: string[] = variantId
         ? [variantId]
+        : variantIds?.length
```
sfl-pe-sync, line 1247: ERROR: The new multi-variant scope is not applied to the bulk Export action: `handleExportAll` still sends only `project_id`, so a `[v1,v2]` view exports every scan in the project, including v3 (and can even fail the 200-scan limit because of hidden variants). Extend `/api/scans/export` to accept and validate the selected `variant_ids`, pass them here, and add a regression test proving an unselected variant is absent from the downloaded payload.

sfl-pe-sync (changes requested):
The history and scan wizard now honor the selected subset, but the Scans tab's bulk export still leaks scans from unselected variants, so the scoped workflow is not yet complete.

frontend/src/pages/ScanHistory.tsx
ValentinBoudevin: Export is made to be global

reh-sfl (approved):

sfl-zwang:
Found a small display bug. 

image


How to reproduce:
1. Go to the Scan page, select a variant, then click Apply.
2. Go to Settings and delete that variant.
3. Go back to the Scan page. The variant selector now shows (0) selected, but other variants are still intact.

sfl-zwang (approved):

PR #561, opened 2026-09-17
perf: optimize Explorer loading times

reh-sfl:

* ...
* ...

### Status

- [ ] READY
- [ ] HOLD
- [X] WIP (Work-In-Progress)

### How to verify this change

### Additional notes

### Related Issue

## Pull Request Checklist


src/routes/vulnerabilities.py
```diff
@@ -551,11 +572,19 @@ def index_vulns() -> ResponseReturnValue:
                         _affx_q = _affx_q.where(Finding.package_id.in_(_pkg_ids))
                     _pkg_var_rows = db.session.execute(_affx_q).all()
                     _pkgs_by_vuln_var: dict[str, list[str]] = {}
```
sfl-pe-sync, line 577: ERROR: Moving the vulnerability filter into Python makes this variant-scoped query effectively scan/materialize the global `Finding` set. The existing `OR` includes `_PkgVariant.id == Package.id`, which is true for each finding's original package even when it is unrelated to `latest_ids`; the membership test only discards those rows after SQLAlchemy has fetched them. On a multi-project database, request cost and memory now grow with all stored findings rather than the selected variant. Keep the scope in SQL without the large literal list, for example by joining a scoped vulnerability subquery/CTE or restructuring the query to start from scoped findings.

src/routes/vulnerabilities.py
```diff
@@ -740,7 +770,10 @@ def _h(v: int | None) -> Iso8601Duration | None:
             # packages_current: packages from the specific scan(s), expanded to include all
             # same-name+version supplier variants present in the active SBOM scans.
             # This ensures that Grype-linked packages (no supplier) also surface SBOM packages.
-            if current_scan_ids:
+            if current_scan_ids and packages_match_current_scans:
+                for v in vulns.values():
+                    v["packages_current"] = sorted(v["packages"])
```
sfl-pe-sync, line 772: ERROR: `r.packages` is not guaranteed to equal `packages_current` in the variant branch. The query that populates it deliberately retains each original finding package via `_PkgVariant.id == Package.id`, even when that package/version is absent from the active SBOM; a vulnerability with one current and one historical affected version therefore exposes both as current here. This corrupts Explorer package/status calculations that rely on `packages_current`. Retain the active-scan query for this branch, or build `r.packages` strictly from active scans before setting `packages_match_current_scans`.

src/routes/vulnerabilities.py
```diff
@@ -320,6 +322,22 @@ def _populate_found_by(
...
+    """Serialise a large list response without Flask's JSON provider.
+
+    Going through the provider would pretty-print the payload when the app runs
+    in debug mode and would route encoding through the pure-Python encoder;
+    ``json.dumps`` without a ``default`` hook keeps the C implementation.
+    """
+    payload = json.dumps(payload_obj, separators=(",", ":")).encode()
+    response = current_app.response_class(payload, mimetype="application/json")
+    if "gzip" in request.headers.get("Accept-Encoding", "") and len(payload) > 1024:
+        response.set_data(gzip.compress(payload, compresslevel=1))
```
sfl-pe-sync, line 335: WARNING: Substring matching does not implement `Accept-Encoding` negotiation: a client sending `Accept-Encoding: gzip;q=0` explicitly rejects gzip but will still receive a compressed body, while case-insensitive token variants are missed. Parse the header through Werkzeug (for example, require `request.accept_encodings["gzip"] > 0`) and add a regression test for the zero-quality case.

sfl-pe-sync (changes requested):
The optimization introduces two blocking regressions in the variant-scoped Explorer path: stale package versions are reported as current, and the replacement Python filter causes an effectively global findings query. Gzip negotiation also mishandles quality values.

src/routes/vulnerabilities.py
```diff
@@ -764,38 +794,45 @@ def _h(v: int | None) -> Iso8601Duration | None:
                 for v in vulns.values():
                     v["packages_current"] = sorted(pkgs_current_by_vuln.get(v["id"], []))
 
+            if current_scan_ids:
```
sfl-pe-sync, line 797: ERROR: This block still derives `packages_current_by_variant` exclusively from `SBOMDocument`, while the revised variant branch now treats packages observed by an active tool scan as current and copies them into `packages_current`. For the new tool-only regression fixture, `packages_current` contains `openssl@3.0.0` but `packages_current_by_variant` is empty; `buildStatusSummary` uses the latter to decide whether an exact assessment target is current, so valid tool-scan assessments are excluded from current status. Build the per-variant map from active `Observation`/`Scan` rows as well (with the same supplier-expansion rules), and assert the tool-only package appears under its variant ID.

src/routes/vulnerabilities.py
```diff
@@ -764,38 +794,45 @@ def _h(v: int | None) -> Iso8601Duration | None:
...
-                    .where(Finding.vulnerability_id.in_(vuln_ids))
                     .distinct()
                 ).all()
                 current_by_vuln_variant: dict[str, dict[str, list[str]]] = {}
                 for vuln_id, variant_id, pkg_name, pkg_version, pkg_supplier in scoped_pkg_rows:
+                    vuln_key = str(vuln_id)
+                    # Restricting to in-scope vulnerabilities here rather than in
+                    # SQL: a literal IN-list of every vulnerability id makes SQLite
+                    # probe the findings index once per id for each joined row.
+                    if vuln_key not in vulns:
```
sfl-pe-sync: WARNING: This Python membership check does not bound the SQL query to current vulnerabilities. Because `Package` rows are shared globally and the query joins every `Finding` for each package in the active SBOM, a scoped request materializes all historical and cross-project CVEs ever associated with those package IDs before discarding them here; cost can grow with database history rather than the selected scans. Join `Observation` constrained to `current_scan_ids`, or join a scoped vulnerability subquery/CTE, so SQLite filters these rows without restoring the large literal `IN` list.

sfl-pe-sync (changes requested):
The active package query is now scoped, but the optimized per-variant current-package path remains inconsistent for tool scans and still materializes globally shared historical findings.

src/routes/vulnerabilities.py
reh-sfl: Fixed

src/routes/vulnerabilities.py
reh-sfl: Fixed

src/routes/vulnerabilities.py
reh-sfl: Fixed

src/routes/vulnerabilities.py
reh-sfl: fixed

src/routes/vulnerabilities.py
reh-sfl: fixed

sfl-pe-sync (approved):
No actionable defects found.

src/routes/vulnerabilities.py
sfl-pe-sync: Confirmed: the package query now starts from observations in `latest_ids`, so it no longer materializes unrelated findings globally.

src/routes/vulnerabilities.py
sfl-pe-sync: Confirmed: `record.packages` is now populated only from findings observed in the active scans before it is reused as `packages_current`.

src/routes/vulnerabilities.py
sfl-pe-sync: Confirmed: Werkzeug's parsed encoding qualities now reject `gzip;q=0` and handle case-insensitive tokens; the added tests cover both.

src/routes/vulnerabilities.py
sfl-pe-sync: Confirmed: the per-variant current-package map now starts from active observations and includes the tool-only scan case covered by the regression test.

src/routes/vulnerabilities.py
sfl-pe-sync: Confirmed: the current-package query is now bounded in SQL by `Observation.scan_id.in_(current_scan_ids)`.

ValentinBoudevin (approved):

PR #563, opened 2026-09-18
fix(exports): make review data portable

ValentinBoudevin:

Export version 2 targets with variant names instead of database UUIDs. Keep version 1 imports and reject unsupported versions explicitly. Update reconciliation, HTTP, CLI, frontend validation, docs, and tests.

New schema:

```
{
  "version": 2,
  "exported_at": "2026-09-18T14:55:27.955960+00:00",
  "assessments": [
    {
      "vuln_id": "CVE-1999-0034",
      "status": "affected",
      "simplified_status": "",
      "justification": null,
      "impact_statement": null,
      "status_notes": "gh",
      "workaround": "gh",
      "timestamp": "2026-09-18T14:55:21.854000+00:00",
      "packages": [
        "perl@5.38.4"
      ],
      "targets": [
        {
          "variant": "variant_name",
          "package": "perl@5.38.4"
        }
      ]
    }
  ],
  "ai_assessments": [],
  "cvss": [],
  "time_estimates": []
}
```

### Status

- [X] READY
- [ ] HOLD
- [ ]  WIP (Work-In-Progress)

### How to verify this change

1- Generate a custom assessment and export/import it

2- Generate a custom assessment with upstream/main branch, checkout to this branch, and import the custom assessments. The previous schematic should work too


src/helpers/assessment_io.py
```diff
@@ -231,19 +231,41 @@ def _is_review_openvex_export(doc: object) -> bool:
     "time_estimates",
 )
 
+CUSTOM_DATA_VERSION = 2
```
sfl-pe-sync, line 234: ERROR: This changes the already-published version 2 schema without changing its version: current staging exports include `variant_id` in each v2 target, while the new validator/importer explicitly rejects that field. After upgrading, existing v2 backups can neither be imported nor amended, despite the portability change claiming backward compatibility. Either assign the portable schema a new version and retain a v2 reader, or accept and migrate both legacy and portable v2 target shapes.

sfl-pe-sync (changes requested):
The portable target representation is reasonable, but reusing version 2 while rejecting the prior version 2 representation breaks existing exports and blocks merge.

src/helpers/assessment_io.py
ValentinBoudevin: No the current version 2 is a WIP on staging branch. Current version available for users is upstream/main version v0.21 and doesn't contain it, it is version 1 only

src/helpers/assessment_io.py
```diff
@@ -1040,11 +1066,13 @@ def _variants_for_vuln_id(vuln_id: str) -> list[_uuid.UUID | None]:
...
 
-    def _resolve_variant(raw_item: dict) -> "_uuid.UUID | None":
+    def _resolve_variant(
+        raw_item: dict, *, portable_only: bool = False,
+    ) -> "_uuid.UUID | None":
         if variant_id is not None:
             return variant_id
 
-        variant_token = raw_item.get("variant_id")
+        variant_token = None if portable_only else raw_item.get("variant_id")
```
sfl-pe-sync: ERROR: `portable_only` stops reading `variant_id`, but the remaining resolver still interprets the `variant` value as a UUID before looking it up as a name. Because variant names are unrestricted strings, a v2 target whose exact variant name equals another local variant's UUID will resolve to that other variant, attaching imported assessments/CVSS/time estimates to the wrong scope. In portable mode, resolve the token strictly through the project-scoped exact-name map (without UUID parsing), and add a regression test with a UUID-shaped variant name colliding with another variant ID.

src/helpers/assessment_io.py
sfl-pe-sync: Agreed—if version 2 has only existed on the unreleased staging line and version 1 is the sole released format, the previously raised compatibility concern does not apply.

sfl-pe-sync (changes requested):
The portable v2 schema can still misresolve a valid UUID-shaped variant name as another variant's database ID, causing cross-variant import corruption.

src/helpers/assessment_io.py
ValentinBoudevin: Fixed

src/helpers/assessment_io.py
```diff
@@ -231,19 +231,41 @@ def _is_review_openvex_export(doc: object) -> bool:
...
+    """Return a supported VulnScout custom-data version.
+
+    Keeping version dispatch in one helper makes adding a future migration
+    path deliberate instead of silently interpreting a new schema as the
+    current one.
+    """
+    if not isinstance(doc, dict):
+        raise ValueError("Export file must contain a JSON object")
+    version = doc.get("version")
+    if type(version) is not int or version not in SUPPORTED_CUSTOM_DATA_VERSIONS:
```
sfl-pe-sync, line 252: ERROR: This rejects the previously documented legacy version value `"1.0"`. Before this change, both the HTTP and CLI import paths treated every present non-v2 version as the v1 layout, and the API documentation explicitly advertised `"version": "1.0"`; existing backups using that contract now fail despite the stated v1 compatibility. Accept and normalize the documented legacy string as version 1 (while continuing to reject unknown versions), and add HTTP/CLI regression coverage for it.

src/helpers/assessment_io.py
```diff
@@ -1039,11 +1065,23 @@ def _variants_for_vuln_id(vuln_id: str) -> list[_uuid.UUID | None]:
...
+        raw_item: dict, *, portable_only: bool = False,
+    ) -> "_uuid.UUID | None":
         if variant_id is not None:
             return variant_id
 
+        if portable_only:
+            variant_name = raw_item.get("variant")
+            if variant_name in (None, ""):
+                return None
+            mapped_variant = portable_variants_by_name.get(str(variant_name))
```
sfl-pe-sync, line 1087: ERROR: In portable mode an empty variant name is converted to `None`, the same value used for an intentionally unscoped CVSS or time-estimate record. The import endpoint calls `import_custom_data` without running `_is_valid_custom_record`, so a v2 payload containing `"variant": ""` can silently lose its scope and be imported as global (or applied across the vulnerability's variants) instead of reporting an unknown/invalid variant. Reserve missing/JSON null for unscoped records and explicitly reject empty or non-string v2 variant names in the importer before resolving them; cover both CVSS and time-estimate payloads.

sfl-pe-sync (changes requested):
The UUID-shaped-name collision is fixed, but the patch still breaks a documented legacy v1 import value and can silently turn blank v2 CVSS/time-estimate scopes into unscoped data.

src/helpers/assessment_io.py
sfl-pe-sync: Confirmed: portable v2 resolution now performs exact project-scoped name lookup without UUID parsing, and the added collision regression covers assessments, CVSS, and time estimates.

src/helpers/assessment_io.py
ValentinBoudevin: Fixed

src/helpers/assessment_io.py
ValentinBoudevin: Fixed

src/helpers/assessment_io.py
```diff
@@ -231,19 +231,43 @@ def _is_review_openvex_export(doc: object) -> bool:
...
+    Keeping version dispatch in one helper makes adding a future migration
+    path deliberate instead of silently interpreting a new schema as the
+    current one.
+    """
+    if not isinstance(doc, dict):
+        raise ValueError("Export file must contain a JSON object")
+    version = doc.get("version")
+    if version == "1.0":
+        return 1
+    if type(version) is not int or version not in SUPPORTED_CUSTOM_DATA_VERSIONS:
```
sfl-pe-sync, line 252: ERROR: The new strict version gate still rejects the legacy string form `"1"`, even though the pre-PR Review import regression test used that value and the backend previously treated every present non-v2 version as the v1 layout; changing that fixture to numeric `1` masks the compatibility break across HTTP, CLI, and the frontend. Normalize `"1"` to version 1 alongside `"1.0"` in both backend and frontend detection, and retain an end-to-end regression test for the previously accepted payload.

sfl-pe-sync (changes requested):
The portable v2 resolution fixes are sound, but legacy imports using the previously accepted string version `"1"` are now rejected, so the compatibility promise is not yet met.

src/helpers/assessment_io.py
sfl-pe-sync: Confirmed: `"1.0"` is now normalized to version 1, with both HTTP and CLI regression coverage.

src/helpers/assessment_io.py
sfl-pe-sync: Confirmed: direct v2 import now rejects empty and non-string CVSS/time-estimate variant values, and the regression test verifies that no scoped records are created.

src/helpers/assessment_io.py
ValentinBoudevin: Fixed

src/helpers/assessment_io.py
```diff
@@ -231,19 +231,43 @@ def _is_review_openvex_export(doc: object) -> bool:
...
+    Keeping version dispatch in one helper makes adding a future migration
+    path deliberate instead of silently interpreting a new schema as the
+    current one.
+    """
+    if not isinstance(doc, dict):
+        raise ValueError("Export file must contain a JSON object")
+    version = doc.get("version")
+    if isinstance(version, str) and version in {"1", "1.0"}:
+        return 1
+    if type(version) is not int or version not in SUPPORTED_CUSTOM_DATA_VERSIONS:
```
sfl-pe-sync, line 252: ERROR: Numeric `1.0` is still rejected because `type(version) is not int`, although the pre-PR detector accepted it (`1.0 == 1`) and the frontend continues to accept JSON `1.0` after parsing it as number `1`. Such a legacy file passes browser validation but receives a backend 400/CLI failure. Normalize non-boolean integral numeric supported versions (at minimum `1.0`) before dispatch and add HTTP/CLI coverage using raw JSON that preserves the decimal spelling.

sfl-pe-sync (changes requested):
The prior string-version fix is present, but legacy numeric `1.0` remains a blocking compatibility break and is inconsistently accepted by the frontend.

src/helpers/assessment_io.py
sfl-pe-sync: Confirmed: both `"1"` and `"1.0"` now normalize to version 1 in backend and frontend paths, with HTTP and CLI regression coverage.

src/helpers/assessment_io.py
ValentinBoudevin: Fixed

sfl-pe-sync (approved):
No actionable defects found.

src/helpers/assessment_io.py
sfl-pe-sync: Confirmed: integral numeric versions are normalized before dispatch, and the added HTTP and CLI regressions preserve the raw `1.0` spelling.

reh-sfl:
Please squash the legacy version verification commits:

image

PR #564, opened 2026-09-18
feat(operations): add the operation queue core and extract scan/refresh jobs

reh-sfl:
> [!NOTE]
> **Stacked PR 1 of 6 — split out of #529.** Review and merge in order; each PR targets the one before it.
>
> 1. **#548 — operation queue core, scan/refresh jobs extracted → base `staging` ← you are here**
>

---

* Add `OperationRegistry`, the single source of truth for operation state, replacing the five disconnected progress mechanisms that exist today (four `ProgressTracker` singletons, four per-variant `_*_scans_in_progress` closure dicts, `_upload_status`, and `_export_jobs`).
* Add `OperationQueue`, an in-process queue that owns scan and refresh sequencing server-side, so ordering no longer has to be reconstructed in the browser on every page load.
* Add an `EventBus` that fans registry changes out to subscribers, and `JobContext` / `ProgressReporter`, the reporting surface a job body writes progress through.
* Extract the four scan bodies out of `routes/scan_triggers.py` into `controllers/scan_jobs.py`, reporting through a `JobContext` instead of mutating progress dictionaries directly.
* Extract the four refresh bodies out of `routes/bulk_refresh.py` into `controllers/refresh_jobs.py`, and point `cmd_process.py` and `controllers/vulnerabilities.py` at the extracted versions.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

This layer is purely additive — nothing is wired to a route yet and every existing endpoint still behaves exactly as before. `make test` in `tests/` should pass, including the new `tests/unit_tests/test_operation_primitives.py` and `tests/unit_tests/test_refresh_jobs.py`.

### Additional notes

Splitting the queue core out on its own keeps the diff reviewable: this PR is all new code plus a mechanical move of job bodies, with no behaviour change and no endpoint touched.

### Related Issue

No linked issue.

## Pull Request Checklist

### DIFF
--- a/src/controllers/event_bus.py
+++ b/src/controllers/event_bus.py
@@ -1,0 +1,156 @@
+# Copyright (C) 2026 Savoir-faire Linux, Inc.
+# SPDX-License-Identifier: GPL-3.0-only
+
+"""In-process publish/subscribe bus backing the SSE event stream.
+
+Publishers (scan threads, refresh trackers, the operation watcher) call
+:meth:`EventBus.publish`.  Every connected SSE client owns one
+:class:`Subscription` and drains its queue.
+
+State is process-local: a multi-worker deployment would need a shared
+transport (Redis, database polling) before a single stream could serve
+events produced by another worker.
+"""
+
+from __future__ import annotations
+
+import queue
+import threading
+from collections import deque
+from typing import Any, Deque, Iterator, List, Optional, Set
+
+# Events kept for reconnecting clients that send ``Last-Event-ID``.
+DEFAULT_REPLAY_BUFFER = 512
+
+# Per-client backlog. A client slower than this is dropped and must resync
+# from a fresh snapshot rather than stalling the publisher.
+DEFAULT_CLIENT_BACKLOG = 256
+
+
+class Subscription:
+    """A single client's view of the bus."""
+
+    def __init__(self, bus: "EventBus", backlog: int) -> None:
+        self._bus = bus
+        self._queue: queue.Queue[Optional[dict]] = queue.Queue(maxsize=backlog)
+        self.overflowed = False
+        self.closed = False
+
+    def _offer(self, event: dict) -> None:
+        try:
+            self._queue.put_nowait(event)
+        except queue.Full:
+            self.overflowed = True
+
+    def close(self) -> None:
+        self._bus.unsubscribe(self)
+        try:
+            self._queue.put_nowait(None)
+        except queue.Full:
+            pass
+
+    def drain(self, timeout: float) -> List[dict]:
+        """Block up to *timeout* seconds for events, then take what is queued.
+
+        Returns an empty list when the timeout expires with nothing pending,
+        which is the caller's cue to emit a heartbeat.
+        """
+        events: List[dict] = []
+        try:
+            first = self._queue.get(timeout=timeout)
+        except queue.Empty:
+            return events
+        if first is None:
+            self.closed = True
+            return events
+        events.append(first)
+        while True:
+            try:
+                nxt = self._queue.get_nowait()
+            except queue.Empty:
+                break
+            if nxt is None:
+                self.closed = True
+                break
+            events.append(nxt)
+        return events
+
+    def __enter__(self) -> "Subscription":
+        return self
+
+    def __exit__(self, *_exc: Any) -> None:
+        self.close()
+
+
+class EventBus:
+    """Thread-safe fan-out with a bounded replay buffer."""
+
+    def __init__(
+        self,
+        replay_size: int = DEFAULT_REPLAY_BUFFER,
+        client_backlog: int = DEFAULT_CLIENT_BACKLOG,
+    ) -> None:
+        self._lock = threading.Lock()
+        self._subscribers: Set[Subscription] = set()
+        self._replay: Deque[dict] = deque(maxlen=replay_size)
+        self._client_backlog = client_backlog
+        self._seq = 0
+
+    @property
+    def seq(self) -> int:
+        with self._lock:
+            return self._seq
+
+    def subscriber_count(self) -> int:
+        with self._lock:
+            return len(self._subscribers)
+
+    def publish(self, event_type: str, data: dict) -> dict:
+        """Stamp *data* with a monotonic sequence id and fan it out."""
+        with self._lock:
+            self._seq += 1
+            event = {"seq": self._seq, "event": event_type, "data": data}
+            self._replay.append(event)
+            targets = list(self._subscribers)
+        for sub in targets:
+            sub._offer(event)
+        return event
+
+    def subscribe(self) -> Subscription:
+        sub = Subscription(self, self._client_backlog)
+        with self._lock:
+            self._subscribers.add(sub)
+        return sub
+
+    def unsubscribe(self, sub: Subscription) -> None:
+        with self._lock:
+            self._subscribers.discard(sub)
+
+    def close_all(self) -> None:
+        """Signal every stream to terminate so clients reconnect."""
+        with self._lock:
+            targets = list(self._subscribers)
+            self._subscribers.clear()
+        for sub in targets:
+            try:
+                sub._queue.put_nowait(None)
+            except queue.Full:
+                pass
+
+    def replay_since(self, last_seq: int) -> Optional[Iterator[dict]]:
+        """Return buffered events after *last_seq*.
+
+        ``None`` means the gap is too large to bridge and the client must
+        restart from a snapshot.
+        """
+        with self._lock:
+            if not self._replay:
+                return iter(()) if last_seq == self._seq else None
+            oldest = self._replay[0]["seq"]
+            if last_seq < oldest - 1 or last_seq > self._seq:
+                return None
+            return iter([e for e in self._replay if e["seq"] > last_seq])
+
+
+# Single bus shared by every stream in this process.
+operation_events = EventBus()
--- a/src/controllers/job_context.py
+++ b/src/controllers/job_context.py
@@ -1,0 +1,122 @@
+# Copyright (C) 2026 Savoir-faire Linux, Inc.
+# SPDX-License-Identifier: GPL-3.0-only
+
+"""Reporting handle passed to every queued job.
+
+A job body never touches the registry or the event bus directly; it calls
+``report`` / ``log`` / ``is_cancelled`` on its context.  The context
+coalesces updates so a scan emitting thousands of log lines produces a
+bounded number of SSE frames.
+"""
+
+from __future__ import annotations
+
+import threading
+import time
+from typing import Any, Callable, Dict, List, Optional
+
+from .operation_registry import registry
+
+# Minimum spacing between published updates for a single operation.
+FLUSH_INTERVAL_SECONDS = 0.25
+
+
+class CancelledError(Exception):
+    """Raised by :meth:`JobContext.check_cancelled` when a stop was requested."""
+
+
+class JobContext:
+    """Progress channel and cancellation signal for one operation."""
+
+    def __init__(self, op_id: str, options: Optional[Dict[str, Any]] = None) -> None:
+        self.op_id = op_id
+        self.options: Dict[str, Any] = options or {}
+        self._lock = threading.Lock()
+        self._cancel = threading.Event()
+        self._pending_logs: List[str] = []
+        self._pending_progress: Optional[Dict[str, Any]] = None
+        self._current = 0
+        self._total = 0
+        self._last_flush = 0.0
+        self._on_cancel: Optional[Callable[[], None]] = None
+
+    # ------------------------------------------------------------------
+    # Cancellation
+    # ------------------------------------------------------------------
+
+    def request_cancel(self) -> None:
+        with self._lock:
+            on_cancel = self._on_cancel
+        self._cancel.set()
+        if on_cancel is not None:
+            on_cancel()
+
+    def set_cancel_hook(self, hook: Optional[Callable[[], None]]) -> None:
+        """Register a callback that aborts blocking work (e.g. kills a subprocess)."""
+        with self._lock:
+            self._on_cancel = hook
+
+    def is_cancelled(self) -> bool:
+        return self._cancel.is_set()
+
+    def check_cancelled(self) -> None:
+        if self._cancel.is_set():
+            raise CancelledError()
+
+    # ------------------------------------------------------------------
+    # Progress reporting
+    # ------------------------------------------------------------------
+
+    def report(self, current: int, total: int, message: str) -> None:
+        with self._lock:
+            self._current = current
+            self._total = total
+            self._pending_progress = {
+                "current": current,
+                "total": total,
+                "message": message,
+            }
+        self._maybe_flush()
+
+    def message(self, message: str) -> None:
+        """Update the status line while keeping the current counters."""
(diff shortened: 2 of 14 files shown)


src/controllers/vulnerabilities.py
```diff
@@ -345,8 +346,7 @@ def remove(self, vuln_id: str) -> bool:
             return True
         return False
 
-    def fetch_epss_scores(self) -> EnrichmentResult:
-        from ..controllers.epss_progress import EPSSProgressTracker
+    def fetch_epss_scores(self, reporter: ProgressReporter = NULL_REPORTER) -> EnrichmentResult:
```
sfl-pe-sync: ERROR: Defaulting the new reporter argument to `NULL_REPORTER` removes the existing `EPSSProgressTracker`/`NVDProgressTracker` lifecycle for every unchanged no-argument caller. In particular, boot enrichment still calls `post_treatment(controllers)` while the shipped frontend polls `/api/epss/progress`, so that endpoint now remains idle throughout enrichment; `fetch_nvd_data()` has the same regression for its existing callers. Preserve a tracker-backed default/adapter until those legacy endpoints and clients are removed, or update all production call sites and endpoints atomically in this PR.

src/controllers/refresh_jobs.py
```diff
@@ -0,0 +1,444 @@
...
+    return 0.6 if os.getenv("NVD_API_KEY") else 6.0
+
+
+def _safe_commit(label: str) -> None:
+    """Commit, rolling back on failure, then expunge so the identity map stays bounded."""
+    try:
+        db.session.commit()
+    except Exception as exc:
+        print(f"[{label}] commit error: {exc}", flush=True)
+        db.session.rollback()
```
sfl-pe-sync, line 62: ERROR: `_safe_commit` rolls back a failed batch but suppresses the exception, so every refresh runner continues and the queue ultimately records `done` even though updates were discarded (for a final-commit failure, potentially all updates). Re-raise after rollback so `OperationQueue` marks the operation as errored; only successful commits should advance to the completion report.

src/controllers/job_context.py
```diff
@@ -0,0 +1,122 @@
...
+        self._on_cancel: Optional[Callable[[], None]] = None
+
+    # ------------------------------------------------------------------
+    # Cancellation
+    # ------------------------------------------------------------------
+
+    def request_cancel(self) -> None:
+        with self._lock:
+            on_cancel = self._on_cancel
+        self._cancel.set()
```
sfl-pe-sync: WARNING: Cancellation is not atomic with cancel-hook registration. If cancellation occurs after `_execute` starts the job but before `_run_cancellable` installs `process.terminate`, `request_cancel` observes no hook, sets the event, and the later hook registration does nothing; the subprocess then runs until completion or its 10-minute timeout despite an accepted cancellation. Synchronize setting the event and hook, and have `set_cancel_hook` immediately invoke a newly registered hook when cancellation was already requested (or check cancellation before spawning).

src/controllers/operation_registry.py
```diff
@@ -0,0 +1,201 @@
...
+        with self._lock:
+            operation = self._operations.get(op_id)
+            if operation is None or operation["status"] not in TERMINAL_STATUSES:
+                return False
+            del self._operations[op_id]
+            self._finished_at.pop(op_id, None)
+        operation_events.publish("operation_removed", {"op_id": op_id})
+        return True
+
+    def prune(self) -> None:
```
sfl-pe-sync, line 149: WARNING: The documented one-hour retention is never enforced because production code never calls `prune()`; only its unit test does. Consequently every completed operation and up to 500 log entries remain in the process for its lifetime, making the registry grow without bound. Invoke pruning from a regular production path such as `create`/`snapshot`, or schedule periodic pruning when the queue is initialized.

sfl-pe-sync (changes requested):
The queue foundation has blocking compatibility and correctness defects: legacy enrichment progress is silently disabled, and database commit failures are reported as successful operations. Cancellation and retention also need hardening before this core is used by the subsequent SSE layers.

src/controllers/vulnerabilities.py
reh-sfl: Fixed

src/controllers/refresh_jobs.py
reh-sfl: Fixed

src/controllers/job_context.py
reh-sfl: Fixed

src/controllers/operation_registry.py
reh-sfl: Fixed

src/controllers/scan_jobs.py
```diff
@@ -0,0 +1,755 @@
...
+    return uuid_module.UUID(str(ctx.options["variant_id"]))
+
+
+def _exclude_kernel(ctx: JobContext) -> bool:
+    return bool(ctx.options.get("exclude_kernel", True))
+
+
+def _active_packages(ctx: JobContext) -> Sequence[Package]:
+    packages, error = resolve_active_packages(
+        _variant_uuid(ctx), exclude_kernel=_exclude_kernel(ctx)
```
sfl-pe-sync: ERROR: The extracted runners drop the existing `exclude_native` behavior. `_active_packages()` forwards only `exclude_kernel`, and the Grype de-duplication/output filters likewise never check native package names, so a queued scan requested with `exclude_native=true` will scan and persist findings for `*-native` packages even though every legacy scan endpoint honors that option. Carry `exclude_native` through `JobContext.options`, pass it to `resolve_active_packages`, and apply it in both Grype filtering stages, with regression coverage for all scan runners.

src/controllers/refresh_jobs.py
```diff
@@ -0,0 +1,445 @@
...
+
+    package_ids = active_package_ids_for_scans(scan_ids)
+    query = (
+        db.select(Vulnerability.id)
+        .join(Finding, Vulnerability.id == Finding.vulnerability_id)
+        .join(Observation, Finding.id == Observation.finding_id)
+        .join(Scan, Observation.scan_id == Scan.id)
+        .where(Observation.scan_id.in_(scan_ids))
+    )
+    if package_ids:
```
sfl-pe-sync: ERROR: Skipping the package predicate when `package_ids` is empty makes an empty/no-SBOM variant inherit every finding from its active tool scans. For example, after importing an empty replacement SBOM, `active_scan_ids_for_variant()` still includes the latest historical tool scans, this branch omits `Finding.package_id.in_(package_ids)`, and a deferred refresh treats those stale vulnerabilities as current. Apply the active-package restriction unconditionally (or explicitly exclude tool-scan observations when the active package set is empty) so removed package versions cannot drive refresh jobs.

src/controllers/scan_jobs.py
```diff
@@ -0,0 +1,755 @@
...
+    from .scc_engine import get_engine
+
+    variant_uuid = _variant_uuid(ctx)
+    ctx.log("Loading local NVD advisory database…")
+    try:
+        engine = get_engine(progress=_engine_progress(ctx))
+    except Exception as error:
+        raise RuntimeError(f"Failed to load local NVD database: {error}")
+
+    scan = Scan.create(
```
sfl-pe-sync: ERROR: The DB-backed scan runners persist a new `Scan` before entering their cancellable loops. Cancelling NVD, OSV, or SCC after this point raises `CancelledError`, but the surrounding rollback cannot undo `Scan.create()`'s commit (and SCC/OSV may already have committed partial findings), leaving an empty or partial scan as the latest active scan and hiding the previous complete result. Ensure cancelled/failed runs delete their newly created scan and dependent observations, or only make a scan active after successful completion.

src/controllers/event_bus.py
```diff
@@ -0,0 +1,156 @@
...
+    def _offer(self, event: dict) -> None:
+        try:
+            self._queue.put_nowait(event)
+        except queue.Full:
+            self.overflowed = True
+
+    def close(self) -> None:
+        self._bus.unsubscribe(self)
+        try:
+            self._queue.put_nowait(None)
```
sfl-pe-sync, line 51: WARNING: `close()` and `close_all()` silently fail to enqueue their sentinel when a slow subscriber's queue is full. After that subscriber drains its backlog, `closed` is still false and the SSE loop blocks again forever instead of emitting `bye` and releasing its stream slot. Record closure independently of queue capacity or evict an item before inserting the sentinel, and cover the full-queue shutdown path.

src/controllers/operation_queue.py
```diff
@@ -0,0 +1,238 @@
...
+            job.ctx.flush()
+            registry.update(
+                job.op_id,
+                status=STATUS_CANCELLED,
+                progress={"current": 0, "total": 0, "message": "Cancelled"},
+                append_logs=["Operation cancelled"],
+            )
+        except Exception as error:  # noqa: BLE001 - surfaced to the client verbatim
+            job.ctx.flush()
+            message = str(error)[:500] or error.__class__.__name__
```
sfl-pe-sync: WARNING: Arbitrary exception text is stored verbatim in the operation and is intended to be returned by the operations API/SSE stream. SQLAlchemy and subprocess exceptions can contain SQL, parameters, filesystem paths, or upstream response details, exposing internal data to clients. Log the full exception server-side and publish only a sanitized, user-safe error message (with explicit mappings for expected job failures).

sfl-pe-sync (changes requested):
The queue primitives introduce blocking correctness regressions: native-package exclusion is lost, deferred refreshes can include stale tool findings when the active SBOM is empty, and cancellation can promote partial scans to active state. Stream shutdown and error sanitization also need hardening.

src/controllers/scan_jobs.py
reh-sfl: Fixed

src/controllers/refresh_jobs.py
reh-sfl: Fixed

src/controllers/scan_jobs.py
reh-sfl: Fixed

src/controllers/event_bus.py
reh-sfl: Fixed

src/controllers/operation_queue.py
reh-sfl: Fixed

src/controllers/scan_jobs.py
```diff
@@ -0,0 +1,789 @@
...
+        scan_type="tool",
+        scan_source=source,
+    )
+    try:
+        yield scan
+    except Exception:
+        db.session.rollback()
+        persisted_scan = db.session.get(Scan, scan.id)
+        if persisted_scan is not None:
+            db.session.delete(persisted_scan)
```
sfl-pe-sync, line 105: ERROR: Deleting only the failed scan does not undo rows committed while that scan was running. OSV/NVD can commit via `Vulnerability.create_record()`, and SCC explicitly commits each `_SccBulkWriter` chunk; removing the `Scan` cascades its observations but leaves newly created findings and variant-targeted pending assessments behind with no successful scan. Those stale assessments are returned by assessment queries and can suppress creation of the correct initial assessment on a later scan. Make a failed/cancelled run atomic, or track and remove every finding/assessment/target created solely by that run (and orphan vulnerabilities where appropriate), with a regression test that cancels after a committed new finding rather than using a pre-existing one.

src/controllers/refresh_jobs.py
```diff
@@ -0,0 +1,446 @@
...
+    """Return vulnerability IDs visible in the active scans for the variants."""
+    scan_ids = [
+        scan_id
+        for variant_id in variant_ids
+        for scan_id in active_scan_ids_for_variant(uuid.UUID(variant_id))
+    ]
+    if not scan_ids:
+        return set()
+
+    package_ids = active_package_ids_for_scans(scan_ids)
```
sfl-pe-sync, line 102: ERROR: Active package IDs are unioned across all requested variants and then applied to every selected tool scan. If variant A's latest tool scan references package P that is stale in A, but the same global package row P is active in variant B, A's historical finding passes this predicate even when B never observed it. Deferred refreshes therefore include vulnerabilities that are not current in any selected variant. Resolve vulnerabilities per variant (pairing each variant's tool scans with that variant's active SBOM package IDs) and union only the resulting vulnerability IDs.

src/controllers/refresh_jobs.py
```diff
@@ -0,0 +1,446 @@
...
+    operation = registry.get(ctx.op_id)
+    if operation is not None and operation.get("queue_id") is not None:
+        preceding_scans = [
+            candidate for candidate in registry.snapshot()
+            if candidate.get("queue_id") == operation["queue_id"]
+            and candidate.get("kind") == "scan"
+            and candidate.get("position", 0) < operation.get("position", 0)
+        ]
+        if any(scan.get("status") in {STATUS_ERROR, STATUS_CANCELLED} for scan in preceding_scans):
+            ctx.report(0, 0, "Skipped because an earlier scan did not complete")
```
sfl-pe-sync, line 132: WARNING: Returning normally when a preceding scan failed makes `OperationQueue._run_body()` overwrite this skipped refresh with `status=done`. Clients then see a successful refresh even though it never ran. Propagate a dedicated skipped/cancelled outcome (or raise an expected operation error) so the terminal status reflects the dependency failure.

src/controllers/refresh_jobs.py
```diff
@@ -0,0 +1,446 @@
...
+            return
+
+    source = str(ctx.options["source"])
+    existing_ids = set(ctx.options.get("exclude_ids", []))
+    scoped_ids = _scoped_vulnerability_ids(list(ctx.options["variant_ids"]))
+    if source == "ghsa":
+        ids = normalise_ghsa_ids(sorted(scoped_ids - existing_ids))
```
sfl-pe-sync, line 139: WARNING: Deferred GHSA refreshes do not enforce `MAX_GHSA_IDS`; only NVD API mode is bounded here. A large multi-variant batch can therefore enqueue thousands of one-second-spaced GitHub requests and monopolize the single pipeline lane for hours, bypassing the 500-ID safeguard used by explicit GHSA refreshes. Reject or deliberately truncate deferred GHSA IDs at the same limit before invoking the runner.

sfl-pe-sync (changes requested):
The queue primitives are improved, but cancelled/failed scans can still leave committed assessment data behind, and multi-variant deferred refresh scoping can include historical findings. Deferred dependency status and GHSA workload bounds also need correction.

sfl-zwang (approved):
LGTM

src/controllers/scan_jobs.py
reh-sfl: I verified this and it can safely be resolved without a fix.

src/controllers/refresh_jobs.py
reh-sfl: I verified this and it can safely be resolved without a fix.

src/controllers/refresh_jobs.py
reh-sfl: I verified this and it can safely be resolved without a fix.

src/controllers/refresh_jobs.py
reh-sfl: I verified this and it can safely be resolved without a fix.

PR #565, opened 2026-09-18
feat(operations): expose the queue API and the SSE event stream

reh-sfl:
> [!NOTE]
> **Stacked PR 2 of 6 — split out of #529.** Review and merge in order; each PR targets the one before it.
>
> 2. **#549 — queue REST API and SSE event stream → base `#548` ← you are here**
>

---

* Add `GET /api/events/stream`, a single Server-Sent Events connection carrying every operation update: a full `snapshot` first, then `operation` / `operation_removed` deltas, a `heartbeat` every 15 s, and `bye` on shutdown. `Last-Event-ID` replays from a bounded buffer, or triggers a fresh snapshot when the gap is too wide.
* Add `POST /api/operations`, which accepts a whole batch of scans and refreshes in one request and expands it into individually tracked operations. Scans run before refreshes; each family keeps its canonical order (`grype → nvd → osv → scc`, then `nvd → epss → ghsa → euvd`).
* Add cooperative cancellation for scans, which previously had none — only the four bulk refreshes could be cancelled. A cancel request also terminates the Grype subprocess.
* Run three lanes reproducing the concurrency the frontend used to enforce by hand: `pipeline` (all scans and refreshes, strictly serial), `export` (two at a time), and `upload` (one thread each).
* Register the queue on the app in `bin/webapp.py` and surface the boot enrichment thread as a tracked operation — it previously reported nothing at all.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

The legacy polling endpoints are all still present and untouched in this layer, so the running app is unaffected. `curl -N http://localhost:7275/api/events/stream` should emit a `snapshot` immediately and a `heartbeat` every 15 s. `POST /api/operations` with a multi-scan batch should return one operation id per expanded operation, and those should appear on the stream in canonical order.

### Additional notes

Adding the new surface before removing the old one keeps this PR and the next one independently reviewable, and means the stack is never in a state where neither mechanism works.

### Related Issue

No linked issue.

## Pull Request Checklist

### DIFF
--- a/src/bin/webapp.py
+++ b/src/bin/webapp.py
@@ -8,8 +8,10 @@
 from ..helpers.add_middleware import FlaskWithMiddleware as Flask
 from ..helpers.env_vars import get_bool_env
 from ..extensions import db, migrate, setup_write_serialization
+from ..controllers.operation_queue import queue as operation_queue
 from ..routes import init_app
 from ..routes.documents import MAX_ASSET_UPLOAD_BYTES
+from ..routes.events import STREAM_PATH
 from .. import models  # noqa: F401
 from .merger_ci import init_app as init_merger_cli, post_treatment
 import sys
@@ -28,26 +30,34 @@ DEFAULT_BACKGROUND_TASK_DELAY = 120.0
 
 
 def _launch_enrichment(app):
-    """Spawn background thread for EPSS enrichment.
+    """Queue boot EPSS enrichment so its progress reaches the event stream."""
+    from ..controllers.job_context import JobContext
+    from ..controllers.operation_queue import queue
+    from ..controllers.operation_registry import (
+        KIND_ENRICHMENT, LANE_PIPELINE, registry,
+    )
 
-    Runs in its own thread so it doesn't block Flask request handlers
-    (WAL journal mode allows concurrent reads).
-    """
-    def _enrich_epss():
-        with app.app_context():
-            # Disable autoflush: without this, every SELECT triggers a flush
-            # which acquires the write-lock and holds it across the slow HTTP
-            # calls until the next explicit commit().  With autoflush=False
-            # the lock is only held during commit() itself (milliseconds).
-            db.session.autoflush = False
-            try:
-                from ..controllers import ControllersCache
-                controllers = ControllersCache()
-                post_treatment(controllers)
-            except Exception as e:
-                print(f"[enrichment/epss] {e}", flush=True)
+    op_id = "enrichment:boot"
+    if registry.has_active(op_id):
+        return
 
-    threading.Thread(target=_enrich_epss, name="enrichment-epss", daemon=True).start()
+    def _enrich_epss(ctx):
+        # Disable autoflush: without this, every SELECT triggers a flush which
+        # acquires the write-lock and holds it across the slow HTTP calls until
+        # the next explicit commit(). With autoflush=False the lock is only
+        # held during commit() itself (milliseconds).
+        db.session.autoflush = False
+        from ..controllers import ControllersCache
+        post_treatment(ControllersCache(), reporter=ctx)
+
+    registry.create(
+        op_id=op_id,
+        kind=KIND_ENRICHMENT,
+        source="epss",
+        label="EPSS enrichment",
+        lane=LANE_PIPELINE,
+    )
+    queue.submit(op_id, LANE_PIPELINE, _enrich_epss, JobContext(op_id))
 
 
 def _warm_scan_list_cache(app):
@@ -238,12 +248,18 @@ def create_app():
 
     @app.middleware("/api")
     def fail_scan_not_finished(*args, **kw):
-        if request.path in {"/api", "/api/openapi", "/api/openapi.json", "/api/openapi/ui"}:
+        # The event stream is exempt like /api/scan/status: the UI opens it
+        # during boot import so it can render enrichment progress live.
+        if request.path in {
+            "/api", "/api/openapi", "/api/openapi.json", "/api/openapi/ui",
+            STREAM_PATH,
+        }:
             return None
         if not is_scan_finished():
             return {"error": "Scan not finished"}, 503
 
     init_app(app)
+    operation_queue.init_app(app)
     init_merger_cli(app)
     return app
 
--- a/src/routes/events.py
+++ b/src/routes/events.py
@@ -1,0 +1,155 @@
+# Copyright (C) 2026 Savoir-faire Linux, Inc.
+# SPDX-License-Identifier: GPL-3.0-only
+
+"""Server-sent event stream carrying every operation update.
+
+One connection replaces the per-scan, per-refresh, upload and export polling
+loops the frontend used to run. A new connection starts with a full snapshot;
+a reconnecting client that sends a usable ``Last-Event-ID`` receives the
+buffered deltas it missed instead, and falls back to a fresh snapshot when the
+gap is too wide to replay.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import threading
+from typing import Iterator, Optional
+
+from flask import Flask, Response, jsonify, request, stream_with_context
+from flask.typing import ResponseReturnValue
+
+from ..controllers.event_bus import operation_events
+from ..controllers.operation_registry import registry
+
+STREAM_PATH = "/api/events/stream"
+
+# Long enough to stay cheap, short enough that proxies and load balancers do
+# not treat an idle stream as dead.
+HEARTBEAT_SECONDS = 15.0
+
+DEFAULT_MAX_STREAMS = 32
+
+_active_streams = 0
+_streams_lock = threading.Lock()
+
+
+def _max_streams() -> int:
+    """Cap concurrent streams; each one occupies a server thread."""
+    try:
+        return max(1, int(os.environ.get("VULNSCOUT_SSE_MAX_STREAMS", DEFAULT_MAX_STREAMS)))
+    except ValueError:
+        return DEFAULT_MAX_STREAMS
+
+
+def _acquire_slot() -> bool:
+    global _active_streams
+    with _streams_lock:
+        if _active_streams >= _max_streams():
+            return False
+        _active_streams += 1
+        return True
+
+
+def _release_slot() -> None:
+    global _active_streams
+    with _streams_lock:
+        _active_streams = max(0, _active_streams - 1)
+
+
+def _frame(event_type: str, data: dict, seq: Optional[int] = None) -> str:
+    lines = []
+    if seq is not None:
+        lines.append(f"id: {seq}")
+    lines.append(f"event: {event_type}")
+    lines.append(f"data: {json.dumps(data, default=str)}")
+    return "\n".join(lines) + "\n\n"
+
+
+def _snapshot_frame() -> tuple[str, int]:
+    seq = operation_events.seq
+    return _frame("snapshot", {"seq": seq, "operations": registry.snapshot()}), seq
+
+
+def _parse_last_event_id() -> Optional[int]:
+    raw = request.headers.get("Last-Event-ID") or request.args.get("last_event_id")
+    if not raw:
+        return None
+    try:
+        return int(raw)
+    except ValueError:
+        return None
+
+
+def _stream(last_event_id: Optional[int]) -> Iterator[str]:
+    # Subscribe before snapshotting so nothing published in between is lost;
+    # overlapping events are filtered by sequence below.
+    subscription = operation_events.subscribe()
+    try:
+        highest_sent = 0
+        replay = (
+            operation_events.replay_since(last_event_id)
+            if last_event_id is not None else None
+        )
+        if replay is not None:
+            highest_sent = last_event_id or 0
+            for event in replay:
+                yield _frame(event["event"], event["data"], event["seq"])
+                highest_sent = max(highest_sent, event["seq"])
+        else:
+            # No usable history: the client rebuilds from a full snapshot.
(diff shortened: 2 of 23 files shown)


src/routes/operations.py
```diff
@@ -0,0 +1,412 @@
...
+
+    planned: List[dict] = []
+    for source in SCAN_ORDER:
+        planned.extend(scans.get(source, []))
+    for source in REFRESH_ORDER:
+        planned.extend(refreshes.get(source, []))
+    return planned
+
+
+def _conflicts(planned: List[dict]) -> List[str]:
```
sfl-pe-sync: ERROR: This conflict check only sees the new registry, while the retained `/api/variants/<id>/*-scan` and bulk-refresh endpoints track running work separately. Starting a legacy scan and then submitting the same variant/source here admits both jobs concurrently; likewise a legacy refresh can overlap the pipeline. They can write competing scans/assessments despite the promised serial lane. Coordinate admission with the legacy trackers (or route both APIs through the same queue) until those endpoints are removed.

src/routes/operations.py
```diff
@@ -0,0 +1,412 @@
...
+    if source not in SCAN_JOBS:
+        raise _PlanError(f"Unknown scan source: {source}")
+
+    variant_ids = job.get("variant_ids") or []
+    if not isinstance(variant_ids, list) or not variant_ids:
+        raise _PlanError(f"{source} scan requires a non-empty variant_ids list")
+
+    options = _job_options(job)
+    mode = _mode_option(options)
+    exclude_kernel = _bool_option(options, "exclude_kernel", True)
```
sfl-pe-sync: ERROR: The batch API drops `options.exclude_native`: `_plan_scan_job` only reads `exclude_kernel`, while all four queued scanner bodies consult `ctx.options['exclude_native']` (defaulting to false). The existing Scan History wizard exposes this filter and the legacy triggers honor it, so submitting the equivalent batch silently scans `-native` packages and changes findings. Validate and forward `exclude_native` for every scan job.

src/bin/webapp.py
```diff
@@ -29,26 +31,34 @@
...
+        return
+
+    def _enrich_epss(ctx):
+        # Disable autoflush: without this, every SELECT triggers a flush which
+        # acquires the write-lock and holds it across the slow HTTP calls until
+        # the next explicit commit(). With autoflush=False the lock is only
+        # held during commit() itself (milliseconds).
+        db.session.autoflush = False
+        from ..controllers import ControllersCache
+        post_treatment(ControllersCache(), reporter=ctx)
```
sfl-pe-sync: WARNING: `post_treatment` returns an `EnrichmentResult` whose `failed` count includes API and commit failures, but this runner ignores it. The queue therefore marks boot enrichment `done` even when EPSS fetches failed and scores were not saved; unlike the old untracked thread, the new status now falsely reports success. Inspect the result and surface a failed/partial outcome in the operation instead of treating every normal return as success.

src/routes/operations.py
```diff
@@ -0,0 +1,412 @@
...
+        ``/api/events/stream``; there is no status endpoint to poll.
+
+        OpenAPI:
+        body JsonObject optional Batch of jobs to queue.
+        response 202 JsonObject Queue accepted with the created operations.
+        response 400 Error Invalid job specification.
+        response 404 Error Referenced variant not found.
+        response 409 Error One of the requested operations is already active.
+        """
+        body = request.get_json(force=True, silent=True) or {}
```
sfl-pe-sync: WARNING: A valid JSON request whose root is an array/string/number reaches `body.get('jobs')` and raises an unhandled exception, producing HTTP 500 instead of the documented validation error. Reject non-object JSON bodies before calling `.get` (and reject non-string/unhashable `source` values in `_plan_scan_job`/`_plan_refresh_job`, which also currently raise TypeError on arrays or objects).

src/routes/events.py
```diff
@@ -0,0 +1,155 @@
...
+                yield _frame(event["event"], event["data"], event["seq"])
+                highest_sent = max(highest_sent, event["seq"])
+        else:
+            # No usable history: the client rebuilds from a full snapshot.
+            frame, highest_sent = _snapshot_frame()
+            yield frame
+
+        while True:
+            events = subscription.drain(HEARTBEAT_SECONDS)
+            if subscription.closed:
```
sfl-pe-sync: WARNING: The advertised `bye` frame has no production shutdown path: it is emitted only after `subscription.closed`, but `operation_events.close_all()` is never called by the app or its signal handler. On shutdown active connections simply drop without this event. Wire bus closure into the server shutdown lifecycle (and test it via the production path) or remove the promised `bye` contract.

sfl-pe-sync (changes requested):
The new batch and SSE surfaces are not ready to merge: the queue can overlap retained legacy jobs, and the batch scanner silently ignores the existing native-package filter. Boot enrichment reporting, malformed-input handling, and shutdown signaling also need correction. Review was read-only; runtime checks and final-container startup were not run.

src/routes/operations.py
reh-sfl: This is a real overlap while #565 is deployed alone: legacy scan/refresh handlers are outside the operation registry. This PR keeps those routes during the staged transition; #566 removes the legacy scan/refresh trigger handlers, so the overlap ends when the next stack PR is applied. Until then, callers must not mix legacy triggers with the new queue for the same variant/source.

src/routes/operations.py
reh-sfl: Fixed. `exclude_native` is validated and forwarded for each queued scan; the scanner behavior is also carried through #566–#568.

src/bin/webapp.py
reh-sfl: Fixed. Boot enrichment now fails the operation when any EPSS enrichment fails.

src/routes/operations.py
reh-sfl: Fixed. Non-object JSON roots and non-string scan/refresh sources now return validation errors.

src/routes/events.py
reh-sfl: Fixed. The production SIGINT/SIGTERM shutdown handler now closes the event bus, emitting `bye` to active SSE subscriptions.

src/routes/events.py
```diff
@@ -0,0 +1,155 @@
...
+        return None
+
+
+def _stream(last_event_id: Optional[int]) -> Iterator[str]:
+    # Subscribe before snapshotting so nothing published in between is lost;
+    # overlapping events are filtered by sequence below.
+    subscription = operation_events.subscribe()
+    try:
+        highest_sent = 0
+        replay = (
```
sfl-pe-sync: ERROR: Replay IDs are process-local counters, but this accepts a `Last-Event-ID` from a previous server instance as soon as the new instance's counter catches up. For example, a client retaining ID 2 reconnects after a restart when two boot-operation events have been published: `replay_since(2)` is empty, no snapshot is sent, and the client keeps stale operations from the old process indefinitely. Include a server epoch in event IDs and force a snapshot when it changes (or otherwise distinguish IDs from earlier processes).

src/routes/events.py
```diff
@@ -0,0 +1,155 @@
...
+    if seq is not None:
+        lines.append(f"id: {seq}")
+    lines.append(f"event: {event_type}")
+    lines.append(f"data: {json.dumps(data, default=str)}")
+    return "\n".join(lines) + "\n\n"
+
+
+def _snapshot_frame() -> tuple[str, int]:
+    seq = operation_events.seq
+    return _frame("snapshot", {"seq": seq, "operations": registry.snapshot()}), seq
```
sfl-pe-sync: WARNING: Full snapshots, including those sent when the replay buffer is exhausted or a subscriber overflows, omit the SSE `id:` field even though they establish state at `seq`. A browser reconnecting after such a resync keeps its previous `Last-Event-ID`, repeatedly requests an obsolete replay position, and must download the full snapshot again on every disconnect. Emit the snapshot sequence as its event ID and cover replay-gap/overflow reconnection in tests.

src/routes/operations.py
```diff
@@ -0,0 +1,416 @@
...
+        raise _PlanError(f"{key} must be a boolean")
+    return value
+
+
+def _plan_scan_job(job: dict) -> List[dict]:
+    source = job.get("source")
+    if not isinstance(source, str) or source not in SCAN_JOBS:
+        raise _PlanError(f"Unknown scan source: {source}")
+
+    variant_ids = job.get("variant_ids") or []
```
sfl-pe-sync: WARNING: A single request can supply an arbitrarily long `variant_ids` list (including repeated IDs); planning performs a database lookup for every entry before duplicate detection, then enqueues/publishes one operation per entry for valid distinct IDs. A large JSON body can monopolize the request and flood the in-memory registry/SSE bus and pipeline with work. Enforce a bounded batch/expanded-operation count before DB lookup and reject repeated IDs early.

sfl-pe-sync (changes requested):
The queue API and stream expose useful operation tracking, but reconnecting across a server restart can silently preserve stale client state. Snapshot cursor handling and unbounded batch expansion also need attention.

src/routes/events.py
reh-sfl: Fixed. SSE event IDs now include a per-process epoch, so a cursor from before a restart triggers a fresh snapshot even when sequence numbers overlap. Covered by a restart reconnect test; propagated through #566–#568.

src/routes/events.py
reh-sfl: Fixed. Every snapshot now carries the current epoch-prefixed sequence as its SSE id, including replay-gap and subscriber-overflow resyncs. Reconnect tests cover both paths; propagated through #566–#568.

src/routes/operations.py
reh-sfl: Fixed. The API now caps expanded batches at 100 operations, bounds deferred refresh variant lists, and detects canonical duplicate variant IDs before any variant DB lookup. Tests cover oversized and repeated IDs across jobs; propagated through #566–#568.

src/routes/events.py
```diff
@@ -0,0 +1,156 @@
...
+            if subscription.overflowed:
+                subscription.overflowed = False
+                frame, highest_sent = _snapshot_frame()
+                yield frame
+                continue
+            if not events:
+                yield _frame("heartbeat", {"seq": highest_sent})
+                continue
+            for event in events:
+                if event["seq"] <= highest_sent:
```
sfl-pe-sync: ERROR: Concurrent publishers can deliver frames out of sequence: `EventBus.publish()` assigns the sequence under its lock but calls each subscriber's `_offer` after releasing it. If operation A gets seq 1 and pauses before `_offer`, then B gets seq 2 and is offered first, this filter discards A when it finally arrives. The client permanently misses A (and reconnecting from 2 cannot replay it). Serialize sequence assignment and subscriber delivery together, or otherwise guarantee ordered fan-out before advancing `highest_sent`.

src/bin/webapp.py
```diff
@@ -278,19 +292,27 @@ def loading():
...
 
 
 def stop_handler(signal, frame):
     print("Stopping Flask server")
+    operation_events.close_all()
     sys.exit(0)
 
 
 signal.signal(signal.SIGINT, stop_handler)
+signal.signal(signal.SIGTERM, stop_handler)
```
sfl-pe-sync: ERROR: Registering this SIGTERM handler at module import also affects every `flask --app src.bin.webapp` CLI invocation used by `src/entrypoint.sh` for imports, scans, exports and Grype subprocesses. A SIGTERM during an import now calls `sys.exit(0)`, so the shell sees success and can delete the staged input files even though work was interrupted. Scope the SSE shutdown handler to the web server, and preserve a nonzero/signal exit status for CLI processes.

src/routes/events.py
```diff
@@ -0,0 +1,156 @@
...
+        """Stream every operation update over a single connection.
+
+        Frames are ``snapshot`` (full state, sent first), ``operation`` and
+        ``operation_removed`` deltas, ``heartbeat`` every 15 seconds, and
+        ``bye`` on shutdown. Reconnecting clients that send ``Last-Event-ID``
+        are replayed from the buffer, or resnapshotted when the gap is too wide.
+
+        OpenAPI:
+        header Last-Event-ID string optional Resume from this instance-scoped cursor.
+        response 200 string Server-sent event stream.
```
sfl-pe-sync: WARNING: `response 200 string` is emitted by `build_openapi_spec` as `application/json` (`_media_type_for_token` only recognizes html/multipart/binary), although this route returns an unbounded `text/event-stream`. OpenAPI clients generated from the new endpoint will use the wrong parser. Add an SSE response media type/schema to the OpenAPI helper and declare it here.

sfl-pe-sync (changes requested):
The queue and SSE API still need ordered fan-out and safe SIGTERM handling before this patchset is ready; the stream's OpenAPI media type also needs correction. I did not run local commands because this review prohibits shell tools.

src/routes/events.py
reh-sfl: This was fixed as part of the PR 3/5, the comment can be resolved safely.

src/bin/webapp.py
reh-sfl: This was fixed as part of the PR 3/5, the comment can be resolved safely.

src/routes/events.py
reh-sfl: This was fixed as part of the PR 3/5, the comment can be resolved safely.

sfl-zwang (approved):
LGTM

PR #566, opened 2026-09-18
refactor(operations): remove the 21 legacy polling endpoints

reh-sfl:
> [!NOTE]
> **Stacked PR 3 of 6 — split out of #529.** Review and merge in order; each PR targets the one before it.
>
> 3. **#550 — delete the 21 legacy polling endpoints → base `#549` ← you are here**
>

---

* Delete 21 endpoints now served by `POST /api/operations` and `GET /api/events/stream`: 9 status/progress, 8 triggers, and 4 cancels.
* Remove `routes/scan_triggers.py` and `routes/bulk_refresh.py` entirely — their job bodies already moved to `controllers/scan_jobs.py` and `controllers/refresh_jobs.py` earlier in the stack.
* Remove the four `ProgressTracker` singletons and their route wrappers (`routes/{nvd,epss,ghsa,euvd}_progress.py` and the matching controllers), along with the now-unused half of `routes/_scan_helpers.py`.
* Drop the test modules that only covered the deleted endpoints.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Backend-only removal: `make test` in `tests/` should pass. The frontend still calls some of these paths until the last PR in the stack, so this layer is not meant to be merged to `staging` on its own — it merges into #549 and reaches `staging` with the rest of the stack.

Confirm the OpenAPI document no longer advertises any `/progress`, `/scan/trigger`, or `/cancel` path.

### Additional notes

Kept as its own PR because it is almost pure deletion (-5366 lines). Reviewing it separately means the additive work in the surrounding PRs is not buried under the removal diff.

### Related Issue

No linked issue.

## Pull Request Checklist


src/routes/__init__.py
```diff
@@ -26,15 +22,11 @@ def init_app(app):
     init_vuln_app(app)
     init_assess_app(app)
     init_doc_app(app)
-    init_nvd_progress_app(app)
-    init_epss_progress_app(app)
-    init_ghsa_progress_app(app)
-    init_euvd_progress_app(app)
+    init_events_app(app)
```
sfl-pe-sync: ERROR: Removing the legacy route registrations while shipping the unchanged frontend breaks the Scan History wizard: `ScansHandler.triggerGrypeScan` (and the NVD/OSV/SCC equivalents) still POST to the deleted `/api/variants/<id>/*-scan` paths, and `activeScanQueue` still POSTs to the deleted bulk-refresh paths. These calls now return 404 and no scans or refreshes start. Merge the frontend migration in the same deployable change (or retain compatibility routes until it ships); the new `/api/operations` route does not redirect existing callers.

src/controllers/refresh_jobs.py
```diff
@@ -0,0 +1,444 @@
...
+
+    package_ids = active_package_ids_for_scans(scan_ids)
+    query = (
+        db.select(Vulnerability.id)
+        .join(Finding, Vulnerability.id == Finding.vulnerability_id)
+        .join(Observation, Finding.id == Observation.finding_id)
+        .join(Scan, Observation.scan_id == Scan.id)
+        .where(Observation.scan_id.in_(scan_ids))
+    )
+    if package_ids:
```
sfl-pe-sync: ERROR: Deferred refresh combines active SBOM package IDs across all requested variants and skips the filter entirely when that set is empty. After an empty replacement SBOM, the latest historical tool scan's findings are refreshed as if still active; with multiple variants, a package active in B can also make a stale tool finding in A pass the union filter. Correlate each tool finding with the active SBOM packages of its own scan's variant and keep the tool filter active even when that set is empty.

src/controllers/event_bus.py
```diff
@@ -0,0 +1,158 @@
...
+            return len(self._subscribers)
+
+    def publish(self, event_type: str, data: dict) -> dict:
+        """Stamp *data* with a monotonic sequence id and fan it out."""
+        with self._lock:
+            self._seq += 1
+            event = {"seq": self._seq, "event": event_type, "data": data}
+            self._replay.append(event)
+            targets = list(self._subscribers)
+        for sub in targets:
```
sfl-pe-sync: ERROR: Sequence allocation and replay insertion happen under `_lock`, but subscriber delivery happens after releasing it. Two publisher threads can allocate seq 1 then seq 2 yet enqueue seq 2 first; `_stream` sets `highest_sent=2` and discards seq 1, potentially leaving an operation missing from the live client until reconnect. Serialize enqueueing with sequence assignment, or reorder drained events by sequence and recover gaps.

src/controllers/operation_registry.py
```diff
@@ -0,0 +1,201 @@
...
+            status = operation["status"]
+            if status == STATUS_RUNNING and operation["started_at"] is None:
+                operation["started_at"] = _now()
+            if status in TERMINAL_STATUSES:
+                if operation["finished_at"] is None:
+                    operation["finished_at"] = _now()
+                self._finished_at[op_id] = time.monotonic()
+            snapshot = dict(operation)
+            snapshot["logs"] = list(operation["logs"])
+        operation_events.publish("operation", snapshot)
```
sfl-pe-sync: ERROR: Publishing after releasing the registry lock allows an older snapshot to be published after a newer update (or after `remove`), for example a running update delayed while a completion publishes first. The SSE client then ends with an obsolete running operation although `registry.get()` says done. Publish each create/update/remove in the same mutation order under a shared lock or attach/version-check revisions before emission.

src/controllers/operation_queue.py
```diff
@@ -0,0 +1,238 @@
...
+            job.ctx.flush()
+            registry.update(
+                job.op_id,
+                status=STATUS_CANCELLED,
+                progress={"current": 0, "total": 0, "message": "Cancelled"},
+                append_logs=["Operation cancelled"],
+            )
+        except Exception as error:  # noqa: BLE001 - surfaced to the client verbatim
+            job.ctx.flush()
+            message = str(error)[:500] or error.__class__.__name__
```
sfl-pe-sync: ERROR: Every unexpected exception is copied verbatim into the public operation `error` and `logs` exposed by GET `/api/operations` and the SSE stream. A SQLAlchemy failure can include SQL and parameter values; subprocess/engine failures can include filesystem paths or environment-derived details. Log the full exception server-side and send a generic message for unexpected errors; expose only explicitly safe `OperationError` messages.

src/controllers/operation_registry.py
```diff
@@ -0,0 +1,201 @@
...
+            operation = self._operations.get(op_id)
+            if operation is None or operation["status"] not in TERMINAL_STATUSES:
+                return False
+            del self._operations[op_id]
+            self._finished_at.pop(op_id, None)
+        operation_events.publish("operation_removed", {"op_id": op_id})
+        return True
+
+    def prune(self) -> None:
+        """Drop terminal operations older than :data:`TERMINAL_TTL_SECONDS`."""
```
sfl-pe-sync: WARNING: The advertised one-hour TTL never takes effect: production code does not call `registry.prune()` (only its unit test does). Every distinct variant/scan operation and up to 500 logs remain in process memory and in every snapshot for the server lifetime. Invoke pruning on a timer or during bounded read/write paths so finished operations expire.

sfl-pe-sync (changes requested):
Request changes: deleting the old routes leaves the shipped frontend unable to start scans or refreshes, and the replacement queue has scoping, SSE ordering, and error-disclosure defects. The declared operation expiry also needs wiring.

src/routes/__init__.py
reh-sfl: This is a valid deployability issue. PR #566 intentionally removes the legacy routes as step 3/5, but the current frontend (including the #568 branch as checked here) still calls some old scan/refresh URLs. I am leaving this thread open: the frontend callers must migrate to /api/operations before this stack is deployed, or the compatibility routes must remain until that migration lands. The backend fixes in 69d3cd86 address the other review threads, not this one.

src/controllers/refresh_jobs.py
reh-sfl: Refresh variant-specific filtering and the empty-SBOM regression are covered in 69d3cd86.

src/controllers/event_bus.py
reh-sfl: Event enqueue now occurs under the sequence lock, with a concurrency regression test added.

src/controllers/operation_registry.py
reh-sfl: Registry create/update/remove and prune event publication now occur under the registry lock, with regression coverage added.

src/controllers/operation_queue.py
reh-sfl: Unexpected errors are already sanitized to a generic string and logged server-side in the current branch; see the existing operation_queue code and test.

src/controllers/operation_registry.py
reh-sfl: Snapshot triggers now perform TTL pruning without resetting the terminal timestamp, covered by the existing test.

src/controllers/operation_registry.py
```diff
@@ -178,6 +177,7 @@ def get(self, op_id: str) -> Optional[dict]:
 
     def snapshot(self) -> List[dict]:
         with self._lock:
```
sfl-pe-sync: ERROR: Pruning on every snapshot can erase a failed scan before a deferred refresh in the same batch starts. If an early scan fails and subsequent queued scans take over an hour (for example NVD API scans at 6 seconds per CPE), `run_deferred_refresh()` calls `registry.snapshot()`, which now removes that failure; its preceding-scan check then passes and refreshes vulnerabilities from an incomplete pipeline. Retain batch scan outcomes until all dependent refreshes finish, or track the batch failure independently of the display TTL.

sfl-pe-sync (changes requested):
The new snapshot-time expiry can discard scan failures that deferred refreshes still need, allowing an incomplete scan batch to proceed with enrichment.

src/controllers/operation_registry.py
reh-sfl: This is intentional as enrichment can happen even if some scans fail, in fact we enrich whatever new data was introduced to the system at the user's request.

src/routes/__init__.py
```diff
@@ -30,17 +24,11 @@ def init_app(app):
     init_assess_app(app)
     init_assess_review_app(app)
     init_doc_app(app)
-    init_bulk_refresh_app(app)
-    init_nvd_progress_app(app)
-    init_epss_progress_app(app)
-    init_ghsa_progress_app(app)
-    init_euvd_progress_app(app)
     init_events_app(app)
```
sfl-pe-sync: WARNING: Removing these route registrations leaves `doc/source/api.md` advertising `POST /api/vulnerabilities/bulk-nvd-refresh`, its cancel endpoint, all four `/api/variants/<id>/*-scan` triggers and their `/status` endpoints, plus `/api/nvd/progress` and `/api/epss/progress`. API users following the published examples now receive 404. Replace those sections with `POST /api/operations`, `GET /api/operations`, `GET /api/events/stream`, and operation cancellation examples as part of this removal.

sfl-pe-sync (approved):
The legacy routes are removed, but the published REST API guide still directs clients to deleted endpoints. No additional blocking defect beyond the previously reported frontend incompatibility was found; no validation commands were run in this review.

sfl-zwang (changes requested):
Most LGTM, please update docs like `api.md`.

src/routes/__init__.py
sfl-zwang: Agreed, please update `api.md`.

sfl-zwang (approved):
As agreed. Docs will be updated in a later PR

PR #567, opened 2026-09-18
refactor(operations): route exports and uploads through the queue

reh-sfl:
> [!NOTE]
> **Stacked PR 4 of 6 — split out of #529.** Review and merge in order; each PR targets the one before it.
>
> 4. **#551 — route exports and uploads through the queue → base `#550` ← you are here**
>

---

* Move document export off `_export_jobs` and onto the queue's `export` lane in `routes/documents.py`, so export progress arrives on the same event stream as everything else.
* Move SBOM/CVE upload off `_upload_status` and onto the queue's `upload` lane in `routes/settings.py`.
* Update the affected endpoint tests to assert against operation state instead of the removed job dictionaries.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

`make test` in `tests/` should pass. Triggering an export or an upload should now produce an operation on `GET /api/events/stream` with the correct lane, and two exports requested at once should run concurrently while a third waits.

### Additional notes

With this PR the backend has a single progress mechanism; the remaining stack is entirely frontend.

### Related Issue

No linked issue.

## Pull Request Checklist


src/routes/documents.py
```diff
@@ -719,49 +734,37 @@ def update_progress(current: int, total: int, element_name: str) -> None:
...
+            registry.create(
+                op_id=op_id,
+                kind=KIND_EXPORT,
+                source="documents",
+                label=f"Export {project.name}",
+                lane=LANE_EXPORT,
+                scope={"project_id": str(project_uuid), "project_name": project.name},
+            )
+            operation_queue.submit(op_id, LANE_EXPORT, generate_archive, JobContext(op_id))
+            return {"op_id": op_id}, 202
```
sfl-pe-sync: ERROR: The shipped `frontend/src/handlers/exportQueue.ts` still requires `body.job_id` and polls `GET /api/documents/export/<job_id>`. This response now contains only `op_id`, and that status route has been removed, so every export started from the UI is reported as failed and never downloaded even when generation succeeds. Update the frontend to consume the operation stream and download using `op_id` in this patch (or retain the old contract until the frontend is deployed).

src/routes/settings.py
```diff
@@ -1279,37 +1254,31 @@ def _cleanup_validated() -> None:
...
+            op_id, LANE_UPLOAD,
+            lambda job_ctx: _process_sbom_background(
+                job_ctx, tmp_paths, scan.id, variant.id, refresh_sources
+            ),
+            ctx,
+        )
 
         return jsonify({
-            "upload_id": upload_id,
+            "op_id": op_id,
```
sfl-pe-sync: ERROR: The shipped Settings upload flow reads `result.upload_id` and polls `GET /api/sbom/upload/<upload_id>/status`; this response now supplies only `op_id` and the polled route is removed. Every successful upload therefore polls `/api/sbom/upload/undefined/status`, reports an error, and never refreshes the UI after import. Migrate the frontend upload handler and completion flow to operations before removing the old contract, or retain compatibility.

src/routes/documents.py
```diff
@@ -436,49 +439,73 @@ def flush(self) -> None:
...
 
-    retained.sort(key=finished_at)
+def _retained_archives() -> List[Tuple[str, Dict[str, object], Optional[datetime]]]:
+    """Export operations still owning an archive on disk, oldest first.
+
+    The operation registry is the archive table: the path lives in the
+    operation ``result``, so retention accounting reads straight from it.
+    """
+    entries: List[Tuple[str, Dict[str, object], Optional[datetime], datetime]] = []
+    for operation in registry.snapshot():
```
sfl-pe-sync: ERROR: `_retained_archives()` calls `registry.snapshot()`, which prunes terminal operations after its own 3600-second TTL without unlinking their archive files. Once any operation snapshot or create prunes an undownloaded export, its path disappears from the only archive table, so subsequent export retention and exit cleanup cannot find it; repeated exports can fill the container filesystem despite the 512 MiB budget. Keep archive ownership outside registry pruning or make pruning/dismissal unlink owned archives before removing operations.

src/routes/documents.py
```diff
@@ -719,49 +734,37 @@ def update_progress(current: int, total: int, element_name: str) -> None:
...
-                    with _export_jobs_lock:
-                        _prune_export_jobs()
+                    with _export_archive_lock:
+                        _prune_export_archives()
                         _reserve_retained_archive(os.path.getsize(archive_path))
-                        _export_jobs[job_id].update({
-                            "status": "done",
-                            "progress": "Export ready",
+                        ctx.message("Export ready")
+                        ctx.set_result({
```
sfl-pe-sync: ERROR: `ctx.set_result()` publishes the archive while the operation is still `running`; the queue sets `done` only after this function returns and releases `_export_archive_lock`. If another export finishes in that interval and the retention budget is full, `_reserve_retained_archive()` selects this running result, deletes its ZIP, but `registry.remove()` refuses to remove a nonterminal operation. The first export then reports `done` with `download_ready: true` but cannot be downloaded. Make publication of a retained, terminal export atomic with eviction, or exclude in-flight results from eviction while accounting for their bytes.

src/routes/documents.py
```diff
@@ -683,34 +710,22 @@ def export_documents() -> ResponseReturnValue:
         filename = f"{project_name}_{suffix}_export.zip"
 
         if body.get("async") is True:
-            if not _export_capacity.acquire(blocking=False):
+            active_exports = sum(
```
sfl-pe-sync: WARNING: The new capacity check and `registry.create()` are not atomic. Concurrent export requests can all observe fewer than six active exports before any registers, then all enqueue; unlike the previous bounded semaphore this permits an unbounded number of queued ZIP jobs under concurrent load. Serialize the count-and-register step or acquire/release a bounded capacity token per accepted export.

src/routes/documents.py
```diff
@@ -719,49 +734,37 @@ def update_progress(current: int, total: int, element_name: str) -> None:
...
-                    if isinstance(exc, FileNotFoundError):
-                        message = "Required conversion tool was not found"
-                    with _export_jobs_lock:
-                        _export_jobs[job_id].update({
-                            "status": "error",
-                            "error": message,
-                            "progress": "Export failed",
-                            "finished_at": time.monotonic(),
-                        })
+                except ExportGenerationError as exc:
```
sfl-pe-sync: WARNING: Converting expected `ExportGenerationError` and missing-tool failures into plain `RuntimeError` causes `OperationQueue._run_body()` to replace their messages with the generic `Operation failed; check server logs`. Users no longer see actionable errors such as archive size exceeded or converter missing, which the old export status returned. Raise the queue's `OperationError` for these sanitized expected failures and leave unexpected exceptions generic.

src/routes/documents.py
```diff
@@ -719,49 +734,37 @@ def update_progress(current: int, total: int, element_name: str) -> None:
...
-                        _prune_export_jobs()
+                    with _export_archive_lock:
+                        _prune_export_archives()
                         _reserve_retained_archive(os.path.getsize(archive_path))
-                        _export_jobs[job_id].update({
-                            "status": "done",
-                            "progress": "Export ready",
+                        ctx.message("Export ready")
+                        ctx.set_result({
                             "archive_path": archive_path,
```
sfl-pe-sync: WARNING: Publishing `archive_path` in `ctx.set_result()` sends the absolute temporary filesystem path to every SSE subscriber and `GET /api/operations` caller, whereas the old export status explicitly omitted it. Keep the path in a server-only archive map and expose only `filename` and `download_ready` in the public operation result.

sfl-pe-sync (changes requested):
Request changes: moving exports and uploads to the shared queue breaks both shipped frontend workflows, and the new export archive lifecycle can leak files or report a deleted archive as ready. The capacity check, error reporting, and public result also need correction.

src/routes/documents.py
reh-sfl: This is the backend 4/5 layer. The frontend SSE operation store is in #568 (5/5); the export workflow must consume that store and download by op_id in the subsequent frontend workflow migration. Keeping this thread open until that integration is verified.

src/routes/settings.py
reh-sfl: This is the backend 4/5 layer. The frontend SSE operation store is in #568 (5/5); the Settings upload flow must move to op_id and operation completion in the subsequent frontend workflow migration. Keeping this thread open until that integration is verified.

src/routes/documents.py
reh-sfl: Fixed in e7bd599. Archive paths now live in a server-only table independent of registry TTL pruning; retention and exit cleanup unlink from that table.

src/routes/documents.py
reh-sfl: Fixed in e7bd599. Retention counts in-flight archive bytes, never evicts a running export, and rejects a new archive when it cannot fit; covered by a regression test.

src/routes/documents.py
reh-sfl: Fixed in e7bd599. Active-export counting, registration, and queue submission are serialized under the submission lock.

src/routes/documents.py
reh-sfl: Fixed in e7bd599. Expected export and missing-tool errors now use OperationError, preserving actionable messages while unexpected errors remain generic.

src/routes/documents.py
reh-sfl: Fixed in e7bd599. Public operation results include only filename and download_ready; archive paths are kept in the server-only table.

src/routes/documents.py
```diff
@@ -786,35 +771,28 @@ def update_progress(current: int, total: int, element_name: str) -> None:
...
-        with _export_jobs_lock:
-            _prune_export_jobs()
-            job = _export_jobs.get(job_id)
-            if job is None:
+    @app.route('/api/documents/export/<op_id>/download', methods=['GET'])
+    def download_export(op_id: str) -> ResponseReturnValue:
+        with _export_archive_lock:
+            _prune_export_archives()
+            operation = registry.get(op_id)
+            if operation is None or operation.get("kind") != KIND_EXPORT:
```
sfl-pe-sync: WARNING: A finished export can be dismissed through `DELETE /api/operations/<op_id>`, which removes its registry entry but leaves its ZIP in `_export_archives`. This download check then returns 404 even though the archive still occupies up to 512 MiB; no timer prunes the file unless another export/download request arrives or the process exits. Coordinate export dismissal with archive cleanup (or prevent dismissal until download) so forgetting an operation also releases its artifact.

sfl-pe-sync (approved):
The queue migration still leaves dismissed exports with inaccessible archives consuming disk space until another export or download triggers cleanup.

sfl-zwang (approved):

PR #568, opened 2026-09-18
feat(operations): add the frontend SSE operation store and render the queue from it

reh-sfl:
> [!NOTE]
> **Stacked PR 5 of 6 — split out of #529.** Review and merge in order; each PR targets the one before it.
>
> 5. **#552 — frontend SSE operation store, queue rendered from it → base `#551` ← you are here**
>

---

* Add `handlers/operationStore.ts`, a single subscribable store fed by one `EventSource` on `/api/events/stream`. It applies the initial `snapshot`, then `operation` / `operation_removed` deltas, tracks `Last-Event-ID` for replay, and reconnects with backoff.
* Add `handlers/operations.ts` (batch submit and cancel) and `types/operation.ts`.
* Render `OperationQueueModal` and `OperationQueuePanel` from store state instead of from locally reconstructed progress, so every tab shows the identical queue.

### Status

- [X] READY
- [ ] HOLD
- [ ] WIP (Work-In-Progress)

### How to verify this change

Open the app in two browser tabs and launch a scan. Both tabs should show the identical queue, including still-`queued` items after a reload. DevTools should show exactly one open `/api/events/stream` request.

The old polling handlers are still present in this layer and are removed in the next PR.

### Additional notes

Split from the migration PR so the new store can be reviewed on its own, before the diff that deletes the code it replaces.

### Related Issue

No linked issue.

## Pull Request Checklist


frontend/src/components/OperationQueueModal.tsx
```diff
@@ -1,38 +1,102 @@
...
+}
 
 function OperationQueueModal({ isOpen, onClose }: Readonly<Props>) {
-    const grypeEntries = useSyncExternalStore(grypeSubscribe, grypeGetSnapshot);
-    const nvdEntries = useSyncExternalStore(nvdSubscribe, nvdGetSnapshot);
-    const osvEntries = useSyncExternalStore(osvSubscribe, osvGetSnapshot);
-    const sccEntries = useSyncExternalStore(sccSubscribe, sccGetSnapshot);
-    const refreshEntries = useSyncExternalStore(subscribeToRefreshQueue, getRefreshQueueSnapshot);
-    const exportEntries = useSyncExternalStore(exportSubscribe, exportGetSnapshot);
+    const operations = useSyncExternalStore(subscribe, getSnapshot);
```
sfl-pe-sync: ERROR: This replaces the only rendering of the legacy scan/refresh stores, but `ScanHistory.runSelectedScans()` still queues into `grypeQueueScan`/the other legacy managers and `queueVulnerabilityRefresh()` instead of `Operations.enqueue()`. Those locally queued entries (including later variants and refreshes) are never published as SSE operations, so launching the existing scan wizard now leaves its queue absent or incomplete. Migrate the producers in this patchset, or keep rendering the legacy entries until they are migrated.

frontend/src/components/OperationQueueModal.tsx
```diff
@@ -45,26 +109,29 @@ function OperationQueueModal({ isOpen, onClose }: Readonly<Props>) {
             size="large"
             contentClassName="overflow-y-auto"
         >
-            {hasEntries ? (
+            {connection === "reconnecting" && (
+                <p role="status" className="border-b border-amber-700/60 bg-amber-900/30 px-4 py-2 text-xs text-amber-200">
+                    Reconnecting to the operation stream…
+                </p>
+            )}
+            {operations.length > 0 ? (
```
sfl-pe-sync: ERROR: The modal now displays SSE operations, but `Explorer` still calculates `trackedScanCount` and `activeScanCount` solely from the old managers, and `NavigationBar` only renders its Operation queue button when `trackedScanCount > 0`. After reloading a tab with server-side queued operations (or when another tab starts an operation), the SSE snapshot contains the queue but there is no way to open this modal. Drive the navbar counts and auto-open behavior from the same operation store.

frontend/src/handlers/operationStore.ts
```diff
@@ -0,0 +1,295 @@
...
+
+function parse<T>(event: MessageEvent): T | null {
+    try {
+        return JSON.parse(event.data) as T;
+    } catch {
+        return null;
+    }
+}
+
+function trackEventId(event: MessageEvent) {
```
sfl-pe-sync: WARNING: The server emits SSE IDs as `<epoch>:<seq>` (`events._frame`) and only replays if that complete cursor is supplied. `parseInt` either returns NaN for most epochs or truncates an epoch starting with a digit; the snapshot path likewise saves only `frame.seq`. Every reconnect therefore sends an invalid `last_event_id` and forces a full snapshot, losing replay of transient events and doing unnecessary full-state transfers. Preserve `event.lastEventId` verbatim (including for snapshots) and URL-encode it on reconnect; exercise this against an epoch-qualified test ID.

frontend/src/handlers/operationStore.ts
```diff
@@ -0,0 +1,295 @@
...
+    snapshot.some(operation => isActive(operation) && (kind === undefined || operation.kind === kind));
+
+export const hasActiveSource = (kind: OperationKind, source_: string): boolean =>
+    snapshot.some(operation => isActive(operation) && operation.kind === kind && operation.source === source_);
+
+/** Resolves once no operation of *queueId* is queued or running. */
+export const waitForQueue = (queueId: string): Promise<Operation[]> =>
+    new Promise(resolve => {
+        const settled = () => {
+            const batch = selectByQueue(queueId);
```
sfl-pe-sync: WARNING: `waitForQueue` considers the batch settled as soon as any one matching operation is terminal. The backend publishes each operation separately while enqueueing, so an SSE consumer can observe the first finished operation before later jobs in the same batch have arrived and resolve early, letting dependent work start while the batch is still running. Use the expected operation IDs/count returned by enqueue (or a server batch-terminal signal) to ensure every job has been observed and settled before resolving.

sfl-pe-sync (changes requested):
The SSE queue rendering is not yet wired to the existing workflow producers or navbar, so the shipped scan queue is incomplete or inaccessible; reconnect replay also uses an incompatible cursor. Code inspection only; runtime checks were not run because shell use was prohibited.

frontend/src/pages/ScanHistory.tsx
```diff
@@ -1381,57 +1376,29 @@ function ScanHistory({ variantId, projectId, variantIds, onScanComplete }: Reado
...
+            ...scanSources.filter(source => selectedScanTypes.has(source))
+                .map(source => Operations.scanJob(source, targetVariantIds, { excludeKernel, excludeNative })),
+            ...refreshTypes.map(source => Operations.deferredRefreshJob(
+                source, targetVariantIds, [...existingVulnerabilityIds],
+            )),
+        ];
+        const result = await Operations.enqueue(jobs);
+        if (!result.ok) throw new Error(result.error);
+        await waitForQueue(result.queueId, result.operations.map(operation => operation.op_id));
+        refreshScans();
```
sfl-pe-sync: ERROR: If a scan is launched for variant A and the user selects variant B before the batch finishes, this suspended `runSelectedScans` retains A's `refreshScans` and `onScanComplete` callbacks. When it resumes, `refreshScans()` aborts B's in-flight history request and loads A's scans into the still-mounted B view; `onScanComplete()` likewise reloads A's packages/vulnerabilities into the B view. Only refresh the currently selected scope on completion (or guard by a scope generation/current-scope ref), and test changing variants during a queued scan.

frontend/src/components/OperationQueueModal.tsx
```diff
@@ -45,26 +109,29 @@ function OperationQueueModal({ isOpen, onClose }: Readonly<Props>) {
...
+                    {operations.map(operation => {
+                        const { icon, colors } = appearanceFor(operation);
+                        return (
+                            <OperationQueuePanel
+                                key={operation.op_id}
+                                operation={operation}
+                                icon={icon}
+                                colors={colors as never}
+                                positionLabel={positionLabelFor(operation, operations)}
+                                onDismiss={() => void Operations.dismiss(operation.op_id)}
```
sfl-pe-sync: ERROR: Dismissing a completed export now calls `DELETE /api/operations/<op_id>`, which removes it from the registry. The export download endpoint requires that same registry entry and returns 404 without it (`src/routes/documents.py:778-780`); if the user dismisses the SSE card before the legacy export poll/download fires, the generated archive can no longer be downloaded. Keep export operations until their download is consumed, or make the download independent of the dismissed registry entry.

frontend/src/components/RefreshVulnerabilityData.tsx
```diff
@@ -94,34 +98,36 @@ function RefreshVulnerabilityData({ vulnerabilities, getRefreshVulnerabilities,
...
+                latest.map(vulnerability => vulnerability.id).filter(id =>
+                    source === 'ghsa' ? id.toUpperCase().startsWith('GHSA-') : id.toUpperCase().startsWith('CVE-')
+                ),
+            )).filter(job => job.ids?.length);
+            if (jobs.length === 0) return;
+            const result = await Operations.enqueue(jobs);
+            if (!result.ok) throw new Error(result.error);
             setRefreshMenuOpen(false);
             triggerBanner('Vulnerability data refresh added to the operation queue', 'success');
+            await waitForQueue(result.queueId, result.operations.map(operation => operation.op_id));
```
sfl-pe-sync: WARNING: Refreshes started through `Operations.enqueue` update only the SSE operation registry, but `TableVulnerabilities` still obtains all four progress values from the legacy `/progress` handlers and starts its completion-driven reload from those values. This waiter reloads only the initiating tab; another tab already showing a vulnerability table never sees the new refresh complete and continues displaying stale scores/dates until manually reloaded. Wire the table's progress and completion effects to the operation store (including the SSE adapter already provided) so every mounted tab updates.

frontend/src/handlers/operationStore.ts
```diff
@@ -0,0 +1,296 @@
...
+
+/** Resolves once every operation returned by enqueue has been observed and settled. */
+export const waitForQueue = (queueId: string, expectedOperationIds: readonly string[]): Promise<Operation[]> =>
+    new Promise(resolve => {
+        const completed = new Map<string, Operation>();
+        const settled = () => {
+            selectByQueue(queueId).forEach(operation => {
+                if (!isActive(operation)) completed.set(operation.op_id, operation);
+            });
+            return expectedOperationIds.length > 0
```
sfl-pe-sync: WARNING: Completion is inferred solely from terminal operations still observed in the snapshot. A fast first operation dismissed from another tab before this tab receives its terminal event (or during a stream gap followed by a fresh snapshot) will never enter `completed`, so `waitForQueue` never resolves even after the remaining jobs finish; scan-history and vulnerability refresh completion callbacks then never run. Make batch completion observable independently of dismissible cards, e.g. a server batch terminal signal/status, or retain completed batch outcomes until all waiters have settled.

sfl-pe-sync (changes requested):
The SSE queue is connected to its producers, but scope changes during a scan can populate the wrong variant's view, and dismissing an export can prevent its download; refresh completion also remains disconnected from other tabs' vulnerability tables.

sfl-zwang (approved):

frontend/src/handlers/operationStore.ts
reh-sfl: This comment can safely be ignored.

frontend/src/components/RefreshVulnerabilityData.tsx
reh-sfl: This comment can safely be ignored.

frontend/src/components/OperationQueueModal.tsx
reh-sfl: This comment can safely be ignored.

frontend/src/pages/ScanHistory.tsx
reh-sfl: This comment can safely be ignored.

frontend/src/components/OperationQueueModal.tsx
```diff
@@ -1,38 +1,102 @@
...
+
+/**
+ * " (variant 2 of 3)" for multi-variant scan batches.
+ *
+ * The backend numbers positions across the whole batch, so the index is
+ * recomputed within each scan source.
+ */
+function positionLabelFor(operation: Operation, all: readonly Operation[]): string {
+    if (operation.kind !== "scan" || operation.queue_id === null) return "";
+    const siblings = all.filter(
```
sfl-pe-sync: WARNING: This derives the variant index and total from cards still present in the SSE snapshot. In a three-variant scan, dismissing variant 1 after it finishes makes variant 2 read “variant 1 of 2” while it is still queued; dismissing another changes the labels again. The old entry carried the original `variantPosition`/`variantCount`. Preserve the batch positions/count independently of dismissible cards (for example, include the per-source position and total in operation metadata).

frontend/src/components/OperationQueueModal.tsx
```diff
@@ -1,38 +1,102 @@
...
 
-    const hasEntries = grypeEntries.length + nvdEntries.length + osvEntries.length + sccEntries.length + refreshEntries.length + exportEntries.length > 0;
+    const requestCancel = (opId: string) => {
+        setCancelRequested(previous => new Set(previous).add(opId));
+        const restore = () => setCancelRequested(previous => {
+            const next = new Set(previous);
+            next.delete(opId);
+            return next;
+        });
+        Operations.cancel(opId).then(cancelled => {
```
sfl-pe-sync: WARNING: If cancelling a scan/refresh returns 404/409 or the request fails, this only re-enables the Cancel button and shows no error. A user can believe the job was stopped while it keeps running, especially after closing the modal. Surface the failure with the application's banner/error UI, as the refresh-menu cancel path does, while restoring the button.

sfl-pe-sync (approved):
The SSE queue integration has two non-blocking UI defects: dismissing a finished variant renumbers the remaining batch, and failed cancellation is silent.

